PR #246, opened 2012-06-14
Inherit sidekiq_options in subclass

### LINKED ISSUE #221: Subclass doesn't inherit sidekiq_options
Let's say I have a superclass like this:

``` ruby
class BaseWorker
  include Sidekiq::Worker
  sidekiq_options :retry => false
end
```

And the following subclass:

``` ruby
class UpdateGamesWorker < BaseWorker

  def perform
    GamesWorker.execute
  end

end
```

UpdateGamesWorker does not inherit the sidekiq_options; it'll just use the default retry setting unless you define it explicitly in the subclass.

kulesa:
Fixes #221 by storing sidekiq_options as a method on the eigenclass rather than a class instance variable.

### DIFF
--- a/lib/sidekiq/worker.rb
+++ b/lib/sidekiq/worker.rb
@@ -51,11 +51,16 @@ module Sidekiq
       #   :backtrace - whether to save any error backtrace in the retry payload to display in web UI,
       #      can be true, false or an integer number of lines to save, default *false*
       def sidekiq_options(opts={})
-        @sidekiq_options = get_sidekiq_options.merge(stringify_keys(opts || {}))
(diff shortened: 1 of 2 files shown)


RichardsonWTR:
@mperham , we faced this issue today in the 0.15.0 version. 
Is this PR relevant (with the proper cleaner version just like that you've mentioned)?

RichardsonWTR:
Hmm.... I've made a mistake!  Sorry!
I've just realized, by reading your comment.    

The option that isn't being inherited is `sidekiq_throttle_as` from sidekiq-throttle.

Just forget about this comment. 
If you're curious, I've opened an issue there https://github.com/sensortower/sidekiq-throttled/issues/109

PR #1272, opened 2013-10-22
Display the full Redis connection in Web footer

jlecour:
The web interface shows the Redis location in the footer.

Since we can connect to Redis with the full array of options, we can have a combination of socket and non-0 database, which would not be properly displayed.

The Redis client supports an `#id` method that just does what we need.

I've added a helper method `#redis_connection` that delegates to Redis' client ID method.

sidekiq 2013-10-22 16-23-09

Since there is no test for this, I've not added some, and I didn't delete the existing `#location` helper either.

### DIFF
--- a/lib/sidekiq/web_helpers.rb
+++ b/lib/sidekiq/web_helpers.rb
@@ -76,6 +76,10 @@ module Sidekiq
     def location
       Sidekiq.redis { |conn| conn.client.location }
     end
+    
+    def redis_connection
+      Sidekiq.redis { |conn| conn.client.id }
+    end
 
     def namespace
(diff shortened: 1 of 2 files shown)


newdark:
Heads up by displaying the redis connection in the footer in some cases if the domain name is to long it will prevent you from clicking the buttons at the bottom. We have this problem in the Pro Version as well.

newdark:
kinda hard to see but as you can see because of the length you can't click the button even though I am at the bottom of the page. 

screen shot 2015-02-26 at 10 28 05 pm

mperham:
@newdark Please don't comment on ancient issues; open a new issue if it's something you feel needs work.

PR #1545, opened 2014-03-11
Implement process lifecycle events, fixes #1534

### LINKED ISSUE #1534: Add process lifecycle events
There's probably a lot of people interested in hearing when Sidekiq has:
1. Finished booting but before work is started
2. Starting quiet (USR1)
3. Starting shutdown (TERM)

Allow one or more blocks to be defined for each event.  Possible API:

``` ruby
  config.on(:start) do
  end
  config.on(:quiet) do
  end
  config.on(:shutdown) do
  end
end
```

mperham:
See #1534

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -5,28 +5,31 @@
 
 Please see [Upgrading.md](Upgrading.md) for more comprehensive upgrade notes.
 
-- **Global Error Handlers** - blocks of code which handle errors that
-  occur anywhere within Sidekiq, not just within middleware.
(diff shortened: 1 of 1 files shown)


kylejw2:
Were these changes removed? If so, is there another way to hook into the process's lifecycle?

mperham:
> Were these changes removed? If so, is there another way to hook into the process's lifecycle?

Do not resurrect old issues. Open your own new discussion or issue.

PR #1973, opened 2014-09-29
Allow label color selection

pgeraghty:
This may or may not be undesirable fluff, but it's a very small addition - the main usage case being varied label colour rather than a vast number of labels.

labels


mperham:
I think I'll decline this.  Looks like you've got more customizations already on your branch anyways.

PR #2008, opened 2014-10-16
Customizable title

seuros:


mperham:
I don't want any more class methods on Sidekiq::Web.  I like your last suggestions much better.

PR #2073, opened 2014-12-04
Test Celluloid's TaskThread runner for stability

mperham:
Sidekiq uses Celluloid's TaskFiber by default.  MRI was recently discovered to have a Fiber bug which could lead to uncommitted transactions, open locks and other really odd behavior.  I'm unclear how to reproduce the problem but we should check TaskThread again to see if it's a stable replacement.

https://bugs.ruby-lang.org/issues/10540

/cc @halorgium

### DIFF
--- a/lib/sidekiq/actor.rb
+++ b/lib/sidekiq/actor.rb
@@ -7,6 +7,8 @@ module Sidekiq
       def new_link(*args)
         new(*args)
       end
+      def task_class(*args)
+      end
     end
 
     module InstanceMethods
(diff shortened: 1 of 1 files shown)


halorgium:
I have been running this change in production for a few months. 
We have not encountered the same issues, but have received some `Celluloid::Task::TerminatedError` exceptions. 
These might not be related.

mperham:
Is there some documentation on the differences/advantages of TaskThread vs TaskFiber that we can link in this issue?  If people see instability with this change, I want them to understand the pros/cons of TaskThread.

PR #2088, opened 2014-12-16
Added profiling with ruby-prof

felipeclopes:
I'm adding back the profiling with ruby-prof. 

It was removed due to uselessness, but proved to be useful in some cases.

### DIFF
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -76,6 +76,12 @@ module Sidekiq
       @launcher = Sidekiq::Launcher.new(options)
 
       begin
+        if options[:profile]
+          logger.info 'Profiling Sidekiq, hit Ctrl-C and check profile.html for results'
+          require 'ruby-prof'
+          RubyProf.start
+        end
+
         launcher.run
 
         while readable_io = IO.select([self_read])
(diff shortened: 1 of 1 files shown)


aidansteele:
@mperham I stumbled across your `rb_profile_frames()` issue earlier today as I bumped up against the same problem. I was a bit too impatient to wait for the function to be added in Ruby 2.x, so I hacked support for it into a [fork][1] of `stackprof`. Emphasis on "hack". 

In any case, it allowed me to make a [middleware][2] to profile Sidekiq jobs. Right now it's fairly rigid and expects to work with a [GUI tool][3] I started last month to browse Stackprof profiles.

There's one (somewhat substantial) limitation with the middleware in that right now it can't profile multiple jobs concurrently. I don't think this is *too* much of an issue at the moment as devs will probably be drilling down into issues with a given job and don't want to be flooded with data. In any case, it's an issue I'm working on.

This has gone really off-topic for this issue. I apologise for the rambling. The reason I brought it up in *this* issue on the Sidekiq tracker is because the middleware may be of interest to @felipeclopes, even if it's very early state.

[1]: https://github.com/glassechidna/stackprofx
[2]: https://github.com/glassechidna/stackprofiler-sidekiq
[3]: https://github.com/glassechidna/stackprofiler

mperham:
@aidansteele Looks promising.  Keep it up, Ruby needs better developer tools.

PR #2102, opened 2014-12-26
Cleans up execute_job

jejacks0n:
A while back I submitted [a PR](https://github.com/mperham/sidekiq/pull/1938) that exposed a way to customize how a job was executed, in so much as how perform was actually called on the worker. I duplicated that in the sidekiq/testing, and have since determined that was a mistake and am rectifying that here.

In doing what I talk about in that original PR, I realized there was still some finagling to get it to behave correctly for the specs, and then I realized that it wasn't very unified and didn't push through to the actual functionality that I wanted to make sure was working -- after some back and forth (captured on https://github.com/mperham/sidekiq/pull/2097) @mperham decided he'd rather remove it.

Also cleans up the drain method in the test harness.

### DIFF
--- a/lib/sidekiq/processor.rb
+++ b/lib/sidekiq/processor.rb
@@ -49,7 +49,7 @@ module Sidekiq
 
         stats(worker, msg, queue) do
           Sidekiq.server_middleware.invoke(worker, msg, queue) do
-            execute_job(worker, cloned(msg['args']))
+            worker.perform(*cloned(msg['args']))
           end
         end
       rescue Sidekiq::Shutdown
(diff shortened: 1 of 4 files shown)


lib/sidekiq/processor.rb
```diff
@@ -71,10 +71,6 @@ def inspect
       "<Processor##{object_id.to_s(16)}>"
     end
 
-    def execute_job(worker, cloned_args)
```
mperham: BTW, I wasn't offended.  I was literally agreeing with you: I wasn't sure I understood the issue either!  :smile:

Issue #2071, opened 2014-12-03
Cleaning up after hard shutdowns


michaeldiscala:
Hi there,

We're running Sidekiq v3.3.0 with  sidekiq-pro v1.9.1 on Heroku and are having some issues with workers that don't get a chance to cleanup after themselves. Our workers get hard-shutdown if  either 1) the dyno shuts down abruptly without sending the SIGTERM or 2) the worker fails to exit within the 10s period. If jobs are actively running during one of these hard shutdowns, they remain permanently on the busy page until we manually clear the data structures in Redis. Aside from being annoying in the interface, these phantom jobs are also problematic because they interfere with our processes for auto-scaling dynos.

Two questions:
1) Is this the expected behavior in the face of a hard shutdown?
2) If so, what would your thoughts be on incorporating some sort of cleaning process that eliminates jobs associated with workers who haven't sent a heartbeat for a long time?

We're working to make these hard shutdowns less common (by talking with heroku and by getting our app to shutdown faster). Since I doubt that we'll ever eliminate them completely, however, we'd definitely appreciate thoughts on how to tackle them from the Sidekiq perspective as well.

Thanks so much!

michaeldiscala:
Addressed in #2113

PR #2113, opened 2014-12-31
Add UUID to process identity.

michaeldiscala:
In environments with deterministic hostnames and PIDs (i.e. Heroku), the
current system for constructing a process's identity string produces
collisions (e.g. a new process can assume the identity of one that has
died). In order to address this, we add a UUID to the identity string.

This has been tested with a local sidekiq server (I ran some jobs and
verified that signaling from the UI works properly).

Issue #2071.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -1,6 +1,6 @@
 HEAD
 -----------
-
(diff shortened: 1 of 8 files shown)


mperham:
Thank you!

PR #2133, opened 2015-01-12
Some more frozen strings.

jonhyman:
Note that both in this and the sidekiq-pro pull I didn't `.freeze` any hash accessors, because according to https://groups.google.com/forum/#!topic/rack-devel/SKAE_yqN8-0, it should be that in Ruby 2.2 that'll happen automatically, so I'd rather not litter the code any more than it needs to be.


mperham:
TIL. Thanks!

PR #2141, opened 2015-01-16
Add processes count to /stats

ismaelga:
Also removed unused `each_with_index`. Tell me if there is any reason to keep them or if I should remove it from the PR.

:yellow_heart:

### DIFF
--- a/lib/sidekiq/web_helpers.rb
+++ b/lib/sidekiq/web_helpers.rb
@@ -46,6 +46,14 @@ module Sidekiq
       @workers ||= Sidekiq::Workers.new
     end
 
+    def processes_size
+      @processes_size ||= processes.size
+    end
+
+    def processes
+      @processes ||= Sidekiq::ProcessSet.new
+    end
+
     def stats
       @stats ||= Sidekiq::Stats.new
     end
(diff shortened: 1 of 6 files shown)


seuros:
Please add an entry to the Changelog

lib/sidekiq/web_helpers.rb
```diff
@@ -46,6 +46,14 @@ def workers
       @workers ||= Sidekiq::Workers.new
     end
 
+    def processes_size
```
mperham: I don't think this method is necessary.

mperham:
Thanks for the cleanup, this is nice.  I should probably go through the Stats call and optimize the Redis usage.  We're probably making too many roundtrips to gather all the different stats.

lib/sidekiq/web_helpers.rb
ismaelga: I agree.

ismaelga:
Are you thinking on using pipelined to reduce the roundtrips?

mperham:
Exactly.  Merge all the operations to gather the stats into a single pipelined block.

mperham:
If you want to do that, feel free.  It'd be a welcome optimization.

ismaelga:
Sure. I'll have a look and send another PR.

mperham:
I'll take this as is and polish it up a bit.

ismaelga:
Cool. I'm close to send the other PR

PR #2142, opened 2015-01-16
Load stats with pipeline

ismaelga:
My attempt to load the stats in pipeline. This was the simplest way I could think of.

Comment are welcome :smile:

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -3,28 +3,89 @@ require 'sidekiq'
 
 module Sidekiq
   class Stats
+    def initialize
+      fetch_stats!
+    end
+
     def processed
-      stat 'processed'
+      stat :processed
     end
 
     def failed
-      stat 'failed'
+      stat :failed
     end
 
     def scheduled_size
-      stat 'scheduled_size'
+      stat :scheduled_size
     end
 
     def retry_size
-      stat 'retry_size'
+      stat :retry_size
     end
 
     def dead_size
-      stat 'dead_size'
+      stat :dead_size
+    end
+
+    def enqueued
+      stat :enqueued
+    end
+
+    def processes_size
+      stat :processes_size
+    end
+
+    def workers_size
+      stat :workers_size
+    end
+
+    def default_queue_latency
+      stat :default_queue_latency
     end
 
     def fetch_stats!
-      stats
+      pipe1_res = Sidekiq.redis do |conn|
+        conn.pipelined do
+          conn.get('stat:processed')
+          conn.get('stat:failed')
+          conn.zcard('schedule')
+          conn.zcard('retry')
+          conn.zcard('dead')
+          conn.scard('processes')
+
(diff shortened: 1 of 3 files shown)


mperham:
I'd write it simpler: load all metrics when Stats is initialized.  Include the process and worker counts too.

ismaelga:
I'm not sure that's a good idea. Sometimes you just want one or two stats so loading everything on initialisation does not seem a good approach.

Take this for example https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/web.rb#L234 (actually might be the only one :confused: )

About the process and worker counts I totally agree.

mperham:
This is true but the dominant cost is the network roundtrip and Stats is heavily used by the Web UI where it accesses most of the values on every page load.

ismaelga:
Maybe it could be optional? `Stats.new(true)` where true would mean load all. Or the other way around.

mperham:
I'm not clear what you are trying to optimize with the additional complexity.  Are you trying to minimize Redis operations?  I'd much rather avoid network round-trips and not make the developer worry about which mode they want to use.

jonhyman:
Fwiw I agree with @mperham here. Optimizing the network roundtrips is much
more important, especially for these calls which are really fast on Redis.

ismaelga:
Yes. I was trying to avoid extra operations on redis if someone just wanted to do `Stats.new.processed`.

To avoid roundtrips at all cost would you rather fit workers_size into this pipeline instead of using `Workers.new.size` for example?

mperham:
Exactly.

ismaelga:
Alright! Now I got it :sweat_smile:

ismaelga:
This should pretty close to what you said. It now only uses 2 pipelines + redis_stats. Should that also go into one pipeline?

I'm still figuring out why one test is not passing.

ismaelga:
Did a dummy benchmark by loading the data for /stats 1000 times: 1.527251s vs 5.147929s
:sunglasses:

ismaelga:
Did some dummy benchmarks:
```
before
/stats  0.790000   0.190000   0.980000 (  1.149291)
summary 0.480000   0.120000   0.600000 (  0.689375)

after
/stats   0.340000   0.060000   0.400000 (  0.440624)
summary  0.280000   0.060000   0.340000 (  0.377405)
```
:sunglasses:

mperham:
Fantastic work, @ismaelga!  3x improvement!!

ismaelga:
Actually the difference should be even bigger when redis and sidekiq are not running on the same machine. :)

I just forgot to add it to the Changelog. You do that?

davetoxa:
So, now it has a small transformation, of getting list of queues, is it? @ismaelga
`Sidekiq::Stats.new.queues` -> `Sidekiq::Stats::Queues.new.lengths`

mperham:
@davetoxa ^^^

davetoxa:
@mperham :smiley: :+1:

PR #2144, opened 2015-01-18
Improve web page's load speed

cookiebody:
Most of Google's services are blocked in China, including **fonts.googleapis.com**. The requests to it woud be timeout in China:

During loading fonts, the page is blank. After timeout(10-20 seconds), the page can be rendered.

So I make fonts lazy loading, try to improve page's load speed, especially in China.


seuros:
Let include the font in sidekiq. That will work in firewalled/offline setups.

mperham:
Alternatively, remove the font usage of the UI still looks good.

seuros:
:+1: for removal.

mperham:
I plan on a 3.3.1 release later this month.

PR #2155, opened 2015-01-26
Retry job without raising exception

dlibanori:
I've built a callback system based on Sidekiq. It is really simply. You register an URL and when something happens my server will make post to it. My clients love it. Joey, one of my best client, has the worst sysadmin I know. It is not so rare to get his site offline. I told him: don't worry, Sidekiq will try to post to your server sometime after first fail. Everything is working well, except, when Joey's site is offline, I got thousands exceptions. My boss gets crazy and start to call me. I would like to use Sidekiq retry system but I don't want to see those exceptions at my log, actually, I just want to say to Sidekiq: there is nothing I can do for it right know but, please, process this code as it has failed.

This _pr_ works fine, but it is still just a _POC_. I would like to know what do you think about it.

ps: as you see, english is not my first language ;)

### DIFF
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -18,6 +18,7 @@ module Sidekiq
   # otherwise Ruby's Thread#kill will commit.  See #377.
   # DO NOT RESCUE THIS ERROR.
   class Shutdown < Interrupt; end
+  class Retry < Exception; end
 
   class CLI
     include Util
--- a/lib/sidekiq/middleware/server/retry_jobs.rb
+++ b/lib/sidekiq/middleware/server/retry_jobs.rb
@@ -75,17 +75,22 @@ module Sidekiq
         rescue Sidekiq::Shutdown
           # ignore, will be pushed back onto queue during hard_shutdown
           raise
+        rescue Sidekiq::Retry => e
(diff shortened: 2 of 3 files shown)


mperham:
Retries are for exceptional problems.  If you want to handle something as part of your normal application logic, your worker should rescue and handle it as appropriate:

```ruby
def perform(*args)
  begin
    do_stuff
  rescue OfflineError
    if joey?
      self.class.perform_in(2.hours, *args)
    else
      raise
    end
  end
end
```

dlibanori:
@mperham unfortunately this job will never be exhausted, I will try after 2 hours forever :(

dlibanori:
I know I can build a retry system, something like:

```ruby
def perform(*args)
  begin
    do_stuff
  rescue OfflineError
    if retries_count < 100
      self.class.perform_in(2.hours, retries_count + 1)
    else
      raise Exhausted
    end
  end
end
```

Why I should code a poor retry system while Sidekiq has already implemented a nice retry heuristic? I just want use it as a built-in.

mperham:
You should have your own retry argument to cap the retries or build your own application retry middleware.  The point of Sidekiq's built-in retry system is to never reach it unless you have an actual bug which needs developer attention.  It is not meant to handle application-specific logic.

dlibanori:
Ok

PR #2156, opened 2015-01-26
Massive freeze

ismaelga:


mperham:
Not sure I want this, it litters the code with some much junk and Ruby's GC and compiler get smarter with every release.  Can you gather benchmarks that show a significant reduction in garbage?

ismaelga:
Fair enough. Will try to get some numbers.

ismaelga:
I haven't found free time to benchmark this. And probably it wouldn't make much difference, so I'm closing it.

PR #2157, opened 2015-01-26
Clean up logic and docs for Client#push(_bulk)

ryansch:
`Client#raw_push` always returns true now.  Therefore, there is no need to
check for a false return value and `Client#push` can only return nil if middleware stops the job.

The possible outcomes of a call to `push` are:
- a `jid` returned
- `nil` due to middleware
-  an exception raised

This change is motivated by changing the classification of a `nil` return from an error condition to a no-op that can be ignored unless the client code needs to know that the middleware interrupted.


ryansch:
Ahh I missed the early return from `process_single`.

PR #2162, opened 2015-01-28
Small Sidekiq::Stats refactor with metaprogramming

davetoxa:
I think its a little bit clear to use metaprogramming here

That do you think?

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -7,40 +7,15 @@ module Sidekiq
       fetch_stats!
     end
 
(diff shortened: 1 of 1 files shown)


seuros:
That make the code more complex to read and break autocomplete in my IDE.

mperham:
Go has taught me to prefer longer but simpler and more explicit coding styles.  It may seem wordy but the current  way means that no one needs to decipher the define_method block in their head.

davetoxa:
okey

PR #2169, opened 2015-01-30
Processor is not required as needed

AlexRiedler:
Calling Sidekiq.server_middleware will actually cause it to crash; This requires it earlier so that doesn't happen, but not late enough that we can't include it.


mperham:
server_middleware can only be called in the Sidekiq process.  You are trying to call it in console.  This is by design.

AlexRiedler:
there is middleware like sidetiq that will try to access this not in a sidekiq process.  Will move the fix to there then.

PR #2173, opened 2015-02-04
Allows configuration of the Dead Job Queue.

jonhyman:
We're going to start using the dead jobs. 10,000 jobs may be too small for us, I'd like to be able to configure it.

If this pull is merged in, I'll update the wiki.


mperham:
Seems pretty legit.  Doesn't look like any testing is really necessary.

PR #2182, opened 2015-02-06
Add an environment variable for backtrace depth:

alexbartlow:
We use exceptions for handing exceptional cases in our pipeline. We're seeing our papertrail logs fill up with spam from framework-level backtrace ; this adds an environment variable to cut that down to fewer lines.


mperham:
You should remove the default logger error handler and add your own custom copy to Sidekiq.options[:error_handlers] as line 14 does.

alexbartlow:
very cool, thanks.

PR #2188, opened 2015-02-12
Changed &block method argument to yield construction

davydovanton:
Hi,
As you may know `yield` construction more faster than `&block` in arguments (with `block.call`).
For example, for me `yield` faster than `block` to 5 times (ruby 2.1.5; *you can see benchmark below*).
And because that i changed `&block` method arguments on `yield` construction in all methods where `block` isn't need in other methods call.

### Benchmarks
*(all developers love benchmarks :smiley:)*

#### Code for compare yield and block
``` ruby
require 'benchmark/ips'

def yield_proc
  yield
end

def block_proc(&block)
  block.call
end

Benchmark.ips do |x|
  x.report('yield'){ yield_proc{ 1 + 1 } }
  x.report('block'){ block_proc{ 1 + 1 } }
end
```

#### Results
```
               yield   128.265k i/100ms
               block    62.716k i/100ms
-------------------------------------------------
               yield      5.821M (± 5.0%) i/s -     29.116M
               block      1.079M (±24.0%) i/s -      5.143M
```

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -240,7 +240,7 @@ module Sidekiq
     end
 
     def find_job(jid)
-      self.detect { |j| j.jid == jid }
+      detect { |j| j.jid == jid }
     end
 
     def clear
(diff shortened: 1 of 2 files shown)


mperham:
lgtm, waiting for travis.

davydovanton:
@mperham all passed :tada:

mperham:
thank you!

davydovanton:
and you to :star2:

PR #2205, opened 2015-02-27
Footer style fix #2204

### LINKED ISSUE #2204: Can't Click Buttons when Redis domain name to long
You can't click the buttons at the bottom when the domain name for the redis domain name is to long. We have this problem came up using services like object rocket.  

screen shot 2015-02-26 at 10 28 05 pm

seuros:


newdark:
It is almost worked I remove another 50px and it worked a lot better for me.
screen shot 2015-02-27 at 12 48 29 am

I would set the width on the namespace as well because that is another fluctuation variable that can effect the available width.

seuros:
will do

seuros:
can you try now ?

newdark:
Woot that worked great. Thanks so much :+1:

PR #2209, opened 2015-03-01
Travis jruby update

PragTob:
Removed jruby-19mode as this is kind of the default now and added testing with jruby-head (that may fail) to help JRuby get prepped for an awesome 9k release :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,8 @@ rvm:
   - 2.1
   - 2.0.0
   - 2.2
-  - jruby-19mode
   - jruby
   - rbx-2
 matrix:
   allow_failures:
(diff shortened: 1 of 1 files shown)


seuros:
We still need to test jruby 1.9 until version sidekiq 4.x

PragTob:
@seuros jruby-19 refers to the current jruby used in 1.9 mode. 1.9 is the default since jruby 1.7. Both jruby and jruby-19mode refer to jruby 1.7.19 so I don't think it's necessary.

seuros:
Thanks you for your contribution.

PragTob:
Thanks for your work on coderwall and state_machine gem among other things ;) :)

PR #2238, opened 2015-03-16
search feature for schedule, retries, dead queues

vishalzambre:
I have added search feature for schedule, retries, dead pages. Will help to search any job.


jonhyman:
Sidekiq pro offers search using zscan. Also, your solution is prohibitively slow for large sets, so it's not usable at scale.

mperham:
Sidekiq Pro already offers this feature, with a much more efficient implementation.  You can see a screenshot here:

https://github.com/mperham/sidekiq/wiki/UI-Filtering

PR #2242, opened 2015-03-19
ability to specify global queue for ActionMailer extension

jjb:
Threw this together just to get feedback.

If you like the idea of this feature I will write tests and improve the code.

(clearly `default_options` will not be named since it's generic to ActionMailer after being mixed in, and also I'll do something about the repeated code)

### DIFF
--- a/lib/sidekiq/extensions/action_mailer.rb
+++ b/lib/sidekiq/extensions/action_mailer.rb
@@ -38,13 +38,19 @@ module Sidekiq
     end
 
     module ActionMailer
+      def default_options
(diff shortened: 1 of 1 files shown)


mperham:
It's easy to reopen the class in your initializer and use `sidekiq_options`. WDYT?

jjb:
Okay, sounds good.

For anyone coming across this in the future wanting the same thing, here's how:

```ruby
module Sidekiq
  module Extensions
    class DelayedMailer
      sidekiq_options queue: "mailer"
    end
  end
end
```

PR #2246, opened 2015-03-20
Add '/stats/monitor' with useful stats for monitoring systems

mrsimo:
We have a custom endpoint in our application that returns this data and we have a couple of sensu scripts that warn us when the maximum latency gets over a certain threshold, and that records the total concurrency, total busy (of all our processes) and maximum latency in statsd.

I'd like to open source the sensu scripts too, but I'd like to make sure what the format will end up being before doing that. [This is how those scripts look](https://github.com/sensu/sensu-community-plugins/compare/master...mrsimo:sidekiq) right now (ignore the link to harvesthq/sidekiq-stats for now).

I opened the PR with this format since this is what we currently use, but as long as the endpoint lets us know those metrics we'd be happy. I realise this clashes a little with the `/stats/queues` endpoint, so I'd like to know what you think. I have no problem refactoring to match whatever we think is best, or close this PR altogether and release this in gem format.

An alternative would be to enhance the `/stats/queues` to also include the latency (we really like the latency metric over the backlog for monitoring) and create a new endpoint `/stats/processes` with the proper information.

This is how the final json would look:

```json
{
  "queues":{
    "default":{
      "backlog":15,
      "latency":2
    },
    "low":{
      "backlog":533,
      "latency":54
    }
  },
  "processes":[
    {
      "hostname":"kip1.example.com",
      "pid":23324,
      "queues":[
        "high"
      ],
      "concurrency":5,
      "busy":2
    },
    {
      "hostname":"kip2.example.com",
      "pid":23390,
      "queues":[
        "low"
      ],
      "concurrency":5,
      "busy":5
    }
  ]
}
```

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -191,15 +191,12 @@ module Sidekiq
         results = conn.pipelined do
           queues.each do |queue_name|
             rname = "queue:#{queue_name}"
-            conn.exists(rname)
             conn.llen(rname)
             conn.lrange(rname, -1, -1)
           end
         end
 
-        results.each_slice(3).with_index do |(exists, size, entries), idx|
-          next unless exists
-
+        results.each_slice(2).with_index do |(size, entries), idx|
           latency = if entry = entries.first
                       Time.now.to_f - Sidekiq.load_json(entry)['enqueued_at']
                     else
@@ -224,13 +221,12 @@ module Sidekiq
 
         results = conn.pipelined do
           procs.each do |key|
-            conn.exists(key)
             conn.hmget(key, 'info', 'busy')
           end
         end
 
-        results.each_slice(2) do |exists, (json_info, busy)|
-          next unless exists
+        results.each do |json_info, busy|
+          next unless json_info
 
           info = Sidekiq.load_json(json_info)
           info['busy'] = busy.to_i
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -176,6 +176,69 @@ module Sidekiq
   end
 
   ##
+  # Specific metrics useful for monitoring. These can all be gathered using the other
+  # classes in ths file, but this way we only use three roundtrips to Redis.
+  #
+  # Exposed in the web interface on /stats/monitor.
+  #
+  class Monitor
+    def initialize
+      Sidekiq.redis do |conn|
+        @procs, @queues = conn.pipelined do
+          conn.smembers('processes'.freeze)
+          conn.smembers('queues'.freeze)
+        end
+      end
+    end
+
+    def all_queue_metrics
+      metrics = {}
+
+      Sidekiq.redis do |conn|
+        results = conn.pipelined do
+          @queues.each do |queue_name|
+            rname = "queue:#{queue_name}"
+            conn.llen(rname)
+            conn.lrange(rname, -1, -1)
+          end
+        end
+
+        results.each_slice(2).with_index do |(size, entry), idx|
+          latency = if entry.first
+                      Time.now.to_f - Sidekiq.load_json(entry)['enqueued_at']
+                    else
+                      0
+                    end
+
+          metrics[@queues[idx]] = {
+            backlog: size,
+            latency: latency.to_i
+          }
+        end
+      end
+
+      metrics
+    end
+
+    def all_process_metrics
+      Sidekiq.redis do |conn|
+        results = conn.pipelined do
+          @procs.each do |key|
+            conn.hmget(key, 'info', 'busy')
+          end
+        end
+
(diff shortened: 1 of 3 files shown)


mperham:
Some people have lots of queues/machines/processes so collecting this data can be inefficient if not done carefully.  Your current implementation requires a lot of round trips to Redis.  Look into using `conn.pipeline` for much better performance.

mrsimo:
Right! I went with whatever I found in the `sidekiq/api`. Any thoughts on the format of the response? I'll get my hands dirty this weekend and try to get it all nice and fast.

mrsimo:
I pushed a couple of commits.

* It gathers everything in three roundtrips. We could reduce it to two, but I thought the tradeoffs weren't worth it (the code would get quite messy).
* There's a bunch of new fields for the `processes`, since I just return the parsed info. It's nice information to see anyway.

Cheers!

lib/sidekiq/api.rb
```diff
@@ -176,6 +176,69 @@ def date_stat_hash(stat)
...
+      Sidekiq.redis do |conn|
+        results = conn.pipelined do
+          @queues.each do |queue_name|
+            rname = "queue:#{queue_name}"
+            conn.llen(rname)
+            conn.lrange(rname, -1, -1)
+          end
+        end
+
+        results.each_slice(2).with_index do |(size, entry), idx|
```
mperham: You need to handle the case where a queue disappears.  The name that you got in initialize might not exist anymore.

lib/sidekiq/api.rb
```diff
@@ -176,6 +176,69 @@ def date_stat_hash(stat)
...
+
+    def all_process_metrics
+      Sidekiq.redis do |conn|
+        results = conn.pipelined do
+          @procs.each do |key|
+            conn.hmget(key, 'info', 'busy')
+          end
+        end
+
+        results.map do |json_info, busy|
```
mperham: Ditto for processes.  A process name that you got in initialize might not exist anymore.

mperham:
Really nice work.  Code looks great.

mrsimo:
Pushed another commit! I used `exists` to figure out if the key still exists.

I never thought about that case, since both methods were intended to be called pretty much immediately after being instantiated, but it's definitely something that could happen.

Do you think it makes sense to fetch the queues and processes when we call `all_queue_metrics` or `all_process_metrics` instead of at initialization? That would take us to 4 roundtrips, but would let people instantiate a Monitor and keep using it to get up-to-date information.

I'll implement that really quick and we can always revert it if you disagree!

mperham:
I think you might be misunderstanding how Redis and pipeline works.  You don't need to use `exists`.  `pipeline` is not atomic so sending two operations means that the first one can succeed and the second one fail:

```ruby
results = conn.pipeline do
  conn.llen(q)
  # queue is deleted here
  conn.lrange(q)
end

results # -> [100, []]
```

You'll need to handle the empty case for every piece of data you fetch from Redis.  Exists is not the solution.

mrsimo:
Sorry Mike, you're completely right. I somehow thought if pipelines as some kind of transaction. The queue metrics were already resilient, I fixed the process metrics to be properly resilient. I'm just checking for `json_info` to have some value, since both `json_info` and `busy` come from the same `hmget`, that should be enough, right?

Thank you for your patience and guidance!

mperham:
How would you feel about making this a 3rd party gem?  There's no reason you can't add this functionality today as the Web UI is extensible, e.g. here's how Sidetiq does it:

https://github.com/tobiassvn/sidetiq/blob/master/lib/sidetiq/web.rb#L92

mrsimo:
I was going to open source this (we use it in gem version) together with our sensu plugins, but decided to try with main sidekiq first, since it seemed like a nice functionality to have and the implementation was very very simple (at least the original version).

If you think it's better I'll go that way!

Before closing this entirely, would you feel there's some opportunity to add some more stats endpoints (like the `/stats` and `/stats/queues` ones)? I'd be more than happy to help with that if you like.

Thank you for the consideration anyway!

mperham:
I'm worried about an endless variety of endpoints, each with slightly different sets of data.  I'd rather not add another endpoint beyond /stats and /stats/queues.  Since it is easy for 3rd party gems to add endpoints to the web ui, I'd rather you take that path.

mrsimo:
Hi there!

Finally [open sourced this in gem format](https://github.com/harvesthq/sidekiq-monitor-stats). There's also a [set of sensu scripts](https://github.com/sensu/sensu-community-plugins/tree/master/plugins/sidekiq) in the community plugins that work well with that gem. I've added it to https://github.com/mperham/sidekiq/wiki/Related-Projects, but wanted to let you know.

Again, thanks for your help!

PR #2257, opened 2015-03-24
Centered text in footer navbar [tablet view]

davydovanton:
Hello!
I found ui bug with text in footer navbar. If you open sidekiq web page on ipad or other tablet you can see something like as:
screenshot 2015-03-24 23 45 57

In my branch this looks like this:
screenshot 2015-03-24 23 45 41

_**P.S.: also I want to suggest hide header navbar on mobile or tablet screen size as well as in the [bootstrap](http://getbootstrap.com/components/#navbar)**_
screenshot 2015-03-24 23 56 07

### DIFF
--- a/web/assets/stylesheets/application.css
+++ b/web/assets/stylesheets/application.css
@@ -644,6 +644,7 @@ div.interval-slider input {
   .navbar-text {
     float:none;
     line-height: 30px;
+    margin: 15px auto;
   }
 }
 


davydovanton:
@mperham thanks!

mperham:
Thank you, good eye for detail.

mperham:
I don't use the mobile view much so if there are other improvements, please send them with before and after screenshots.

davydovanton:
yes, of course! And what do you think about header navbar?

mperham:
I don't understand.  What's the issue?

PR #2263, opened 2015-03-25
Update application.css

x235:
移除googlefonts引用，天朝政府我艹尼玛

### DIFF
--- a/web/assets/stylesheets/application.css
+++ b/web/assets/stylesheets/application.css
@@ -1,6 +1,3 @@
-@import url(//fonts.googleapis.com/css?family=Gudea:400,700);
-@import url(//fonts.googleapis.com/css?family=Armata);
-
 * {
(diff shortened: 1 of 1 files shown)


aprescott:
Instead of removing these fonts altogether, perhaps they could be downloaded and stored locally?

mperham:
#2144 I thought we had already removed Google fonts for Chinese users?

aprescott:
Oh, weird. They're not in `master`. I have no idea why git/github would show a diff for this commit...

mperham:
It's not in master: https://github.com/mperham/sidekiq/blob/master/web/assets/stylesheets/application.css

We do need to remove the Google font names from font-face.

mperham:
@x235 I think you need to pull the latest master.  This was removed in 3.3.1.

PR #2271, opened 2015-03-30
Use addressable gem for support with URLs with underscores

msaffitz:
The built-in URI class for ruby doesn't properly handle hostnames with underscores in them. See bug at:

https://bugs.ruby-lang.org/issues/8241

This updates to use the addressable gem, which provides better support for current URI standards.

### DIFF
--- a/lib/sidekiq/redis_connection.rb
+++ b/lib/sidekiq/redis_connection.rb
@@ -1,6 +1,6 @@
 require 'connection_pool'
 require 'redis'
-require 'uri'
+require 'addressable/uri'
 
 module Sidekiq
   class RedisConnection
@@ -57,7 +57,7 @@ module Sidekiq
         # Don't log Redis AUTH password
         redacted = "REDACTED"
         scrubbed_options = options.dup
-        if scrubbed_options[:url] && (uri = URI.parse(scrubbed_options[:url])) && uri.password
+        if scrubbed_options[:url] && (uri = Addressable::URI.parse(scrubbed_options[:url])) && uri.password
           uri.password = redacted
           scrubbed_options[:url] = uri.to_s
         end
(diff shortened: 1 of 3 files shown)


mperham:
I've never heard of hostnames allowing underscore. http://en.wikipedia.org/wiki/Hostname#Restrictions%5Fon%5Fvalid%5Fhost%5Fnames

msaffitz:
FWIW this behavior is changing in Ruby 2.2:

```
$ ruby -v                                                                                                                                                     [2.1.3]
ruby 2.1.3p242 (2014-09-19 revision 47630) [x86_64-darwin14.0]
$ irb                                                                                                                                                         [2.1.3]
irb(main):001:0> require 'uri'
=> true
irb(main):002:0> URI.parse("http://example_1")
URI::InvalidURIError: the scheme http does not accept registry part: example_1 (or bad hostname?)


$ rbenv local 2.2.1                                                                                                                                           $ irb                                                                                                                                                         [2.2.1]
irb(main):001:0> require 'uri'
=> true
irb(main):002:0> URI.parse("http://example_1")
=> #<URI::HTTP http://example_1>
```

The change was prompted by how redis-rb handles sentinels.  It wants the sentinel name as part of the url, e.g.:  `url: redis://my_sentinel`  (see https://github.com/redis/redis-rb#sentinel-support)

I've made an equivalent PR over there too: https://github.com/redis/redis-rb/pull/521

mperham:
Why not rename it to "my-sentinel"?

msaffitz:
We use the Redis Chef cookbook, which unfortunately has that as a default (https://github.com/brianbianco/redisio/blob/master/providers/sentinel.rb#L47).  I could have chased down that side of the rabbit hole instead, but given that using an underscore does actually work (albeit a poor choice) this path seemed more prudent.

seuros:
mysentinel should work then

Issue #2274, opened 2015-03-30
Bootstrap navbar bug in web UI


davydovanton:
I think you know about bootstrap navbar bug. Navbar can breaks when container contain too much elements. If you will install more than two sidekiq plugins you can see this bug.
![screenshot 2015-03-31 00 55 18](https://cloud.githubusercontent.com/assets/1147484/6908047/85e752c6-d743-11e4-9d18-11576c484cca.jpg)

Also you can see this behavior on other sites which using bootstrap navbar too, far examples:
![screenshot 2015-03-31 00 58 42](https://cloud.githubusercontent.com/assets/1147484/6908054/8d8158ce-d743-11e4-97ed-5b2402109890.jpg)
![screenshot 2015-03-31 00 59 06](https://cloud.githubusercontent.com/assets/1147484/6908055/8d879536-d743-11e4-8d9b-51475892539b.jpg)

I want to suggest two ideas for solving this problem:
1. Minimize the navbar at large width (currently 768px);
2. Make the dropdown button, which contain all plugin links;

Who has any ideas how fix this?

seuros:
:+1:  for 2

PR #2277, opened 2015-03-31
Fixed protruding links in header navbar

davydovanton:
*Completed #2274 issue.*

1. All protruding plugin links rolled to dropdown button
2. Trim navbar link weight for screen less than 979px

**Added *bootstrap dropdown.js* plugin for opening and hiding dropdown button**

*P.S.: I didn't change 'Dashbord' to 'Home' because it isn't necessary*

### Large screen
screenshot 2015-04-03 01 00 09

### Small screen
screenshot 2015-04-03 01 06 38

### Large screen
screenshot 2015-04-03 01 02 17

### Small screen
screenshot 2015-04-03 01 05 58


mperham:
Not sure I like this particular solution.  Most people have one plugin so this adds an extra click to the UI for most people without any benefit.  Maybe only add the dropdown for 2+ plugins?  Is it possible to make the UI responsive so it becomes a dropdown only when there are too many tabs to fit the width?

davydovanton:
> Is it possible to make the UI responsive so it becomes a dropdown only when there are too many tabs to fit the width?

I'm not sure that I can create this without any js hacks :confused:

davydovanton:
@mperham could you explain why do it [check](https://github.com/mperham/sidekiq/pull/2277/files#diff-7123f67bfb91cbf5f0f0b409724a8152L24) if `url == ''` ? It seems to me it's excess condition.

Also I changed navbar behaviour. Now if you have 2+ plugins you will see dropdown button, else you will see link to plugin page.

mperham:
@davydovanton Which line of code?  Where?

davydovanton:
Oh, I'm sorry, I a little hastened to conclusions :sweat_smile:. All is ok :thumbsup:

mperham:
What if we changed "Dashboard" to "Home"?  Would that free up a little more space so we could possible raise the break point from > 1 to > 2 plugins?

davydovanton:
@mperham good idea!
But I want to try write js code which will replace navbar links on dropdown button if screen width less than navbar `<ul>` block. That is why I changed PR status on **wip** :smiley:

davydovanton:
I created simple prototype. You can see how it works on gif bellow:
e6qq6tlers

I hope I will complete this PS tomorrow

davydovanton:
@mperham I completed this PR :tada: Also I updated PR description.
Please check this :grimacing:

davydovanton:
thanks! :grimacing:

PR #2287, opened 2015-04-05
Use Hash#each_key instead of Hash#keys.each

davydovanton:
`Hash#keys.each` allocates an array of keys but `Hash#each_key` iterates through the
keys without allocating a new array.


davydovanton:
thanks!

PR #2288, opened 2015-04-05
Reduced by one third allocated memory for #perform_async method 

davydovanton:
I used [memory_profiler](https://github.com/SamSaffron/memory_profiler) gem with this code:
``` ruby
MemoryProfiler.report{ 100.times { HardWorker.perform_async } }.pretty_print
```

#### Before (only for sidekiq)
```
Total allocated 22631
Total retained 886

allocated memory by gem
-----------------------------------
    318864  sidekiq

allocated objects by gem
-----------------------------------
      2416  sidekiq

retained memory by gem
-----------------------------------
      1280  sidekiq

retained objects by gem
-----------------------------------
        11  sidekiq

Allocated String Report
-----------------------------------
       307  "HardWorker"
       200  /Users/anton/work/gems/sidekiq/lib/sidekiq/client.rb:217

       300  "jid"
       200  /Users/anton/work/gems/sidekiq/lib/sidekiq/client.rb:223

       300  "enqueued_at"
       200  /Users/anton/work/gems/sidekiq/lib/sidekiq/client.rb:224

       101  "default"
         1  /Users/anton/work/gems/sidekiq/lib/sidekiq.rb:106

       100  "queue:default"
       100  /Users/anton/work/gems/sidekiq/lib/sidekiq/client.rb:196

       100  "get_sidekiq_options"
       100  /Users/anton/work/gems/sidekiq/lib/sidekiq/client.rb:215
```

#### After (only for sidekiq)
```
Total allocated 21249
Total retained 888

allocated memory by gem
-----------------------------------
    219312  sidekiq

allocated objects by gem
-----------------------------------
      1814  sidekiq

retained memory by gem
-----------------------------------
      1280  sidekiq

retained objects by gem
-----------------------------------
        11  sidekiq

Allocated String Report
-----------------------------------
       307  "HardWorker"
       200  /Users/anton/work/gems/sidekiq/lib/sidekiq/client.rb:217

       100  "queue:default"
       100  /Users/anton/work/gems/sidekiq/lib/sidekiq/client.rb:196
```

### DIFF
--- a/lib/sidekiq/client.rb
+++ b/lib/sidekiq/client.rb
@@ -211,19 +211,23 @@ module Sidekiq
       raise(ArgumentError, "Message args must be an Array") unless item['args'].is_a?(Array)
       raise(ArgumentError, "Message class must be either a Class or String representation of the class name") unless item['class'].is_a?(Class) || item['class'].is_a?(String)
 
-      if item['class'].is_a?(Class)
-        raise(ArgumentError, "Message must include a Sidekiq::Worker class, not class name: #{item['class'].ancestors.inspect}") if !item['class'].respond_to?('get_sidekiq_options')
-        normalized_item = item['class'].get_sidekiq_options.merge(item)
-        normalized_item['class'] = normalized_item['class'].to_s
-      else
-        normalized_item = Sidekiq.default_worker_options.merge(item)
-      end
+      normalized_hash(item['class'.freeze])
+        .each{ |key, value| item[key] = value if item[key].nil? }
 
-      normalized_item['queue'] = normalized_item['queue'].to_s
-      normalized_item['jid'] ||= SecureRandom.hex(12)
-      normalized_item['enqueued_at'] ||= Time.now.to_f
-      normalized_item
+      item['class'.freeze] = item['class'.freeze].to_s
+      item['queue'.freeze] = item['queue'.freeze].to_s
+      item['jid'.freeze] ||= SecureRandom.hex(12)
+      item['enqueued_at'.freeze] ||= Time.now.to_f
+      item
     end
 
+    def normalized_hash(item_class)
+      if item_class.is_a?(Class)
+        raise(ArgumentError, "Message must include a Sidekiq::Worker class, not class name: #{item_class.ancestors.inspect}") if !item_class.respond_to?('get_sidekiq_options'.freeze)
+        item_class.get_sidekiq_options
+      else
+        Sidekiq.default_worker_options
+      end
+    end
   end
 end
--- a/lib/sidekiq/redis_connection.rb
+++ b/lib/sidekiq/redis_connection.rb
@@ -7,10 +7,7 @@ module Sidekiq
     class << self
 
       def create(options={})
-        url = options[:url] || determine_redis_provider
-        if url
-          options[:url] = url
-        end
+        options[:url] ||= determine_redis_provider
 
         # need a connection for Fetcher and Retry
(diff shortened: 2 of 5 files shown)


lib/sidekiq/client.rb
```diff
@@ -193,7 +193,7 @@ def atomic_push(conn, payloads)
         q = payloads.first['queue']
         to_push = payloads.map { |entry| Sidekiq.dump_json(entry) }
         conn.sadd('queues'.freeze, q)
-        conn.lpush("queue:#{q}", to_push)
+        conn.lpush("queue:#{q}".freeze, to_push)
```
mperham: My policy is don't freeze dynamically created strings.  What do you think?

jonhyman: Yes please do not do that. We dynamically create queues.

lib/sidekiq/worker.rb
```diff
@@ -64,7 +64,7 @@ def perform_in(interval, *args)
       #      can be true, false or an integer number of lines to save, default *false*
       #   :pool - use the given Redis connection pool to push this type of job to a given shard.
       def sidekiq_options(opts={})
-        self.sidekiq_options_hash = get_sidekiq_options.merge((opts || {}).stringify_keys)
+        self.sidekiq_options_hash = get_sidekiq_options.merge((opts).stringify_keys)
```
mperham: Remove the extra parentheses.

davydovanton:
@mperham @jonhyman fixed all, thanks for catch :thumbsup:

lib/sidekiq/redis_connection.rb
```diff
@@ -7,10 +7,7 @@ class RedisConnection
     class << self
 
       def create(options={})
-        url = options[:url] || determine_redis_provider
-        if url
-          options[:url] = url
-        end
+        options[:url] ||= determine_redis_provider
```
mperham: I'm worried this will change the logic slightly.  The only difference is that it will create `:url => nil` if the ENV is not set but I don't know if the Redis client will work the same with it.

davydovanton: @mperham as I can see, `url` options uses only in [`log_info`](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/redis_connection.rb#L60-L63) method:
``` ruby
if scrubbed_options[:url] && (uri = URI.parse(scrubbed_options[:url])) && uri.password
  uri.password = redacted
  scrubbed_options[:url] = uri.to_s
end
```
And I think that's not important to set `:url => nil` in option hash or not

davydovanton: @mperham I find [this](https://github.com/redis/redis-rb/blob/master/lib/redis/client.rb#L384). You can see if `options[:url]` equal `nil` redis connect to `defaults[:url]`. This shows that my changes are correct :tada:

davydovanton:
thanks!

mperham:
Thank you for this.  This is really awesome performance tuning.

davydovanton:
I hope that this performance tuning will be useful :star2:

jonhyman:
:racehorse: :smiley:

davydovanton:
@mperham sorry, but I forgot to add this changes to `CHANGELOG` :sweat: 
May be I need update this?

mperham:
I typically only put items in the changelog that might be relevant to users (API change, possible breakage, etc).  I don't think this improvement qualifies.

PR #2296, opened 2015-04-09
Load only required locale files

### LINKED ISSUE #2295: Optimize locale string loading
We have 20+ locales now and all Strings from each file are loaded upon first access to the Web UI.  This means 1500+ strings in memory, most of which are unnecessary.  Change the locale loading so that each locale file is only loaded if someone requests that locale.

davydovanton:
Fixed #2295
Also I check `get_locale` web helper with `memory_profiler`. For this I use this code:
``` ruby
MemoryProfiler.report{ 100.times { get_locale } }.pretty_print
```

### Before
```
Total allocated 17626
Total retained 3145

allocated memory by gem
-----------------------------------
   3033634  ruby-2.2.0/lib
     12942  sidekiq
      5040  other
```

### After
```
Total allocated 1220
Total retained 182

allocated memory by gem
-----------------------------------
    143025  ruby-2.2.0/lib
     87608  sidekiq
       240  other
```

gif

### DIFF
--- a/lib/sidekiq/web_helpers.rb
+++ b/lib/sidekiq/web_helpers.rb
@@ -3,17 +3,28 @@ require 'uri'
 module Sidekiq
   # This is not a public API
   module WebHelpers
-    def strings
-      @@strings ||= begin
+    def strings(lang)
+      @@strings ||= {}
+      @@strings[lang] ||= begin
         # Allow sidekiq-web extensions to add locale paths
         # so extensions can be localized
-        settings.locales.each_with_object({}) do |path,global|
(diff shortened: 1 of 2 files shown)


mperham:
There's a bug in your impl: it won't work with several different locales. Write a test to verify you can access both English and Russian strings.

davydovanton:
@mperham I fixed cache for `strings` method and add test case where I send three request with different locales

mperham:
Nicely done!

davydovanton:
thanks for review!  :smile_cat:

Issue #2317, opened 2015-04-23
options for a more reliable Scheduled::Poller


cainlevy:
I'm coming from resque/resque-scheduler, where the poller is a single thread that is both efficient (low load on Redis) and reliable (regular and consistent poll intervals). I understand that Sidekiq runs multiple Pollers, one for each process on each machine, and has to do something to remain efficient. My concern is that it accomplishes this at the expense of reliability.

## randomness vs averages

The Poller uses [`2 * rand`](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/scheduled.rb#L66) to stagger processes against each other, and then depends on everything averaging out. But averages aren't reliable.

In a simple case, where Sidekiq's threading model is used to full advantage and an application only runs a single Poller on a single server, the process will sometimes wait up to 30 seconds before finding and running scheduled jobs. It may not happen frequently, but it will not be uncommon.

Scaling up doesn't help either; it actually makes the inevitable outliers worse. If someone ends up running many Sidekiq processes and has many Pollers, then the automatically adjusted `poll_interval` will grow to some large number (e.g. [450 seconds with 30 processes](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/scheduled.rb#L78)). At that point each process will sleep between 0 and 900 seconds, which will rarely (but inevitably) result in gaps in scheduling far greater than 30 seconds and rarely (but inevitably) result in load spikes on Redis.

## ideas

The quickest patch would be to make [`15.seconds`](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/scheduled.rb#L90) configurable without overriding Sidekiq's `ProcessSet.new.size`-based strategy. This would allow someone to pick, say, `5.seconds` and narrow the worst scheduling gaps from 30 seconds to 10 seconds on a single server. But it doesn't really solve the problem when scaling to multiple processes, and could make those inevitable load spikes on Redis more common.

Another idea might be for Sidekiq to rely entirely on the [`initial_wait`](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/scheduled.rb#L52) to stagger the Pollers, and then rely on non-random polling intervals afterwards. This would still create random scheduling gaps, and might actually be worse because those gaps would persist until something restarted.

A more robust solution might be for Sidekiq to elect a single Poller from a single process, and run it with a non-random and unadjusted polling interval. The other Pollers might continue to check in randomly and see if they need to take over scheduling duties. This is probably the coolest solution.

[...]

PR #2322, opened 2015-04-24
make the average global poller delay configurable

cainlevy:
from https://github.com/mperham/sidekiq/issues/2317

### what do?

Introduces `Sidekiq.options[:global_poll_interval_average]` and begins transitioning `Sidekiq.poll_interval=` to `Sidekiq.options[:poll_interval_average]`. The intended behavior is as follows:

* **global_poll_interval_average**: How frequently Redis should be checked by some random poller for scheduled and retriable jobs
* **poll_interval_average**: How frequently each Poller should check Redis for scheduled and retriable jobs. Will be calculated from `global_poll_interval_average` by default.

### tests

I noticed that the test coverage on Poller seems to prefer testing only the public interface. I confirmed that the logic in question is covered by the existing tests and opted not to add more specific unit coverage. Let me know if this is best practice for Sidekiq? I'm happy to go back and add something.

### questions

* Is `Sidekiq.options[:poll_interval_average]` complementary or redundant to `Sidekiq.options[:global_poll_interval_average]`?
* I'd also like to submit a change to the `poll_interval_average * 2 * rand` logic so that it calculates a random number in a tighter spread (e.g. 0.5 * average ... 1.5 * average). Separate pull request? **update:** https://github.com/mperham/sidekiq/pull/2323

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -21,7 +21,7 @@ module Sidekiq
     environment: nil,
     timeout: 8,
     poll_interval_average: nil,
-    global_poll_interval_average: 15,
+    average_scheduled_poll_interval: 15,
     error_handlers: [],
     lifecycle_events: {
       startup: [],
@@ -129,14 +129,23 @@ module Sidekiq
     Sidekiq::Logging.logger = log
   end
 
-  # When set, overrides Sidekiq.options[:global_poll_interval_average] and sets the
-  # average interval that this process will delay before checking for scheduled jobs
-  # or job retries that are ready to run.
+  # When set, overrides Sidekiq.options[:average_scheduled_poll_interval] and sets
+  # the average interval that this process will delay before checking for
+  # scheduled jobs or job retries that are ready to run.
   #
   # See sidekiq/scheduled.rb for an in-depth explanation of this value
   def self.poll_interval=(interval)
-    $stderr.puts "DEPRECATION: `Sidekiq.poll_interval = #{interval}` will be removed in Sidekiq 4. Please update to `Sidekiq.options[:global_poll_interval_average] = #{interval}`."
+    $stderr.puts "DEPRECATION: `Sidekiq.poll_interval = #{interval}` will be removed in Sidekiq 4. Please update to `Sidekiq.average_scheduled_poll_interval = #{interval}`."
     self.options[:poll_interval_average] = interval
+  end
+
+  # How frequently Redis should be checked by a random Sidekiq process for
+  # scheduled and retriable jobs. Each individual process will take turns by
+  # waiting some multiple of this value.
+  #
+  # See sidekiq/scheduled.rb for an in-depth explanation of this value
+  def self.average_scheduled_poll_interval=(interval)
+    self.options[:average_scheduled_poll_interval] = interval
   end
 
   # Register a proc to handle any error which occurs within the Sidekiq process.
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -20,6 +20,8 @@ module Sidekiq
     require: '.',
     environment: nil,
     timeout: 8,
+    poll_interval_average: nil,
+    global_poll_interval_average: 15,
     error_handlers: [],
     lifecycle_events: {
       startup: [],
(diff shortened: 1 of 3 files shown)


mperham:
Ok, just noticed your last question.  Nevermind.

mperham:
Needs a rebase.

cainlevy:
:+1:

cainlevy:
oh, i meant to observe out that if anyone is setting `Sidekiq.options[:poll_interval]` directly without using `Sidekiq.poll_interval=`, this change won't be backwards compatible. i could probably ease the transition with something like:

```
def poll_interval_average
  Sidekiq.options[:poll_interval_average] ||= Sidekiq.options[:poll_interval] || scaled_poll_interval
end
```

preferences?

mperham:
No problem.  `config.poll_interval` has been the only documented method for over a year now.  Caveat hacktor.

mperham:
What do you think about a test which verifies the random poll interval math?  Run it 100 times, average and then assert_in_delta the average and verify no outliers?

cainlevy:
squashed tests into the commit.

i ended up tuning the iterations and delta so it wouldn't throw random errors when i looped the whole test `200.times`.

test/test_scheduled.rb
```diff
@@ -82,5 +82,47 @@ def call(worker_class, message, queue, r)
...
+        begin
+          3.times do |i|
+            Sidekiq.redis do |conn|
+              conn.sadd("processes", "process-#{i}")
+              conn.hset("process-#{i}", "info", nil)
+            end
+          end
+
+          assert_equal 30, @poller.send(:scaled_poll_interval)
+        ensure
```
mperham: Redis is generally flushed after each test; no need to be so tidy IMO.

lib/sidekiq.rb
```diff
@@ -127,9 +129,13 @@ def self.logger=(log)
     Sidekiq::Logging.logger = log
   end
 
+  # When set, overrides Sidekiq.options[:global_poll_interval_average] and sets the
+  # average interval that this process will delay before checking for scheduled jobs
+  # or job retries that are ready to run.
+  #
   # See sidekiq/scheduled.rb for an in-depth explanation of this value
   def self.poll_interval=(interval)
```
mperham: This needs to print a deprecated warning.  The only thing that should be exposed in Sidekiq 4.0+ is global_poll_interval_average (which still needs a better name).

test/test_scheduled.rb
cainlevy: nice. much simpler.

cainlevy:
brainstorming names:

* future_jobs_check_delay
* average_checkin_for_dated_jobs
* scheduled_jobs_accuracy
* run_at_jobs_poll_interval

mix and match:

* run_at_jobs_accuracy
* future_jobs_poll_interval
* average_scheduled_jobs_check_delay
* ...

mperham:
Sorry, low on time right now.  Will try to get back to this soon.

cainlevy:
No rush. I probably won't get back to Sidekiq this week myself.

mperham:
Let's go with `average_scheduled_poll_interval`.

lib/sidekiq.rb
mperham: Provide a setter for the new attribute.  I don't like people accessing Sidekiq.options directly and never document it.

cainlevy: Oh sure, will do.

When I'm learning a new library, I like to see what the options are. Sidekiq has bits and pieces of advice throughout the wiki, but I struggled to find a single convenient reference that layed out all of my options ... until I opened the source and found `Sidekiq::DEFAULTS`. So `Sidekiq.options` actually feels natural to me.

If you prefer accessors, would you be interested in seeing a `Sidekiq.options = Sidekiq::Config` singleton at some later point?

mperham: Understood, totally agree.  I think the encapsulation that a setter gives is useful, e.g. when you are deprecating something and want to warn the user.  :smile:

I'd prefer not to add Sidekiq::Config at this point in time.  Sidekiq has gotten pretty stable at this point; I'd rather not refactor unless there is a functional reason.

cainlevy:
i think all feedback has been addressed here. do let me know if anything else stands out.

PR #2323, opened 2015-04-24
calculate each poll interval in a tighter range with the same average

cainlevy:
from https://github.com/mperham/sidekiq/issues/2317

This makes the Poller's randomness more reliable by reducing the worst case outliers: polling too much (unnecessary load on Redis) or too little (undesired gaps in scheduling). The statistical average should remain the same.


mperham:
How would you like me to deal with your two PRs?  Are you suggesting two different possible solutions, as they do seem to be at odds.

PR #2328, opened 2015-04-28
Footer style fix #2205

newdark:
After looking into the footer again found that it did not take responsive into account. Here is a update to that as well as a fix to make the overflow respond to scroll so even if it gets cut off the screen you can still read what is in the footer or copy and past it onto a document that has more room.

[JSfiddle link to example fix](http://jsfiddle.net/newdark/2u1kxy55/8/)

### DIFF
--- a/web/assets/stylesheets/application.css
+++ b/web/assets/stylesheets/application.css
@@ -738,14 +738,36 @@ div.interval-slider input {
   }
 }
 
-.redis-url
-{
-    max-width: 400px;
-    overflow: hidden;
+.redis-url, .redis-namespace {
+  overflow: scroll;
+  white-space: nowrap;
 }
 
-.redis-namespace
-{
-    max-width: 250px;
-    overflow: hidden;
+@media (min-width: 768px) {
+  .redis-url {
+    max-width: 200px;
+  }
+
+  .redis-namespace {
+    max-width: 150px;
+  }
+}
+
+@media (min-width: 992px) {
+  .redis-url {
+    max-width: 340px;
+  }
+
+  .redis-namespace {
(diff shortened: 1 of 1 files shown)


davydovanton:
@newdark thanks! But could you display some screenshots with bug and your solution?

newdark:
Sure @davydovanton, here is what I am seeing. I should of looked closer at this with the issue #2328 however I missed this when I tested out the changes.

screen shot 2015-04-28 at 1 41 39 pm

davydovanton:
I checked your feedle in windows chrome and this looks like as this
c users 123 desktop
Also I remember that rubygems.org have [problem](https://github.com/rubygems/rubygems.org/issues/778) with `overflow: scroll`

newdark:
how about we use hidden rather then scroll. Every Browser. I am guessing your on a linux box correct?
Only looks good with mac scroll bars
screen shot 2015-04-28 at 2 03 38 pm

[http://jsfiddle.net/newdark/2u1kxy55/9/](http://jsfiddle.net/newdark/2u1kxy55/9/)

davydovanton:
hmmw, you are right it looks better with scroll :sweat_smile:
please, revert your last commit and split all this in one :octocat:

newdark:
Are you asking me to squash all my commits into one?

davydovanton:
yes

newdark:
Done

davydovanton:
and please fix formatting (2 spaces instead 4) :purple_heart:

newdark:
Ok all better corrected the spacing

davydovanton:
thank you! :+1:

PR #2332, opened 2015-05-05
Call timeago() for new <time> tags during live pull

### LINKED ISSUE #2331: Polling removes formatted timestamps
The new polling with AJAX requests has an issue with the timestamps on a page. Normally the timestamps are replaced by a representation like "less than a minute ago". When the page is refreshed by AJAX for the first time, the timestamp representation is not changed and the original timestamp is shown.

sidekiq_pro

Sidekiq version 3.3.4

davydovanton:
Fix #2331


mperham:
:+1:

PR #2340, opened 2015-05-10
make `enqueued_at` be accessible from worker

sonots:
I wanted to know the time at enqueued from my worker codes. 
This patch allows to make `enqueued_at` be accessible from worker codes.

### DIFF
--- a/lib/sidekiq/processor.rb
+++ b/lib/sidekiq/processor.rb
@@ -46,6 +46,7 @@ module Sidekiq
         klass  = msg['class'].constantize
         worker = klass.new
         worker.jid = msg['jid']
+        worker.enqueued_at = msg['enqueued_at']
 
         stats(worker, msg, queue) do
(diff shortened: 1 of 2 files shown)


sonots:
Please let me know if there is another way to access to `enqueued_at` without patching.

mperham:
You should pass in Time.now.to_f as an arg. Workers should not access data outside of their arguments.

sonots:
Hmm, I am using https://github.com/mhenrixon/sidekiq-unique-jobs, and adding Time.now.to_f makes all args consider to be unique mistakenly.

We can already get the job-id which is the outside of args as self.jid, so treating enqueued_at simlilarly looked not so wierd.

mperham:
Read the unique jobs docs about unique_args.

sonots:
Okay, I could do it without this patch. Thanks anyway!

PR #2349, opened 2015-05-14
Create table containers with overflow is overlay

### LINKED ISSUE #2290: Update tables for mobile screen
Hello all :v:
I have the idea to update the table layouts, so that they are more adaptive. For example long tables for mobile screen looks like this:
screenshot 2015-04-07 00 00 12
screenshot 2015-04-07 00 00 22

But I want that they looks like this:
screenshot 2015-04-06 23 57 46
screenshot 2015-04-06 23 57 54

What do you think?

davydovanton:
Fix #2290 

sidekiq-tables

### DIFF
--- a/web/assets/stylesheets/application.css
+++ b/web/assets/stylesheets/application.css
@@ -165,6 +165,10 @@ header.row .pagination {
   width: 100%;
 }
 
+.table_container {
+  overflow: overlay;
+}
+
 table.table-white {
   background-color: #fff;
 }
--- a/web/views/_job_info.erb
+++ b/web/views/_job_info.erb
@@ -2,81 +2,83 @@
   <h3><%= t('Job') %></h3>
 </header>
 
-<table class="table table-bordered table-striped">
-  <tbody>
-    <tr>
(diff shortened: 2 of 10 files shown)


mperham:
What do you think about creating a table helper method to generate the div and table tags?

davydovanton:
So, idea is great but i have a problem with `capture`. Because I use simple version based on [sinatra contrib](https://github.com/sinatra/sinatra-contrib/blob/master/lib/sinatra/capture.rb). The problem is that this `capture` method not work with `<%= ... %>` well. You can check this on [documentation](https://github.com/sinatra/sinatra-contrib/blob/master/lib/sinatra/capture.rb#L43-L47). Because that, I can create helper, but it will be look like as this:

``` erb
<% table = table do %>
  <h3><%= t('Processes') %></h3>
<% end %>
<%= table %>
```

I think that it's ugly, but anyway, what do you think?

mperham:
Yeah, that's not worth it.

davydovanton:
:metal:

PR #2352, opened 2015-05-14
Validate parameters in config file

perldork:
The command line parser validates that parameters that require integer options are integers, the configuration file validator was not doing this same type of validation.

* Ensure verbose is either true or false
* Ensure integer parameters are valid integers

### DIFF
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -277,14 +277,14 @@ module Sidekiq
         unless options[option].is_a? Integer
           raise(
             ArgumentError,
-            %{Configuration parameter "#{option}": "#{options[option]}" is not a valid integer}
+            %{"#{option}": "#{options[option]}" is not a valid integer}
           )
         end
 
         if options[option] <= 0
           raise(
             ArgumentError,
-            %{Configuration parameter "#{option}": must be > 0 (current value: #{options[option]})}
+            %{"#{option}": must be > 0 (current value: #{options[option]})}
           )
         end
       end
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -270,21 +270,18 @@ module Sidekiq
         die(1)
       end
 
-      errors = []
-
       # Ensure numeric option values are numbers and fall in reasonable ranges
       [:concurrency, :timeout].each do |option|
         next unless options.has_key?(option)
 
         unless options[option].is_a? Integer
-          errors.push(%{parameter "#{option}": "#{options[option]}" is not a valid integer})
-          next
+          raise(%{Configuration file parameter "#{option}": "#{options[option]}" is not a valid integer})
         end
 
-        errors.push(%{parameter "#{option}": "#{options[option]}" must be > 0}) if options[option] <= 0
+        if options[option] <= 0
+          raise(%{Configuration file parameter "#{option}": "#{options[option]}" must be > 0})
+        end
       end
-
-      raise("Invalid values found in configuration file settings: " + errors.join(", ")) unless errors.empty?
     end
 
     def parse_options(argv)
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -275,11 +275,17 @@ module Sidekiq
         next unless options.has_key?(option)
 
         unless options[option].is_a? Integer
-          raise(%{Configuration file parameter "#{option}": "#{options[option]}" is not a valid integer})
+          raise(
+            ArgumentError,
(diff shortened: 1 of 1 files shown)


lib/sidekiq/cli.rb
```diff
@@ -269,6 +269,19 @@ def validate!
         logger.info @parser
         die(1)
       end
+
+      errors = []
+      [:concurrency, :index, :timeout].each do |arg|
+        if options.has_key? arg
+          Integer(options[arg]) rescue errors.push("#{arg.to_s} must be an integer")
```
davydovanton: Please, delete unnecessary `to_s` method. Ruby use `to_s` by default for interpolation. You can check this in irb session:
```
irb(main):001:0> "#{[1,2,3]}"
=> "[1, 2, 3]"
irb(main):002:0> "#{[1,2,3].to_s}"
=> "[1, 2, 3]"
irb(main):003:0> "#{{a: 1}}"
=> "{:a=>1}"
irb(main):004:0> "#{{a: 1}.to_s}"
=> "{:a=>1}"
```

lib/sidekiq/cli.rb
```diff
@@ -269,6 +269,19 @@ def validate!
...
+        end
+      end
+      if options.has_key? :verbose
+        val = options[:verbose]
+        errors.push("verbose must be true or false") unless val == true || val == false
```
davydovanton: Also you can use `!!val == val` code for check bool type:
```
irb(main):014:0> foo = 1
=> 1
irb(main):015:0> !!foo == foo
=> false
irb(main):016:0> foo = 'test'
=> "test"
irb(main):017:0> !!foo == foo
=> false
irb(main):018:0> foo = []
=> []
irb(main):019:0> !!foo == foo
=> false
irb(main):020:0> foo = false
=> false
irb(main):021:0> !!foo == foo
=> true
```

lib/sidekiq/cli.rb
davydovanton: And also I prefer
```ruby
errors.push unless options[arg].is_a? Integer
```
instead
``` ruby
Integer(options[arg]) rescue ...
```

@mperham what do you think?

lib/sidekiq/cli.rb
mperham: verbose is a flag given to us by optparse.  There's no need to validate it at all.

lib/sidekiq/cli.rb
mperham: I prefer not to use exceptions at all for validation.

lib/sidekiq/cli.rb
```diff
@@ -269,6 +269,19 @@ def validate!
         logger.info @parser
         die(1)
       end
+
+      errors = []
```
mperham: No need to collect all errors, just fail fast on the first error.

lib/sidekiq/cli.rb
```diff
@@ -269,6 +269,19 @@ def validate!
...
+      [:concurrency, :index, :timeout].each do |arg|
```
mperham: `index` is not a field you would pass in the config file (since it's process-specific) so you don't need to validate it.  0 is not a legal value for concurrency and timeout so you can do this:

```ruby
raise ArgumentError, "Invalid timeout value #{options[:timeout]} in #{options[:config_file]}" if options.has_key?(:timeout) && options[:timeout].to_i == 0
raise ArgumentError, "Invalid concurrency value #{options[:concurrency]} in #{options[:config_file]}" if options.has_key?(:concurrency) && options[:concurrency].to_i == 0
```

lib/sidekiq/cli.rb
perldork: Isn't it a nicer user experience to see all errors at once? For me as a user I prefer that instead of fix, fail, fix, fail ..

perldork:
Thanks for all the feedback!
* Removed exception based validation
* Cleaned up error messages to be more clear to user
* Added range validation for timeout and concurrency
* Removed :index

perldork:
Removed the :verbose flag processing as well per Mike's comments.

lib/sidekiq/cli.rb
mperham: As a developer, if you see one error for your config values, you can pretty safely extrapolate that similar values will also be invalid.  I prefer the simpler impl in this case.

syndbg: In favor of collecting errors, wouldn't a setting as :fail_on_first (default - true) solve this problem and make both sides happy?

perldork: It's only 2 parameters with 2 possible error conditions - fail fast works
for me. Patch updated.

davydovanton:
:cool:

PR #2356, opened 2015-05-17
Dynamically update UTC time in footer

davydovanton:


mperham:
I like the fact that the time in the footer reflects when the data in the page was rendered.

newdark:
:+1: This would be nice to have.

mperham:
@newdark if you turn on polling, the footer will update every two seconds.  The time is meant to represent when the page was rendered so you know how old the data is.

Sidekiq uses UTC everywhere (e.g. log output).  I don't want to show timezone-specific times in one place but not others.

newdark:
Ok that adds up. I can see how you would not want that considering it can be confusing. I just thought I might be helpful to show the user time zone. But that is much larger amount of work so I get you.

davydovanton:
Mike, what advantage in displaying the time when the page was loaded?

aprescott:
Knowing what the state of things like the retry queue was at a particular timestamp is useful in certain cases when you're debugging. By comparison, I don't need the current time on Sidekiq's UI. I have a clock for that. :)

davydovanton:
So, Mike, think I will close this PR, okay?

mperham:
Oh yes, sorry, I was on mobile for the discussion so I couldn't close it myself.  Thanks for starting the discussion.

davydovanton:
No problems :)
Thanks for feedback

PR #2360, opened 2015-05-18
exponential sleep and increase retries for retry_and_suppress_exceptions

victor-huang:
when i am testing SideKiq + Pro and Redis sentinels setup. i am experiencing "Error connecting to Redis on 127.0.0.1:6377 (Errno::ECONNREFUSED)" during a master failover. it will cause SideKiq admin to report the wrong stats due to exceptions in workers.

I have a proposed temp fix for the problem. i want to get SideKiq team's opinions on how to resolve the issue better.

### DIFF
--- a/lib/sidekiq/processor.rb
+++ b/lib/sidekiq/processor.rb
@@ -132,7 +132,7 @@ module Sidekiq
 
     # If an exception occurs in the block passed to this method, that block will be retried up to max_retries times.
     # All exceptions will be swallowed and logged.
-    def retry_and_suppress_exceptions(max_retries = 2)
+    def retry_and_suppress_exceptions(max_retries = 5)
       retry_count = 0
       begin
         yield
--- a/lib/sidekiq/processor.rb
+++ b/lib/sidekiq/processor.rb
@@ -132,7 +132,7 @@ module Sidekiq
 
     # If an exception occurs in the block passed to this method, that block will be retried up to max_retries times.
     # All exceptions will be swallowed and logged.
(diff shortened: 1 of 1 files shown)


mperham:
That's a long time to sleep. How about `sleep(retry_count)`?

victor-huang:
the time it takes depends on the Redis sentinels 'down-after-milliseconds' setting. is it possible to expose this number of retry config to server config in "Sidekiq.configure_server { ... }"?

mperham:
What's the default value?

victor-huang:
the default is 30s: https://github.com/antirez/redis/blob/unstable/sentinel.conf#L82
override the constance will throw warning in Ruby. i think having a config method for the Processor might be better.

mperham:
Extract the sleep as a method with a default of `sleep(retry_count)`.  Monkeypatch the method for your purposes.

victor-huang:
i just updated the PR. is this what you mean?
in our code we can patch sleep_retry_count to sleep longer, but we still only have 2 retires.

mperham:
You can increase the max retries to 5 too.

victor-huang:
i also updated the default retry to 5.

mperham:
Thank you!  I'll rename the method.

victor-huang:
Thanks, please let me know when you have the updated name

mperham:
https://github.com/mperham/sidekiq/commit/2fc1d81cc9691c9232a088f1410c5a27fae5e11f

PR #2375, opened 2015-06-03
Make pry-byebug available on ruby 2.2.2 too

mrsimo:
Hopefully a simple one. `pry-byebug` also works under mri 2.2.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,7 @@ platforms :ruby do
   gem 'sqlite3'
 end
 
-platforms :mri_21 do
+platforms :mri do
   gem 'pry-byebug'
 end
 


Gemfile
```diff
@@ -12,7 +12,7 @@ platforms :ruby do
   gem 'sqlite3'
 end
 
-platforms :mri_21 do
+platforms :mri_21, :mri_22 do
```
davydovanton: may be we set `mri` instead `:mri_21, :mri_22` for any MRI versions?

mrsimo: > Only supports MRI 2.0.0 or newer. For MRI 1.9.3 or older, use pry-debugger

This is only a tool for development, so I think that would be fine. Specially since sidekiq doesn't support 1.9. I'll change it!

davydovanton:
@mrsimo please squash your commits :blush:

mrsimo:
Done!

davydovanton:
@mrsimo thank you for contributing :tada:

Issue #2373, opened 2015-06-02
Latency is confusing for scheduled jobs


mrsimo:
We're monitoring our [queue's `latency`](https://github.com/mperham/sidekiq/blob/2fc1d81cc9691c9232a088f1410c5a27fae5e11f/lib/sidekiq/api.rb#L213-L219) to get notified if it gets out of control.

We recently released a feature that schedules a sidekiq job to run in 10 hours. It's working great, but has caused that our monitoring sometines reports a latency of ~3600s (10 hours).

This is some debugging info:

```
low has a latency of 36003.53266143799
entry.enqueued_at: 1433196986.023483
Time.now:          1433232989.5564206
"{\"retry\":0,\"queue\":\"low\",\"class\":\"WarnAboutRunawayTimerWorker\",\"args\":[337121779],\"jid\":\"b5acceac3b94680bc7d06448\",\"enqueued_at\":1433196986.023483}"
```

I guess the `enqueued_at` is correct in a sense, but I wonder if we could do something about it. I've been using `Sidekiq::Stats::Queue#latency` to get that number, but I'd be fine calculating it manually.

Would it make sense to change the `enqueued_at` to the time it's actually pulled from the scheduled set and into the actual queue? Maybe add another field `originally_enqueued_at` or something.

That's just an idea, anything that would let me know a more realistic latency would be welcome.

Thanks a lot!

mperham:
Agreed, that's confusing.  The attribute is poorly named and not implemented precisely.  We could rename it `created_at` and have a separate `enqueued_at` which is set only when actually pushing to a queue.

PR #2376, opened 2015-06-03
Use created_at for the time schedule jobs are scheduled

mrsimo:
Based on the discussion in https://github.com/mperham/sidekiq/issues/2373, wanted to give this a try.

When a job is scheduled, it should have a `created_at` time, and get an `enqueued_at` when it's actually enqueued back into a queue.

### DIFF
--- a/lib/sidekiq/client.rb
+++ b/lib/sidekiq/client.rb
@@ -191,7 +191,10 @@ module Sidekiq
         end)
       else
         q = payloads.first['queue']
-        to_push = payloads.map { |entry| Sidekiq.dump_json(entry) }
+        to_push = payloads.map do |entry|
+          entry['enqueued_at'.freeze] ||= Time.now.to_f
+          Sidekiq.dump_json(entry)
+        end
         conn.sadd('queues'.freeze, q)
         conn.lpush("queue:#{q}", to_push)
       end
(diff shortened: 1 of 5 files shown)


lib/sidekiq/client.rb
```diff
@@ -218,6 +219,7 @@ def normalize_item(item)
       item['queue'.freeze] = item['queue'.freeze].to_s
       item['jid'.freeze] ||= SecureRandom.hex(12)
       item['enqueued_at'.freeze] ||= Time.now.to_f
```
mperham: Should this property be set inside `atomic_push` instead when we know the job is actually being enqueued?

mperham:
Really nice work.  Add a Changes entry as this can easily break people's code.

lib/sidekiq/client.rb
mrsimo: That makes more sense! Pushed https://github.com/mrsimo/sidekiq/commit/b091d5c0542940780d9ddda765e0b8a1bd08ec02

mrsimo:
Added the entry to Changes in https://github.com/mrsimo/sidekiq/commit/af5e59a9915281b4a4d5c9768795152e9bffcb59

Started a new section for 3.3.5, since 3.3.4 is already released. Would you like the commits squashed?

mperham:
Looks great, yes please squash.

mrsimo:
Squashed!

PR #2377, opened 2015-06-04
properly stop sidekiq in upstart one app example

kroky:
Hey, I stumbled across this problem when using the upstart example for one app - it seems it is missing the pre-stop hook which actually uses sidekiqctl to stop the workers. Without it, upstart thinks it stops the service but sidekiq continues to run in the background thus piling up the sidekiq workers on each restart - e.g. on each deploy in my case. I saw that the upstart 'many apps' example script had the hook but the 'one app' example did not, thus this pull request. Hope it helps!

Also, it seems the one-app script did not use `bundle exec` to execute the sidekiq bin file which could cause problems on some setups. Fixed it too.

### DIFF
--- a/examples/upstart/manage-one/sidekiq.conf
+++ b/examples/upstart/manage-one/sidekiq.conf
@@ -53,6 +53,28 @@ exec /bin/bash <<'EOT'
   # Logs out to /var/log/upstart/sidekiq.log by default
 
   cd /var/www/app
-  exec bin/sidekiq -i ${index} -e production
+  exec bundle exec sidekiq -i ${index} -e production -P tmp/pids/sidekiq-${index}.pid
+EOT
+end script
+
+pre-stop script
+# this script runs in /bin/sh by default
+# respawn as bash so we can source in rbenv
+exec /bin/bash <<'EOT'
+  # Pick your poison :) Or none if you're using a system wide installed Ruby.
+  # rbenv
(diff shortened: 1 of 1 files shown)


mperham:
I don't think pre-stop should be necessary.  Upstart should send TERM to the exec'd process.  Is it possible you have an app issue?  sidekiqctl will `kill -9` a process that doesn't die so it's possible this is why it's working for you.

Yes on the bundle exec issue.  I fixed that for a client but forgot to roll it back into the example script.

kroky:
You are right, my sidekiq refuses to stop with a TERM signal, only `kill -9` works. However, I don't see any issue in the logs or something suggesting there is a problem. What should I check to see why sidekiq does not stop with a TERM signal?

mperham:
Look for Kernel.trap usage or any custom Threads.  Also, `rescue Exception` in the right place can do it. http://www.mikeperham.com/2012/03/03/the-perils-of-rescue-exception/

mperham:
Fixed bundle exec on master.

kroky:
Ok, thanks.

kroky:
Neither of the three possible causes are present in the code and the strange thing is that it was refusing to stop when sidekiq was idle - no jobs were being processed at that time since it is a staging env. Even stranger, now it stops with a TERM signal. So, leaving this peculiarity for a future possible debug session...

PR #2381, opened 2015-06-09
Refactor proctitle creation

mperham:
This allows other code to add elements to the proctitle.

### DIFF
--- a/lib/sidekiq/manager.rb
+++ b/lib/sidekiq/manager.rb
@@ -134,17 +134,25 @@ module Sidekiq
       @threads[proxy_id] = thr
     end
 
+    PROCTITLES = [
+      Proc.new { 'sidekiq'.freeze },
+      Proc.new { Sidekiq::VERSION },
+      Proc.new { |mgr, data| data['tag'] },
+      Proc.new { |mgr, data| "[#{mgr.busy.size} of #{data['concurrency']} busy]" },
+      Proc.new { |mgr, data| "stopping" if mgr.stopped? },
+    ]
+
     def heartbeat(key, data, json)
-      proctitle = ['sidekiq', Sidekiq::VERSION]
-      proctitle << data['tag'] unless data['tag'].empty?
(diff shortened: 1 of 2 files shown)


test/test_manager.rb
```diff
@@ -114,7 +114,7 @@ def new_manager(opts)
         end
 
         it 'sets useful info to proctitle' do
-          assert_equal "sidekiq #{Sidekiq::VERSION} myapp [1 of 3 busy]", $0
+          assert_equal "sidekiq #{Sidekiq::VERSION} myapp [1 of 3 busy] ", $0
```
aprescott: Is it worth calling `strip` on the result so this extra space doesn't show up?

mperham:
I knew someone would bring it up.  Dealt with it.  :sunglasses:

lib/sidekiq/manager.rb
```diff
@@ -134,12 +134,18 @@ def real_thread(proxy_id, thr)
       @threads[proxy_id] = thr
     end
 
+    PROCTITLES = [
+      Proc.new { 'sidekiq'.freeze },
+      Proc.new { Sidekiq::VERSION },
+      Proc.new { |mgr, data| data['tag'] },
+      Proc.new { |mgr, data| "[#{mgr.busy.size} of #{data['concurrency']} busy]" },
+      Proc.new { |mgr, data| "stopping" if mgr.stopped? },
```
davydovanton: it's style preference but may be will we use `proc` instead `Proc.new` and `x.(arg1, arg2)` instead `x.call(arg1, arg2)`?

mperham: Please do!

davydovanton: no problem :)

davydovanton:
and may be we need to comment this or update wiki?

mperham:
It's an internal API, not meant for public usage.  Not documented on purpose.

davydovanton:
okay, I understand

PR #2384, opened 2015-06-10
Use new proc syntax instead all Proc.new

davydovanton:
In addition to #2381 and #2382.
[Link](https://github.com/bbatsov/ruby-style-guide#proc) from ruby style guides.


davydovanton:
thanks :)

PR #2399, opened 2015-06-21
Prevent redundant polling in the web UI

tombenner:
Previously, the JS function `updatePage()` was being called twice when polling is enabled (when the query string is `?poll=true`), creating two `setInterval` timers that each make an HTTP request every second and causing polling updates to look jittery. The polling link was rendered twice in the nav, once for desktop-sized browsers and once for mobile-sized browsers, and the function was called once during each render.

This PR breaks the `_poll` partial into two partials, one for the link and one for the JS, so that the JS can be called only once. I've added a test and have also tested this manually for all views in the web UI. Other solutions exist, too; I'm happy to rewrite this if another solution is preferred.

### DIFF
--- a/test/test_web.rb
+++ b/test/test_web.rb
@@ -295,6 +295,12 @@ class TestWeb < Sidekiq::Test
       assert_match /#{msg['args'][2]}/, last_response.body
     end
 
+    it 'calls updatePage() once when polling' do
+      get '/busy?poll=true'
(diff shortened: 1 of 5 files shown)


seuros:
LGTM :+1:

davydovanton:
:shipit:

davydovanton:
thanks for contribution :+1:

Issue #2411, opened 2015-06-29
Floating footer over content of page


edwinv:
I've reported this issue before by mail, but after the last release the issue is present again. The floating footer in de Sidekiq web UI is always hovering over the bottom part of the page. This makes a lot of actions unavailable because the buttons are always on the bottom of the page.

I'd love to see a quick fix and release for this bug, the Sidekiq web interface is mostly useless with this bug.

Screenshot with page scrolled all the way down, no buttons in screen:

![sidekiq_pro_and_new_issue_ _mperham_sidekiq](https://cloud.githubusercontent.com/assets/563294/8404030/2de046c0-1e4a-11e5-9506-5854451ea1ce.png)

davydovanton:
@edwinv hello!
I understand right that you have long redis db url?

edwinv:
@davydovanton 70 characters long. We use an Amazon AWS internal hostname to connect to Redis.

davydovanton:
@edwinv and which version of sidekiq are you use?

edwinv:
```
Sidekiq v3.4.1 / Sidekiq Pro v2.0.3
```

mperham:
Can you try master?  I believe we fixed it to use `overflow: hidden` recently.

edwinv:
Master doesn't solve the issue. The issue seems to be a `max-width` in a responsive version:

```css
@media (min-width: 1200px) {
  .redis-url {
    max-width: 540px;
  }
}
```

540px is too wide, when I lower this to 480px the issue is solved. Not tested for the other media queries.

edwinv:
Fix confirmed, thanks for the prompt action!

PR #2393, opened 2015-06-18
Web: Revert overflow for `redis-url` and `redis-namespace` to `hidden`

warrenguy:
Looks like this was mistakenly? changed to `scroll` in 542b995

PR #2412, opened 2015-06-29
Hide redis url if url length more than 40 characters

davydovanton:
Refer to #2411 
I think that best way to display redis url is hide this url if it's length more than 40 (or may be more?) characters.
Therefore I created toggle button with redis url and `(show)` text (also may be is better to translate this text for each locale file?).

### gif:
hfmvk16xaq

### DIFF
--- a/web/views/_footer.erb
+++ b/web/views/_footer.erb
@@ -6,7 +6,11 @@
             <p class="navbar-text" style="color:white;"><%= product_version %></p>
           </li>
           <li>
-            <p class="navbar-text redis-url">Redis: <%= redis_connection %></p>
+            <p class="navbar-text redis-url">
+              Redis:
+              <button class="redis-url__text"><%= redis_connection %></button>
+              <button class="redis-url__show">(show)</button>
+            </p>
           </li>
           <li>
             <p class="navbar-text"><%= t('Time') %>: <%= Time.now.utc.strftime('%H:%M:%S UTC') %></p>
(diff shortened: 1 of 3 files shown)


seuros:
:+1:

web/views/_footer.erb
```diff
@@ -6,7 +6,11 @@
             <p class="navbar-text" style="color:white;"><%= product_version %></p>
           </li>
           <li>
-            <p class="navbar-text redis-url">Redis: <%= redis_connection %></p>
+            <p class="navbar-text redis-url">
+              Redis:
+              <button class="redis-url__text"><%= redis_connection %></button>
+              <button class="redis-url__show">(show)</button>
```
seuros: what about adding a label class ?
http://getbootstrap.com/components/#labels

davydovanton: I don't knowю I think that it's unnecessary because it's button and label needs only for text information (status, etc).

mperham:
I would rather the hostname be hidden with CSS if it's too long. I don't want to solution that involves JavaScript.

davydovanton:
@mperham if you want use pure css I can add `overflow` for redis url. But this include some problems. For example you can check this PR #2393 where you can see what will happen for window users.
But anyway I think that long redis url it's big problem for web UI.

mperham:
I don't understand the problem of using `overflow: hidden`.  Seems like that's the best solution.

mperham:
PR #2393 switches back to `overflow: hidden` since 3.4.0 shipped with `overflow: scroll`.  I'm wondering if master is now correct again and fixes this issue.

davydovanton:
@mperham okay, I close it

PR #2416, opened 2015-07-01
Stringify keys

asymmetric:
Convert hash keys to strings in the following scenario:

```ruby
Sidekiq.push(class: 'MyClass', args: [])
```

Note that the parameter to `push` is a `Hash` whose keys are `Symbol`s, not `String`s. Sidekiq only checks for `String`s, so we convert them.

This happened to me and took me a while to figure out what was wrong.


mperham:
The API has only accepted Strings from day one.  I don't want it to accept symbols.  Perhaps the error message should mention strings only?

aprescott:
Early on, I found this sort of bug very confusing and would've appreciated clearer error messages telling me to make sure I use strings. I've also come to appreciate the simplicity of having a very consistent interface, as opposed to accepting both strings and symbols.

PR #2422, opened 2015-07-06
CSRF vulnerability in Web UI

mperham:
Form POSTs don't verify an authenticity token so it's possible to do bad stuff.  Fix incoming...

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -1,7 +1,14 @@
-HEAD
+3.4.2
 -----------
 
 - Safer display of job data in Web UI [#2405]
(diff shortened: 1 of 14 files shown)


mperham:
This will break any 3rd party Web UI extensions which try to process POSTs.  They will need to update their forms to include the same hidden token input.

davydovanton:
wow, nice changes! :tada: 
But I have a one question: why `#tkn` instead `#security_token`? I think that this name is more clear.

mperham:
No reason.  Actually I should have pulled the entire line out: `<%= csrf_tag %>`

davydovanton:
okay, I understand, thanks :)

Issue #2401, opened 2015-06-22
Upstart script gets executed without passing the index variable


robinboening:
Hi there,

I have been using this upstart script (https://github.com/mperham/sidekiq/blob/master/examples/upstart/manage-one/sidekiq.conf) for a while. Fortunately there was no situation it was really needed until the day the server rebooted. Unfortunately the upstart script had no effect and sidekiq was not started.

After some investigations I think I know why this happend:
The Ubuntu system probably starts the script using this command `initctl start sidekiq`. The script expects a variable named `index`, but it is not passed when the system calls it on a reboot. So the script fails without starting sidekiq.

I fixed the issue for me by just hardcoding the index in the script, but certainly this is not a general and great solution.

Any ideas or suggestions?

Thanks
Robin

mperham:
The `workers` script is supposed to start up everything.  In fact, the sidekiq script should not have the starts on or stops on line since, like you mention, it can't be auto-started.

https://github.com/mperham/sidekiq/blob/master/examples/upstart/manage-one/workers.conf#L18

robinboening:
I think its worth mentioning that the upstart scripts are not built for this purpose.

Or maybe it can by reading environment vars?

mperham:
They are built for exactly that purpose.  It's a bug, I'll fix it and you are welcome to test it.

robinboening:
Ohh, I think now I got the purpose of the `workers.conf` and the relation to the `sidekiq.conf`. I even introduced that bug weeks ago with my PR. Sorry about that!

PR #2425, opened 2015-07-08
shutdown all upstart instances on 'stop workers'

mpugach:


seuros:
Thanks

mpugach:
that was fast )

seuros:
It was a performed async merge.

mperham:
There's a reason we didn't do this. Check history

mpugach:
Quick search gives me #2377 but it's not quite related. Files are different. Can you point me with link?

mperham:
I was thinking #2401; this change looks ok.

JamesChevalier:
Thanks for your work on this - I can't wait to use it!

I have a quick question, though... How do you `quiet` the sidekiq instances if you have `env NUM_WORKERS=2` in the `workers.conf` file? (My interest is specifically from within capistrano, but that detail might not be important.)

I found [this documentation](https://github.com/mperham/sidekiq/wiki/Deploying-to-Ubuntu#using-capistrano) that suggests this should work from within capistrano:

    puts capture("kill -USR1 $(sudo initctl status workers | grep /running | awk '{print $NF}') || :") 

I was trying it out manually, first, and I found that `sudo initctl status workers | grep /running | awk '{print $NF}'` would not print out PIDs, it would simply print out 'start/running'.

I can use `stop workers` instead, but that seems dangerous. I think the suggested practice is to send the `USR1` signal as early in the deploy as possible, and the restart as close to the end as possible.

mperham:
`killall -USR1 sidekiq`?

JamesChevalier:
`killall -USR1 sidekiq` doesn't work:
`sidekiq: no process found`

My guess is that it's because the processes have different names:

>$ ps aux | grep sidekiq
>... sidekiq 3.4.1 production [0 of 25 busy]
>... sidekiq 3.4.1 production [0 of 25 busy]

Based on some googling, it looks like this might work:
`pgrep -f 'sidekiq' | xargs kill -USR1`

PR #2426, opened 2015-07-09
Allow for the creation directories when indicating a pidfile

frankywahl:
This is useful to avoid failing if indicating a PID file in a folder.


mperham:
Sidekiq does not create directories. It should fail.

frankywahl:
Thanks for the prompt reply

PR #2427, opened 2015-07-10
Adding test case for retry => 0

gleseur:
We want to use retry => 0 for tasks that we don't want to retry but still should go to the dead queue (monitoring these, possibly manually restarting if needed / acceptable)

Good thing is that it's already supported. So I'm just adding a test case for that.


seuros:
LGTM

PR #2435, opened 2015-07-14
Refactor: use iterator instead while loop in API#date_stat_hash

davydovanton:
Also I deleted unnecessary `self`


seuros:
While loop is much faster then the iterator. https://www.ruby-forum.com/topic/4419684

davydovanton:
Interesting. Thanks @seuros!

PR #2436, opened 2015-07-15
Add a flag to disable errors until retries are exhausted

zanker:
A common pattern we run into, is that we want to suppress some jobs errors until they have exhausted retries. For example, a job that makes HTTP calls to another service and logs events. It's not helpful to see 1-2 exceptions for transient HTTP errors when it just retries successfully.

This adds a flag that disables raising until retries are exhausted, which is disabled by default to preserve existing functionality.

### DIFF
--- a/lib/sidekiq/middleware/server/retry_jobs.rb
+++ b/lib/sidekiq/middleware/server/retry_jobs.rb
@@ -133,7 +133,9 @@ module Sidekiq
             retries_exhausted(worker, msg)
           end
 
-          raise exception
+          if !msg['raise_on_dead'] || count >= max_retry_attempts
+            raise exception
+          end
         end
 
         def retries_exhausted(worker, msg)
--- a/test/test_retry.rb
+++ b/test/test_retry.rb
@@ -24,6 +24,26 @@ class TestRetry < Sidekiq::Test
       end
     end
 
+    it 'allows disabling raise until exhaustion' do
+      @redis.expect :zadd, 1, ['retry', String, String]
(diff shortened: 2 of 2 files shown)


lib/sidekiq/middleware/server/retry_jobs.rb
```diff
@@ -133,7 +133,9 @@ def attempt_retry(worker, msg, queue, exception)
             retries_exhausted(worker, msg)
           end
 
-          raise exception
+          if !msg['raise_on_dead'] || count >= max_retry_attempts
```
zanker: There's probably a better flag name. I couldn't think of one short of `raise_on_death`, or maybe `exception_on_death`?

seuros:
```
exhaustion_raise
raise_on_exhaustion
silent_until_exhaustion
```

jkingdon:
Seems like a useful feature (especially if separating out transient errors by the exception class seems too hard, or if max_retry_attempts is small enough that exhaustion will be pretty soon anyway).

mperham:
https://github.com/mperham/sidekiq/wiki/Error-Handling#no-more-bike-shedding

I don't want to add any more knobs which complexify the retry system.  Fork it and tweak.  Create a 3rd party gem with a "better" retry system.  What's here today is already pretty complex and supports 95% of users.

zanker:
Oh now I remember why I just monkeypatched it originally :P 

There's your answer @xaviershay

No problem Mike

xaviershay:
Makes sense.

PR #2454, opened 2015-07-24
Grab the "created at" time as the "enqueued" time for scheduled jobs,…

jonhyman:
… since enqueued_at only gets set when jobs are put into queues, not when they are scheduled (from #2376).

---

After upgrading to the latest gems, doing

```ruby
MyWorker.perform_in(10.minutes)
```

looks like it was enqueued 46 years ago in the dashboard:

image


jonhyman:
This isn't the right solution, let me repatch.

PR #2455, opened 2015-07-24
Grab the "created at" time as the "enqueued" time for scheduled jobs,…

jonhyman:
… since enqueued_at only gets set when jobs are put into queues, not when they are scheduled (from #2376).

---

After upgrading to the latest gems, doing

```ruby
MyWorker.perform_in(10.minutes)
```

looks like it was enqueued 46 years ago in the dashboard:

image

---

You can also see it causing problems here:
```
2.2.1 :001 > s = Sidekiq::ScheduledSet.new
2.2.1 :002 > s.to_a.map {|x| x.enqueued_at}
 => [2015-07-23 20:50:03 UTC, 2015-07-17 21:55:03 UTC, 1970-01-01 00:00:00 UTC, 1970-01-01 00:00:00 UTC, 2015-06-08 21:26:34 UTC, 2015-06-19 20:16:43 UTC, 2015-06-22 15:04:19 UTC]
```

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -312,7 +312,7 @@ module Sidekiq
     end
 
     def enqueued_at
-      Time.at(@item['enqueued_at'] || 0).utc
+      Time.at(@item['enqueued_at'] || @item['created_at'] || 0).utc
     end
 
     def created_at


davydovanton:
nice :+1:

jonhyman:
@mperham you good on this?

mperham:
My initial impression is :-1: because it's not true. I'd prefer the UI to explicitly note the job hasn't been enqueued yet.

jonhyman:
Thoughts on me changing the method to be

```ruby
def enqueued_at
  @item['enqueued_at'] ? Time.at(@item['enqueued_at']).utc : nil
end
```

that way it's clear that it's not enqueued. I can also make the web layer say "Not yet enqueued"

mperham:
I'd be happier with that.  Better reflects reality.

jonhyman:
Cool. Let me close this out and re-open one shortly.

jonhyman:
https://github.com/mperham/sidekiq/pull/2488

PR #2462, opened 2015-07-28
Enable sessions so Rack::Protection will work

### LINKED ISSUE #2461: Sidekiq 3.4.2 produces Rack::Protection::AuthenticityToken Error when mounting in Padrino
[Example Repository](https://github.com/hbrysiewicz/padrino-sidekiq-example/tree/master)

When trying to mount Sidekiq application within Padrino, `Rack::Protection::AuthenticityToken Error` is produced. See example repository for bare bones demonstration. The exact same project works when Sidekiq is explicitly set to 3.4.1. 

Much thanks to the Padrino team and also @jc00ke on the Sidekiq team for the great communication up until now.

jc00ke:
closes #2461

### DIFF
--- a/lib/sidekiq/web.rb
+++ b/lib/sidekiq/web.rb
@@ -11,6 +11,7 @@ module Sidekiq
   class Web < Sinatra::Base
     include Sidekiq::Paginator
 
+    enable :sessions
     use Rack::Protection, :use => :authenticity_token unless ENV['RACK_ENV'] == 'test'
 
     set :root, File.expand_path(File.dirname(__FILE__) + "/../../web")


mperham:
Whoa, that's it?  Is there documentation on what that actually does?  I have a number of high security customers and I want Sidekiq to use the user/app's session store and not some ghetto one we activate.

mperham:
Well, shoot.  RTFM.

http://www.sinatrarb.com/configuration.html

jonhyman:
http://www.sinatrarb.com/configuration.html

> :sessions - enable/disable cookie based sessions
Support for encrypted, cookie-based sessions are included with Sinatra but are disabled by default. Enable them with:

> set :sessions, true
Sessions are implemented by inserting the Rack::Session::Cookie component into the application’s middleware pipeline.

@jc00oke, perhaps you want to re-open `Sidekiq::Web` if you're using the cookie store, and not do this in the base?

jonhyman:
We use a Redis-backed session store, am curious if this interferes.

mperham:
I have a very bad feeling this is going to take weeks and 3-4 patch releases to get the interaction between Rack, Sidekiq Web and the application correct.

mperham:
@hbrysiewicz Would you be able to test this PR in your app?

```
gem 'sidekiq', github: 'mperham/sidekiq', branch: 'enable-sessions-in-web-ui'
```

hbrysiewicz:
I can confirm that using the `enable-sessions-in-web-ui` branch does fix the problem @mperham

mperham:
So dank, much thanks.

mperham:
Now do we just remove all the session stuff from the wiki?

mperham:
So now that we have enabled sessions, I see this issue in the `myapp` Rails app, as reported by @codebender:

https://github.com/rails/rails/issues/15843

mperham:
This monkeypatch is now necessary for Rails apps:

https://github.com/rails/rails/issues/15843#issuecomment-125784043

:angry: :-1:

pboling:
Rails needs help here.  The Rails session is not compatible with the Rack session, yet many Rack apps like to be mounted in Rails, and would probably also like to share a rack session, but at the moment this doesn't work without a monkey patch.  This affects basically Rails vs. any Rack app (Sinatra, Grape, etc).

There hasn't been much movement on the issue, I think because it is complicated, and/or dangerous to fix.

Issue #2460, opened 2015-07-28
Incompatibility between Rack::Session and ActionDispatch::Request::Session


codebender:
It looks like I'm running into this rails bug with the new 3.4.2 CSRF vulnerability fix (https://github.com/rails/rails/issues/15843).

### My Test basic auth config
```ruby
Sidekiq::Web.use Rack::Session::Cookie, secret: 'fake secret here'
Sidekiq::Web.use(Rack::Auth::Basic) do |user, password|
  [user, password] == ['mike perham', 'for president']
end
```

Anyone have a workaround suggestion?

![image](https://cloud.githubusercontent.com/assets/987305/8938605/d917ed48-351c-11e5-95cb-efee6652afee.png)

mperham:
LOL at your test data, but no clue how to fix that.  Looks like an issue between Rack and Rails, not Sidekiq.

jonhyman:
There have been a lot of these issues lately, here's what we do and it works great on the latest version of Sidekiq and we're using Rails.

First, we use Google Auth for our Sidekiq endpoint. Basic auth sucks, you don't want to be typing in a username and password on your phone if you need to check Sidekiq on the go (and we wanted to get the benefit of 2FA behind our Sidekiq dashboard). We use the `omniauth` and `omniauth-google-oauth2`

In a Rails initializer, we have this code

```ruby
# Setup OmniAuth
Rails.application.config.middleware.use OmniAuth::Builder do
  provider :google_oauth2, configatron.oauth2.google.client_id, configatron.oauth2.google.client_secret, {
           :prompt => "select_account",
  }
end
```

```
# Setup Sidekiq
Sidekiq::Web.use Rack::Session::Redis, :redis_server => configatron.redis.session_url, :expires_in => 1.hour
Sidekiq::Web.use OmniAuth::Builder do
  provider :google_oauth2, configatron.oauth2.google.sidekiq_client_id, configatron.oauth2.google.sidekiq_client_secret, {
                             :prompt => "select_account", :name => "sidekiq_google_oauth2",
                         }
end
```

Then we wrote an extremely small auth layer:

```ruby
module Sinatra
  module GoogleOauth
    def self.registered(app)
     app.before do
        if self.session[:authenticated] || request.path_info =~ /auth/
          pass()
        end

        redirect to ('/auth/sidekiq_google_oauth2')
     end

      app.get '/auth/sidekiq_google_oauth2/callback' do
        auth = self.request.env['omniauth.auth']
[...]

PR #2464, opened 2015-07-29
Refactor session store using enumerable

davydovanton:
```
2.2.2 :005 > {a: 1, b: 2, c: 3}.each_with_object({}) { |(key, value), hash| hash[key.to_s] = value }
=> {"a"=>1, "b"=>2, "c"=>3}
```

### DIFF
--- a/myapp/config/initializers/session_store.rb
+++ b/myapp/config/initializers/session_store.rb
@@ -13,11 +13,9 @@ require 'rack/session/abstract/id'
 class Rack::Session::Abstract::SessionHash
   private
   def stringify_keys(other)
-    hash = {}
     other = other.to_hash unless other.is_a?(Hash) # hack hack hack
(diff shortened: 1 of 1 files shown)


mperham:
:-1: I want the monkeypatch to be as close to the original code as possible.

https://github.com/rack/rack/blob/master/lib/rack/session/abstract/id.rb#L149

davydovanton:
I understand, thanks.
So, I need to push PR to rack :trollface:

mperham:
Go for it.  If you could push the session fix to Rails too, that would be great.  :laughing:

davydovanton:
nop, `stringify_keys` with `#each` is most fast implementation. :sweat: 
If you're interested I can show benchmarks.

And what bug with session you have? :hushed:

mperham:
See #2460

PR #2469, opened 2015-07-30
Fix dashboard pulsing beacon alignment

antstorm:
I've noticed that the pulsing beacon was not aligned properly and was animated using properties that are causing repaint of the element (padding):

Fixed this by properly center-aligning the `ring` and the `dot`. Also the `ring` is now animated using `opacity` and `scale` that's transforming the bitmap layer without causing repaint:


seuros:
Thanks :dart:

Issue #2472, opened 2015-08-01
Sidekiq SortedEntry delete performs a redundant lookup


pik:
For example if `Sidekiq::ScheduledSet.newt.find_job(jid).delete` is run it will actually run the job-set delete after the find_job and perform an expensive operation https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/api.rb#L523 when a `Sidekiq.redis {|c| c.zrem(@parent.name, @value) }` should suffice. (Not to mention the loop doesn't break but actually runs even after a matching jid has been found). 

I'm also have a couple questions/concerns about how the Jobset delete is implemented: 

a) What is the purpose of the single-value-only scoring functionality? `conn.zremrangebyscore(name, score, score)` does anyone actually need to remove jobs at an exact time and while still being unable to remove jobs scheduled within a time-range? Does this serve an internal purpose? 

b) It gives no information about what was removed in it's output, which makes extending Sidekiq rather uncomfortable -- for example SidekiqUniqueExtension needs to hook into the `delete` commands of the api in order to prevent mutexes living after the jobs which created them have been deleted, currently it is only possible to do that by redeclaring the entire Jobset delete method, which isn't very nice at all. Can this be reworked?

mperham:
If you can provide a better impl, I'd be happy to review. It's quite possible it's not optimal.

PR #2473, opened 2015-08-02
Do not perform redundant job look-up on delete for SortedEntry

pik:
#2472  

So I've done a number of things to the SortedEntry delete. 
a) Use `zrem` right away if we already have the `@value` for the job by-passing a redundant lookup.
b) Seperated SortedEntry into delete_by_jid and delete_by_score, properly the former could be a LUA Script, but Sidekiq states support for Redis > 2.4 so I didn't do that in this PR. 

I separated the single delete into `delete_by_jid`, `delete_by_score`, and `delete_by_value`. One concern is as I mentioned that a single monolithic method that does potentially different things is hard to hook into, another plus is that unlike the old `delete` -  `delete_by_score` can take a range. 

One thing I'm really not sure about is whether `delete_by_value` should be on SortedEntry or on SortedSet, although I opted for the later because SortedEntry doesn't have access to `@_size` atm. 

An equivalent of the old `delete` is kept on SortedSet for backwards compatibility.

c) Switched to using the return value of `zrem` / `zremrangebyscore` instead of zcard, as far as I understand this should always be equivalent to setting @_size with zcard and will save an extra redis call.

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -369,11 +369,17 @@ module Sidekiq
     end
 
     def delete
-      @parent.delete(score, jid)
+      if @value
+        @parent.delete_by_value(@parent.name, @value)
+      elsif jid
+        @parent.delete_by_jid(score, jid)
+      else
+        @parent.delete_by_score(score, score)
+      end
     end
 
     def reschedule(at)
-      @parent.delete(score, jid)
+      delete
       @parent.schedule(at, item)
     end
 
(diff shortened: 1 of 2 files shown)


lib/sidekiq/api.rb
```diff
@@ -520,35 +526,41 @@ def find_job(jid)
...
-          end
-        end
-        elements_with_jid.count != 0
-      else
-        count, @_size = Sidekiq.redis do |conn|
-          conn.multi do
-            conn.zremrangebyscore(name, score, score)
-            conn.zcard name
+    def delete_by_jid(score, jid)
+      elements = Sidekiq.redis do |conn|
```
mperham: Do all of this within one `Sidekiq.redis` block.

lib/sidekiq/api.rb
```diff
@@ -520,35 +526,41 @@ def find_job(jid)
...
+        conn.zrangebyscore(name, score, score)
+      end
+      elements.each do |element|
+        message = Sidekiq.load_json(element)
+        if message["jid"] == jid
+          break Sidekiq.redis do |conn|
```
mperham: This form is hard to read.  Keep the code simple and explicit.

lib/sidekiq/api.rb
```diff
@@ -520,35 +526,41 @@ def find_job(jid)
...
+            @_size -= 1 if ret
+            ret
           end
         end
-        count != 0
+        false
       end
     end
 
+    def delete_by_score(score_min, score_max)
```
mperham: A dangerous method.  I'd prefer not to expose a method which could delete multiple jobs when the user only intended to delete one.  Make it private or inline it.

jonhyman: When I glanced at this I was going to comment the same until I realized that the current interface exposes this. My two cents would be to break the current interface and not support deleting without a jid at all for the reason you suggest.

mperham: Scumbag @mperham!

pik: @mperham 
So currently Sidekiq will delete multiple entries if they are all scheduled to run at the same-time, but since score is a float unless one explicitly schedules things to do this I imagine it's quite difficult to do by mistake - it also appears to have been an intended behaviour https://github.com/pik/sidekiq/blob/master/test%2Ftest_api.rb#L343.

I can modify this if you would prefer to remove the old behaviour entirely - but if it is there imho there is no reason for not allowing greater versatility with a min/max_score option (and perhaps a comment urging caution when using this).

mperham: @pik I agree with @jonhyman, this is a dangerous option and should be removed even if the current API allows it.  Maybe it's a one in a million chance but at scale it will happen and those are the data loss issues that are so painful to track down.

pik:
@mperham Is this missing anything else?

mperham:
Thank you, nice work!

PR #2484, opened 2015-08-12
prevent invalid concurrency from causing hard to debug problems

grosser:
cocurrency 0 makes the manager just sit there and do nothing,
which is hard to debug / understand without knowing what to look for

we were passing concurrency via an ENV variable, so it was not obvious that this was the problem,
looked into queues / redis trouble first before realizing this ... since I think there is no valid usecase for 0 / negative concurrency, this would be a nice fix to prevent this in the future ...

@mperham 
@melvinram


mperham:
Thank you!

melvinram:
:+1:

PR #2485, opened 2015-08-12
small refactor for Logging middleware

grosser:
@mperham was about to do another PR and then realized that that would not work ... this is still a good change I guess ... do you want it ?

### DIFF
--- a/lib/sidekiq/middleware/server/logging.rb
+++ b/lib/sidekiq/middleware/server/logging.rb
@@ -4,11 +4,7 @@ module Sidekiq
       class Logging
 
         def call(worker, item, queue)
-          # If we're using a wrapper class, like ActiveJob, use the "wrapped"
-          # attribute to expose the underlying thing.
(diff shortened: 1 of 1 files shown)


lib/sidekiq/middleware/server/logging.rb
```diff
@@ -21,6 +17,15 @@ def call(worker, item, queue)
           end
         end
 
+        private
+
+        # If we're using a wrapper class, like ActiveJob, use the "wrapped"
+        # attribute to expose the underlying thing.
+        def log_context(worker, item)
+          klass = item['wrapped'] || worker.class.to_s
+          "#{klass} JID-#{item['jid']}#{" BID-#{item['bid']}" if item['bid']}"
```
mperham: Can you add `.freeze` to the item attribute strings?

grosser: done

PR #2488, opened 2015-08-13
Have enqueued_at time be nil for jobs that have yet to be enqueued

jonhyman:
(i.e., are scheduled).

### DIFF
--- a/test/test_api.rb
+++ b/test/test_api.rb
@@ -207,6 +207,12 @@ class TestApi < Sidekiq::Test
       assert_equal 0, q.size
     end
 
+    it 'has no enqueued_at time for jobs enqueued in the future' do
+      job_id = ApiWorker.perform_in(100, 1, 'foo')
+      job = Sidekiq::ScheduledSet.new.find_job(job_id)
+      assert_nil job.enqueued_at
+    end
+
     it 'unwraps delayed jobs' do
       ApiWorker.delay.foo(1,2,3)
       q = Sidekiq::Queue.new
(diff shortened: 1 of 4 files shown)


test/test_api.rb
```diff
@@ -207,6 +207,13 @@ class ApiWorker
       assert_equal 0, q.size
     end
 
+    it 'has no enqueued_at time for jobs enqueued in the future' do
+      job_id = ApiWorker.perform_in(100, 1, 'foo')
+      q = Sidekiq::Queue.new
```
mperham: Superfluous line.

web/locales/en.yml
```diff
@@ -12,6 +12,7 @@ en: # <---- change this to your locale code
   Scheduled: Scheduled
   Retries: Retries
   Enqueued: Enqueued
+  NotYetEnqueued: Not yet enqueued
```
mperham: Put new locale strings at the bottom of the file.

web/views/_job_info.erb
```diff
@@ -34,7 +34,7 @@
       </tr>
       <tr>
         <th><%= t('Enqueued') %></th>
-        <td><%= relative_time(job.enqueued_at) %></td>
+        <td><%= job.enqueued_at ? relative_time(job.enqueued_at) : t('NotYetEnqueued') %></td>
```
mperham: Use a local so we don't call enqueued_at and do the time math twice.

test/test_api.rb
jonhyman: copy and paste fail. :grin: updated

web/locales/en.yml
jonhyman: sounds good

PR #2495, opened 2015-08-15
Display Process index number (provided by -i option) on Web UI

### LINKED ISSUE #2410: Display Process' index number (i.e. the number provided by -i option) on Web UI
Would be nice to know which processes might conflict with each other if the UI displayed the index and hostname next to the process name?

gaurish:
Fixes #2410

### DIFF
--- a/lib/sidekiq/launcher.rb
+++ b/lib/sidekiq/launcher.rb
@@ -76,6 +76,7 @@ module Sidekiq
         'queues' => @options[:queues].uniq,
         'labels' => Sidekiq.options[:labels],
         'identity' => identity,
+        'index' => @options[:index]
       }
       # this data doesn't change so dump it to a string
       # now so we don't need to dump it every heartbeat.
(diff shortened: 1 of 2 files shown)


davydovanton:
:+1:
@mperham what do you think? Can I merge this changes?

mperham:
Hold off please, I want to think about what to do here.

seuros:
I think we should implement just a hook here.

mperham:
You can already add the index as a tag and it will display in the UI.

gaurish:
^^in that case that we add the following line at the end of `reliable_fetch.rb` in sidekiq-pro

```
Array(Sidekiq.options[:labels]) << Sidekiq.options[:index]
```

mperham:
That is what I recommend.  Don't edit any gem source code, add it to your initializer:

```ruby
  Array(Sidekiq.options[:labels]) << Sidekiq.options[:index]
end
```

PR #2508, opened 2015-08-25
Fixed 'Getting Started' link

ideasasylum:
The Sidekiq homepage no longer(?) contains the setup process. Instead this link should go directly to the Getting Started wiki page.

Also, it's technically a 4-step process which sounds weird and more complicated than it actually is. Replaced '3-step' with 'setup'

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Installation
 Getting Started
 -----------------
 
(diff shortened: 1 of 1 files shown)


mperham:
Thank you!

PR #2517, opened 2015-08-29
Test on Ruby 2.2.3, move 2.0.0 to top of list

maclover7:


davydovanton:
I think that this is cosmetic changes :grin:
@mperham what do you think?

seuros:
2.2 should select the latest version in the 2.2.x serie. I think there is a problem with @travisci

aprescott:
@seuros that's my understanding too, about `2.2` testing the latest `2.2.x` release.

maclover7:
Just checked Travis, and it turns out you are right - specifying 2.2 will have Travis pull the latest 2.2.x release. Will close out this PR.

PR #2523, opened 2015-08-31
Add percent change to queues when polling

jcarbo:
When polling is enabled on `/queues` this adds a percent change to each queue size. It's helpful to get an at-a-glance idea for how various queues are changing or to help determine rate of change. Some future work could be adding average change size over time.

screen shot 2015-08-31 at 9 34 07 am

### DIFF
--- a/web/assets/javascripts/queues.js
+++ b/web/assets/javascripts/queues.js
@@ -1,0 +1,40 @@
+var getQueueSizes = function() {
+  return $('table.queues tbody tr').map(function(index, item) {
+    return parseInt(item.cells[1].innerHTML.match(/[0-9,]+/, '')[0].replace(',', ''), 10);
+  });
+};
+
+var updateQueueChange = function(queueSizes) {
+  $('table.queues tbody tr').map(function(index, item) {
+    var $cell = $(item.cells[1]),
+        count = parseInt(item.cells[1].innerHTML.replace(/\D/g, ''), 10),
+        diff = count - queueSizes[index],
+        percent, span;
+
+    queueSizes[index] = count;
+
+    if (diff === 0) { return; }
+
+    percent = (Math.round(diff / queueSizes[index] * 1000) / 10);
+    span  = '<span class="queue-change" style="color: ' + (diff > 0 ? '#AC1203' : '#7c9b27') + ';">';
+    span += '(' + diff + ' ' + (diff > 0 ? '&#9650; ' : '&#9660; ') + percent + '%)</span>';
+    span += '</span>';
+
+    $cell.width($cell.width());
+    item.cells[1].innerHTML = item.cells[1].innerHTML + span;
+  });
+};
+
+$(function(){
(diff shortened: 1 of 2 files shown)


web/assets/javascripts/queues.js
```diff
@@ -0,0 +1,40 @@
...
+    $cell.width($cell.width());
+    item.cells[1].innerHTML = item.cells[1].innerHTML + span;
+  });
+};
+
+$(function(){
+  var currentQueueSizes = getQueueSizes();
+
+  $(document).ajaxComplete(function(event, jqxhr, settings) {
+    if (window.location.pathname != '/sidekiq/queues') { return; }
```
davydovanton: I think that users will get some problems if they mount sidekiq web url like this:
```
mount Sidekiq::Web, at: '/admin/sidekiq-web'
```

web/assets/javascripts/queues.js
```diff
@@ -0,0 +1,40 @@
...
+        percent, span;
+
+    queueSizes[index] = count;
+
+    if (diff === 0) { return; }
+
+    percent = (Math.round(diff / queueSizes[index] * 1000) / 10);
+    span  = '<span class="queue-change" style="color: ' + (diff > 0 ? '#AC1203' : '#7c9b27') + ';">';
+    span += '(' + diff + ' ' + (diff > 0 ? '&#9650; ' : '&#9660; ') + percent + '%)</span>';
+    span += '</span>';
```
davydovanton: You duplicate `</span>`. It's normal? :confused:

web/assets/javascripts/queues.js
```diff
@@ -0,0 +1,40 @@
...
+
+    $cell.width($cell.width());
+    item.cells[1].innerHTML = item.cells[1].innerHTML + span;
```
davydovanton: may be we can use `item.cells[1].innerHTML += span;` here?

mperham:
Thank you for the nice PR.  I'm not sure I want JavaScript-based features because I have no way of testing them.  If the queues page changes its formatting, this feature will silently break.  Thoughts?

davydovanton:
@jcarbo thanks for contribution! :+1: 
Your changes looks really cool, @mperham what do you think about this feature?

ryansch:
Do we want to pull in js tests?

mperham:
JavaScript is a whole 'nother ball of wax which adds a lot more complexity to the system if we want to use it properly.  I'm going to make an executive decision and decline this PR: it's nice but the value is not worth the cost.

jcarbo:
@mperham makes sense - I didn't love this being a javascript-based solution, especially the part where it piggybacks the ajax request, but I wasn't exactly sure the best way to maintain state across polls server-side. FWIW I modeled it off of [this snippet](https://github.com/mperham/sidekiq/blob/master/web/assets/javascripts/dashboard.js#L92-L120) of `dashboard.js` which maintains the state on the client side javascript.

PR #2534, opened 2015-09-04
Add middleware stack to Testing

ryansch:
This adds a middleware stack to `Testing` that simulates the server side stack.  This is useful for all server middleware that doesn't rely on the job structure in redis.

### DIFF
--- a/lib/sidekiq/testing.rb
+++ b/lib/sidekiq/testing.rb
@@ -48,6 +48,12 @@ module Sidekiq
       def inline?
         self.__test_mode == :inline
       end
+
+      def server_middleware
+        @server_chain ||= Middleware::Chain.new
+        yield @server_chain if block_given?
+        @server_chain
+      end
     end
   end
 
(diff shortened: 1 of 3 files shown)


mperham:
:+1: Let me think about this this weekend to see if I can think of any drawbacks/side effects.

aprescott:
I could be wildly off base, but does this affect testing batches in Sidekiq's inline testing mode?

mperham:
I'm not 100% positive but Batches should work with this, including firing callbacks; all you need to do is add the server middleware to the chain in your test setup.

mperham:
It shouldn't change any testing semantics as is because the server middleware is not added by default in test mode.  You need to opt in by manually adding the middleware in your test setup.

mperham:
Add an entry to Changes and update the wiki, please.

mperham:
Solid contribution, thank you!

ryansch:
Thanks!

I'll make those changes.

PR #2540, opened 2015-09-08
Web UI keyboard shortcuts

mperham:
Could we add keyboard shortcuts to navigate the tabs in the Web UI?  Press "q" to go to Queues, "b" for Busy, etc...  Something like GitHub's keyboard shortcuts.


natematykiewicz:
@mperham, I didn't see where the "accesskey" attribute was ever used. But, I did use it to define the bindings using unobtrusive javascript. That way, you don't have to maintain what constants get looped over to define bindings. You can just add the "accesskey" attribute to any nav link.

https://github.com/natematykiewicz/sidekiq/commit/6c5e8dba2561817be69659f88e762a6cb7571cf7

natematykiewicz:
I just realized that "accesskey" is an actual HTML attribute (hence why you didn't have "data-" before it). Nonetheless, the UJS does define the Chaves bindings.

mperham:
I'm thinking this adds more complexity than it's worth.

PR #2544, opened 2015-09-09
Map retry_in period to integer

Bartuz:
According to [wiki](https://github.com/mperham/sidekiq/wiki/Error-Handling). The return value of the block must be an integer.

However many users like to use 1.hour / 2.days / 34.seconds syntax from `Activesupport::Duration`.  All Activesupport::Duration instances can be mapped to integer with `to_i` method. Lets map it than!

```
 # The current retry count is yielded. The return value of the block must be 
  # an integer. It is used as the delay, in seconds. 
  sidekiq_retry_in do |count|
    10 * (count + 1) # (i.e. 10, 20, 30, 40)
  end
```

### DIFF
--- a/lib/sidekiq/middleware/server/retry_jobs.rb
+++ b/lib/sidekiq/middleware/server/retry_jobs.rb
@@ -181,7 +181,7 @@ module Sidekiq
 
         def retry_in(worker, count)
           begin
-            worker.sidekiq_retry_in_block.call(count)
+            worker.sidekiq_retry_in_block.call(count).to_i
           rescue Exception => e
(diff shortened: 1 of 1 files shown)


mperham:
Seems legit.

Bartuz:
It needs tests in case something will change in ActiveSupport::Duration. Will work on it soon™

mperham:
It doesn't need tests.  Test your own functionality, not the functionality in other libraries.

mperham:
AS::D will always be convertible via to_i, otherwise thousands of apps will break.

Bartuz:
I'm concerned it will always have `to_i` method but was just thinking of AS::D not returning value in seconds. Like for 99.years.

Maybe I'm just paranoid :smiley:

mperham:
You're just paranoid.  That would be a huge semantic change to the API and will never happen.

PR #2545, opened 2015-09-09
Don't ack unless we attempted to start working on the job

### LINKED ISSUE #2531: Sidekiq can lose jobs even with reliable fetch
I'm on sidekiq version `2.17.8` (and sidekiq-pro `1.5.1`) and investigating an issue where sidekiq is not performing some jobs even with reliable queueing enabled. 

Currently, looks like this can happen if there was an exception in any of the middleware in the stack before it hits the `RetryJobs` middleware. Or, say, an exception in the `RetryJobs` middleware anytime before it queues up the retry (less likely).

This is because in [this block of code](https://github.com/mperham/sidekiq/blob/v2.17.8/lib/sidekiq/processor.rb#L40-L62):

```ruby
        ack = true
        begin
          msg = Sidekiq.load_json(msgstr)
          klass  = msg['class'].constantize
          worker = klass.new
          worker.jid = msg['jid']

          stats(worker, msg, queue) do
            Sidekiq.server_middleware.invoke(worker, msg, queue) do
              worker.perform(*cloned(msg['args']))
            end
          end
        rescue Sidekiq::Shutdown
          # Had to force kill this job because it didn't finish
          # within the timeout.  Don't acknowledge the work since
          # we didn't properly finish it.
          ack = false
        rescue Exception => ex
          handle_exception(ex, msg || { :message => msgstr })
          raise
        ensure
          work.acknowledge if ack
        end
```

It ack's the job even though the middleware may not have executed successfully. You can reproduce this locally by running sidekiq within foreman, sending it a lot of jobs and `ctrl-c`ing it. [This line](https://github.com/mperham/sidekiq/blob/v2.17.8/lib/sidekiq/middleware/server/logging.rb#L10) fails with `Broken pipe @ io_write - <STDERR>` and you lose the job (`perform` was never called on the worker, `RetryJobs` was never reached, so it doesn't try to retry but we still ack the work so nobody else is going to work on it). My [test worker](https://gist.github.com/sanjayprabhu/350d952f47c5a71fdae2) if you're interested.

Thoughts on how to fix this? Maybe start with `ack=false` and do something like

```ruby
            Sidekiq.server_middleware.invoke(worker, msg, queue) do
              ack = true # we attempted to perform the job
              worker.perform(*cloned(msg['args']))
            end
            ack = true # middleware invoked successfully (may or may not have decided to perform the job)
```

I can open a PR for that if you think that sounds reasonable.

The real kicker is, this codepath is not hit in production for us (`RetryJobs` is the first middleware in our chain), so there's at least one more bug which is causing us to lose jobs. I'm continuing to investigate that.

sanjayprabhu:
This fixes #2531. 

Tests required a lot of setup, but the actual fix is simple. This should have the exact same behavior as before except for the case where the middleware raises an exception before yielding. In that case, we didn't actually perform the job, so don't ack.

### DIFF
--- a/lib/sidekiq/processor.rb
+++ b/lib/sidekiq/processor.rb
@@ -40,7 +40,7 @@ module Sidekiq
 
       @boss.async.real_thread(proxy_id, Thread.current)
 
-      ack = true
+      ack = false
       begin
         msg = Sidekiq.load_json(msgstr)
         klass  = msg['class'].constantize
@@ -49,9 +49,14 @@ module Sidekiq
 
         stats(worker, msg, queue) do
           Sidekiq.server_middleware.invoke(worker, msg, queue) do
+            # Only ack if we either attempted to start this job or
+            # successfully completed it. This prevents us from
+            # losing jobs if a middleware raises an exception before yielding
+            ack = true
             execute_job(worker, cloned(msg['args']))
           end
         end
+        ack = true
       rescue Sidekiq::Shutdown
         # Had to force kill this job because it didn't finish
         # within the timeout.  Don't acknowledge the work since
--- a/test/test_processor.rb
+++ b/test/test_processor.rb
@@ -88,6 +88,105 @@ class TestProcessor < Sidekiq::Test
       assert_equal [['myarg']], msg['args']
     end
 
+    describe 'acknowledgement' do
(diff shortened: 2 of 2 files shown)


mperham:
Thanks for the hard work, this is great.

PR #2558, opened 2015-09-18
Wait after KILL signal

sebatyler:
Initializer
------------

```ruby
        config.redis = { url: ENV['REDIS_URL'] }
end
```

Versions
-----------

- Ruby : 2.1.6
- Rails : 4.1.5
- Sidekiq : 3.4.1
- OS : ubuntu 14.04 

There's no backtrace.
Before when I was deploying my rails application with capistrano-sidekiq, sidekiq didn't start.
Because a pid file existed, so sidekiq didn't start.
So I want to check process running after sending KILL signal.

### DIFF
--- a/bin/sidekiqctl
+++ b/bin/sidekiqctl
@@ -64,19 +64,18 @@ class Sidekiqctl
   end
 
   def stop
-    Process.kill(:TERM, pid)
-    kill_timeout.times do
-      begin
-        Process.getpgid(pid)
-      rescue Errno::ESRCH
-        FileUtils.rm_f pidfile
-        done 'Sidekiq shut down gracefully.'
+    %i(TERM KILL).each do |signal|
+      Process.kill(signal, pid)
+      kill_timeout.times do
+        begin
+          Process.getpgid(pid)
+        rescue Errno::ESRCH
+          FileUtils.rm_f pidfile
+          done "Sidekiq shut down #{signal == :TERM ? 'gracefully' : 'forcefully'}."
+        end
+        sleep 1
       end
-      sleep 1
     end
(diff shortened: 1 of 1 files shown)


mperham:
I don't understand what you are trying to accomplish.  Waiting for the KILL result is pointless: a process cannot respond to the signal but immediately dies.  With TERM, Sidekiq removes the pidfile.  With KILL, it cannot.

sebatyler:
Waiting for the KILL result is not pointless. The kernel may not be able to successfully kill the sidekiq process in some situations. Here is a [reference link](https://major.io/2010/03/18/sigterm-vs-sigkill/).

mperham:
Thanks for the info.  The only issue I see is that the script used to exit within `timeout` seconds.  Now it is only guaranteed to exit within `2*timeout` seconds.  Not sure if this matters or not.

mperham:
I'm still confused, why does waiting for KILL matter to whether the pidfile exists or not?

sebatyler:
I think that if the sidekiq process has not been killed yet, the pidfile needs to exist.
If the kernel may not be able to kill the sidekiq process, the pidfile may not be removed.
Because FileUtils.rm_f never raises exception, I think it is better to wait KILL.

mperham:
Sidekiq creates the pidfile when it starts and does not remove it.  Only sidekiqctl removes the pidfile.  If the kernel doesn't kill it, you'll have a zombie process hanging around but existence of the pidfile won't matter.  In fact it's better that the pidfile is deleted so the start will work as if the old Sidekiq process didn't exist.

So I still don't understand how this change improves the status quo.

PR #2564, opened 2015-09-21
Update Redis version requirement

andrewjwu:


mperham:
Why do you say Sidekiq requires 2.6?  It does not afaik.  Sidekiq Pro and Sidekiq Enterprise do.

andrewjwu:
Oh, I must have been thinking about the Pro and Enterprise case.

PR #2565, opened 2015-09-21
Put the table back.

gdpelican:


mperham:
Thanks, but we don't accept any changes to easter eggs.  It's unnecessary code churn.

PR #2588, opened 2015-10-04
Fix Norwegian Bokmål (nb) translations

theodorton:
Browsers configured to prefer content in Norwegian Bokmål send 'nb-NO,no' in the Accept-Language header.

Updated to match the English translations.


mperham:
Wouldn't this break anyone using "no" as their locale?  Is that possible?

theodorton:
Actually, I don't think it works at all, due to how "no" behaves as key in yaml.

```ruby
my_yml = "no:\n  foo: bar"
YAML.load(my_yml)
# => {false=>{"foo"=>"bar"}}
```

"no" is parsed as `false`, which is why "nb" is preferred.

mperham:
Friggin yaml.

PR #2269, opened 2015-03-26
Change page reload on ajax request in livepoll mode

davydovanton:
For more information you can see mperham/sidekiq#2266 issue

PR #2590, opened 2015-10-05
Add favicon to UI and update status when polling

natematykiewicz:
I thought it'd be nice to have a favicon for the web UI. Since the UI already shows a a green sidekiq icon when idle and a red one when active, I thought it only made sense to have the favicon reflect the current status.

While I was at it, I thought it would make sense to update both of these status indicators when polling.

I'd really prefer them to update even when not polling (for example, the dashboard updates that graph, but not the icons), but I didn't know what would be the best place to add that JS in.


mperham:
Thanks for this Nate.  I would rather have a static icon and less logic.  Can you do this instead?

natematykiewicz:
So, a static favicon. I can do that. I'll just use the one that's on sidekiq.org right now. Should I commit it to `web/assets/images`?

Do you want the "Idle"/"Active" indicator that's in the nav bar to update when "Live Polling" is enabled (which is also part of this PR)? It feels a bit misleading right now, since it only reflects how the page was when you initially loaded the page, and not the changes that took place when polling. This actually changed when "Live Polling" started using AJAX (https://github.com/mperham/sidekiq/pull/2269/files), since before it would do a full page load, which would in-turn update the status indicator.

mperham:
Yeah, it seems reasonable to update the navbar icon.

natematykiewicz:
I'm kind of new to Github. Do you get a notification when I make a commit on a PR, or should I write a comment when it's ready for feedback?

What do you think of the code, now?

mperham:
I don't get a notification so thanks for following up.  Looks good.

PR #2592, opened 2015-10-07
concurrency value should also accept string as value

justanshulsharma:
We are sidekiq pro user and we use cloudformation templates to generate our UAT environments on Aws Opsworks. Now the problem is we  pass concurrency value as integer but the json that is generated by cloudformation for opsworks treat it as string and the yml that is generated for the sidekiq is like 

```
---
:index: 0
:concurrency: '12'
:verbose: false
:timeout: 600
:queues:
- default
- queueone
- queuetwo
- queuethree
:logfile: log/sidekiq.log
```

because concurrency value is string it fails to start.

### DIFF
--- a/lib/sidekiq/manager.rb
+++ b/lib/sidekiq/manager.rb
@@ -26,7 +26,7 @@ module Sidekiq
     def initialize(condvar, options={})
       logger.debug { options.inspect }
       @options = options
-      @count = options[:concurrency] || 25
+      @count = (options[:concurrency] || 25).to_i
       raise ArgumentError, "Concurrency of #{@count} is not supported" if @count < 1
       @done_callback = nil
       @finished = condvar


lib/sidekiq/manager.rb
```diff
@@ -26,7 +26,7 @@ class Manager
     def initialize(condvar, options={})
       logger.debug { options.inspect }
       @options = options
-      @count = options[:concurrency] || 25
+      @count = options[:concurrency].to_i || 25
```
ismaelga: You probably want to do `(options[:concurrency] || 25).to_i` instead. Because `nil.to_i` is 0 and the 25 default will never be used, making it throw the error `Concurrency of 0 is not supported` if you leave concurrency without any value which is a breaking change

justanshulsharma: make sense updated @ismaelga

mperham:
Why not put `.to_i` in your YAML template?

justanshulsharma:
@mperham In our cloud formation template to create opsworks environment we pass custom json as

```
{
  "sidekiq": {
    "APP_NAME": {
      "slacker": {
        "process_count": 2,
        "config" : {
          "concurrency": 5,
          "verbose": false,
          "queues": ["critical", "default", "low"]
        }
      },
      "hard_worker": {
        "config": {
          "concurrency": 40,
          "queues": [
            ["often", 7],
            ["default", 5],
            ["seldom", 3]
          ]
        }
      }
    }
  }
}
```
and our sidekiq cookbook
```
if node[:sidekiq][application]

    workers = node[:sidekiq][application].to_hash.reject {|k,v| k.to_s =~ /restart_command|syslog/ }
    config_directory = "#{deploy[:deploy_to]}/shared/config"

    workers.each do |worker, options|

      # Convert attribute classes to plain old ruby objects
      config = options[:config] ? options[:config].to_hash : {}
      config.each do |k, v|
        case v
        when Chef::Node::ImmutableArray
          config[k] = v.to_a
        when Chef::Node::ImmutableMash
          config[k] = v.to_hash
        end
      end

      # Generate YAML string
      yaml = YAML::dump(config)

      # Convert YAML string keys to symbol keys for sidekiq while preserving
      # indentation. (queues: to :queues:)
      yaml = yaml.gsub(/^(\s*)([^:][^\s]*):/,'\1:\2:')

      (options[:process_count] || 1).times do |n|
        file "#{config_directory}/sidekiq_#{worker}#{n+1}.yml" do
          mode 0644
          action :create
          content yaml
        end
      end
    end
end
```

mperham:
I suspect the concurrency value is a String in your config hash, above the YAML::dump line.

I don't have much desire to merge this PR; I believe good systems should be strict in what they accept.  Your YAML is malformed, that's the bug that needs to be fixed.

justanshulsharma:
@mperham the reason behind that was cloudformation sends it as a string to opsworks even if it's defined as integer.

PR #2593, opened 2015-10-07
Redesign Sidekiq internals

mperham:
My recent experiment to remove Celluloid has reaped huge rewards.  This branch does several things:

1. Rewrite Sidekiq actors to use raw Threads and data structures.
2. Make Processor#stats asynchronous, which removes two Redis round trips per job.
3. Redesign job fetch so each Processor fetches its own job in parallel, this (along with (2)) makes the system much more resilient to higher Redis latency

Processing 100,000 no-op jobs with one process and 25 threads on MRI 2.2:

* Master: 125 sec
* Without Celluloid: 57 sec
* Without Celluloid and Async stats: 20 sec
* With 1ms of latency, sequential fetch: 7 min
* With 1ms of latency, parallel fetch: 20 sec

This work would be released as Sidekiq 4.0, with an ETA of Thanksgiving.


jonhyman:
:open_mouth:

lib/sidekiq/processor.rb
```diff
@@ -10,36 +8,65 @@ module Sidekiq
...
 
-    def initialize(boss)
-      @boss = boss
+    def kill(wait=false)
+      # unlike the other actors, terminate does not wait
+      # for the thread to finish because we don't know how
+      # long the job will take to finish.  Instead we
+      # provide a `kill` method to call after the shutdown
+      # timeout passes.
+      @thread.raise ::Sidekiq::Shutdown
```
ryansch: I've read in a number of places that `Thread#raise` and `Thread#kill` are unsafe.  Has that changed?

mperham: It's not safe but it's as safe as I can make it.  Processor#kill is only called when the hard shutdown timeout has passed and the process **must** exit.

ryansch: Gotcha.  The whole building is coming down anyway.

mperham:
I reimplemented fetching so that each Processor fetches its own jobs, instead of delegating fetch to a dedicated thread.  When Redis has 1ms of latency, this reduces the time needed to process 100,000 jobs from 7min to 20sec.

uhoh-itsmaciek:
Looks great. Did you by chance measure memory usage as well?

mperham:
8x less garbage generated!

lib/sidekiq/scheduled.rb
```diff
@@ -39,33 +38,57 @@ def enqueue_jobs(now=Time.now.to_f.to_s, sorted_sets=SETS)
...
+        if @thread
+          t = @thread
+          @thread = nil
+          @queue << 0
+          t.value
+        end
+      end
+
+      def start
+        @thread ||= safe_thread("scheduler") do
```
tenderlove: Is `start` meant to be thread safe? There's a read-check-write race here along with a race in `terminate`.  Also since `@thread` isn't initialized, it will cause warnings.

tenderlove: (probably should also raise an exception if `@done` is true when this method gets called)

mperham: Poller#start should only ever be called once by the main thread when calling Launcher#start, so thread safety is not necessary here.  There's definitely room for more defensive design.

tenderlove: Presumably the allocated thread will just sleep while the queue is empty, so it might be possible to just make `start` private and call it from `initialize`. Then you can eliminate the conditional here and in `terminate`. </armchairengineering>

tarcieri:
Have you considered using a `ThreadPoolExecutor`? This will give you a very fast and robust implementation on JRuby (now):

https://ruby-concurrency.github.io/concurrent-ruby/Concurrent/ThreadPoolExecutor.html

mperham:
@tarcieri I don't understand how a TPE would help vs me spinning up threads.

lib/sidekiq/scheduled.rb
mperham: I like to be able to test the object's APIs without a background thread spinning up in the test suite.

tarcieri:
@mperham on JRuby at least, it's backed by a `java.util.concurrent.ThreadPoolExecutor` and should have both lower thread coordination and memory overhead than anything that can be done in pure Ruby

mhenrixon:
Solid! Looks like you were able to delete more code than you added :+1:  Anything I need to worry about with the middleware? Couldn't see anything while going over the code.

mperham:
Most public APIs should be unchanged, including middleware.  As of today. :-)

halorgium:
@mperham very great work here. :dancer:

drewblas:
@mperham What happens if the status updator/heartbeat thread dies?  Will it have status info about jobs that finished but it never sends that info back to Redis?  It seems like the buffering of stats creates a hole for data to get lost in a crash, resulting in even more double-run or stuck jobs.

nviennot:
@mperham Glad to see that you are dropping Celluloid :) -- good move!

mperham:
@drewblas I'm not sure what you are referring to.  If you can be more explicit about a particular failure case, I can explain what should happen.

Magicdream:
@mperham 
> 8x less garbage generated!
Is it because of removed Celluloid?

mperham:
I did a quick Resque 1.25.2 benchmark for comparison.  Time to process 10,000 noop jobs:

1 process, 96 sec
25 processes, 31 sec

Extrapolating to 100,000 jobs would be ~300 seconds.  Sidekiq does the same in 20 seconds so Sidekiq's job overhead appears to be 15x less.

halorgium:
:+1:

narapon:
Excited to try this, good work!

hecbuma:
pretty nice :+1:

esbanarango:
@mperham Can we start using v4.0 now?

mperham:
@esbanarango You can start using it like any other beta software. https://www.reddit.com/r/sidekiq/comments/3qpxfn/sidekiq_400pre1_and_sidekiq_pro_300pre1_now/

PR #2602, opened 2015-10-12
Add average_queue_poll_interval option

tdtran:
The worker polls redis for the next work unit. If there is no work in the queues the worker repeats the call without any delay. This is the equivalent of busy wait loop. A large number of workers, all connecting to the same Redis, can easily kill redis performance. Enqueueing new jobs becomes painfully slow. We observe this behaviour in our production where we have 60+ workers running. To make matter worse we also use the sidekiq-limit_fetch plugin. This plugin runs Lua script embedded in Redis. This script is not fast either. With this patch we decrease the Redis load from 300K ops/s to 10K ops/s.

This patch adds a (configurable) delay to the loop. If there is no work in the queues, the worker waits a little before polling Redis again.

Sidekiq already does something like this for scheduled and retried jobs. I borrow the code from there to do the same for regular queued jobs.

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -22,6 +22,7 @@ module Sidekiq
     timeout: 8,
     poll_interval_average: nil,
     average_scheduled_poll_interval: 15,
+    average_queue_poll_interval: 5,
     error_handlers: [],
     lifecycle_events: {
       startup: [],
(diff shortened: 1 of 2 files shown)


mperham:
Sidekiq does not busy wait, it uses BRPOP. https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/fetch.rb#L102

tdtran:
Indeed the default fetch strategy `BasicFetch` does not busy wait. However custom fetch strategy (e.g. https://github.com/brainopia/sidekiq-limit_fetch/blob/master/lib/sidekiq/limit_fetch.rb) does. 

We can fix it by adding this delay to the plugin `retrieve_work` method or to Sidekiq. Sidekiq seems to be a better place because more than one plugin may have this exact same problem and this patch does not have negative effect on the default (non-busy-wait) BasicFetch.

mperham:
It does have a negative effect: an empty queue will not be checked for a few seconds.

There's a reason why Sidekiq does not implement arbitrary queue limits, a la limit_fetch: it's incredibly expensive.  The check must be implemented in Lua, which cannot use blocking operations.  Now you see one of those costs.  I'd encourage you to open an issue with that project.

PR #2604, opened 2015-10-13
Fix Styling Issues on Busy page

smolnar:
I noticed bit weird layout in Busy page and I figured out there was redundant `div` enclosing tag that was messing up the html structure. I removed it.

Also, I fixed margin on form button groups that looked bit bad because of 5px margin. It just bothered me.

### DIFF
--- a/web/assets/stylesheets/application.css
+++ b/web/assets/stylesheets/application.css
@@ -181,6 +181,10 @@ form .btn {
   margin-right: 5px;
 }
 
+form .btn-group .btn {
+  margin-right: 1px;
+}
+
 td form {
   margin-bottom: 0;
 }
--- a/web/views/busy.erb
+++ b/web/views/busy.erb
@@ -12,7 +12,6 @@
     </form>
   </div>
 </div>
-</div>
 
 <div class="table_container">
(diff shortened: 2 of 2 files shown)


mperham:
Could you show before/after screenshots?

mperham:
And please verify responsive display: mobile and tablet views should look ok too.

smolnar:
Done. Summary and other rows are not aligned (that's the redundant div tag) and Quiet All and Stop All buttons have gap between them.

I looked up `.btn-group` classes and it seems they are used only in Busy tab.

Another thing: There's a button group also on Quiet and Stop buttons in table for each process, but it's not inside of the form, so it doesn't group the buttons properly. I tried to put it inside the form, but it does not look great on mobile, though. So I omitted that change from PR from now.
https://github.com/smolnar/sidekiq/blob/busy-page-styling-bug/web/views/busy.erb#L41

Desktop before:
desktop-before

Desktop current:
desktop-current

Mobile before:
mobile-before

Mobile current:
mobile-current

mperham:
Perfect isn't necessary, better is always better.  Great work, thanks for the QA!

smolnar:
Awesome, thanks!

PR #2609, opened 2015-10-17
Create sidekiq/examples/aws/beanstalk.yml

phcoliveira:
This file is a compilation of many tutorials I'd read in order to make Sidekiq to run along the main web service process inside an instance of AWS Beanstalk.

I can only say that it is working properly, although I would like to be able to adapt the examples given for managing Sidekiq by upstart, which is present in AWS AMI 2015.09 instances.


mperham:
I'll merge this in as soon as it supports Upstart.

PR #2619, opened 2015-10-23
Make exception available in #sidekiq_retry_in

kreynolds:
Sample implementation of allowing exception to dictate retry interval behavior .. eg. for coupling with OverLimit behavior from the enterprise rate limiting primitive

### DIFF
--- a/lib/sidekiq/middleware/server/retry_jobs.rb
+++ b/lib/sidekiq/middleware/server/retry_jobs.rb
@@ -121,7 +121,7 @@ module Sidekiq
           end
 
           if count < max_retry_attempts
-            delay = delay_for(worker, count)
+            delay = delay_for(worker, count, exception)
             logger.debug { "Failure! Retry #{count} in #{delay} seconds" }
             retry_at = Time.now.to_f + delay
             payload = Sidekiq.dump_json(msg)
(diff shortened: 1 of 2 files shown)


mperham:
I think the fundamental design question is: Should Enterprise rate limit failures be considered traditional retries so we collect any backtrace, send an error notification and reschedule the job to try later?  A Sidekiq retry is a job failure where the developer needs to be notified.

My current design for the rate limiter subsystem says no, rate limit failures should not be considered retries.  The "failure" is normal and expected.  In the same way, form validation errors aren't sent to an error service.  Only if the rate limit continually fails should the job move into the retry bucket and the developer notified.

You can change the backoff algorithm here, which gets you one option you want to tune:

https://github.com/mperham/sidekiq/wiki/Ent-Rate-Limiting#back-off

kreynolds:
In the many applications I've been involved with that have used resque and sidekiq, there are a whole host of failures that are normal and expected and handled appropriately with the normal retry subsystem which would, by your definition, all require a separate middleware and configuration. Just because there is a suggested (enforced?) way of handling this one doesn't make it any different from any of those. A failure to run a job is a failure to run a job, whatever the reason may be (is OverLimit really any different than running out of DB connections, or getting a remote resource timeout, or having an API 429 you, or having an S3 upload timeout and retry, or ...). I also think its an incorrect assumption that a limit breach will always be a normal condition, and that a developer would never want to be notified about it. 

In my opinion, the solution isn't to have two very similar retry middlewares that mostly do the same thing, one that notifies a developer and one that doesn't, its to expand the functionality of the error handling/backtrace subsystem to not *have* to treat every exception as what you consider a "failure".

Regarding the back-off, I saw that but I personally find it more intuitive to define that inside the worker unless all back offs should globally treated the same (in my case that has never yet been true). With access to the job hash you can still get per-worker granularity this way, but it feels pretty strange to have that be in such different places.

mperham:
I'm ok with this change.  It's backwards compatible and local to the worker.

PR #2620, opened 2015-10-23
Add the ability to have custom retry limits based on the exception

kreynolds:
This backwards-compatible change is intended to be used with primitives such as OverLimit from the Enterprise version so that retries can be more numerous and frequent for this case.


mperham:
I've redesigned how OverLimit is handled in the middleware for the upcoming release.  No longer am I creating a new job but rather rescheduling the existing job.  I don't believe this change is necessary anymore.

PR #2641, opened 2015-11-03
testing: invoke error handlers for failing jobs

kyrylo:
I wanted to test my error handler, but couldn't do so, because none of
the error handlers were executed.


mperham:
The error handler is a Proc.  It's trivial to test on its own.

kyrylo:
Well, those were integration tests.

mperham:
If you want to integration test you shouldn't use the testing package, since that's not the real system.

PR #2645, opened 2015-11-05
Using percentage sizes for footer information with ellipses overflow.

jcavalieri:
Also added title attributes so on-hover shows full string.

### DIFF
--- a/web/assets/stylesheets/application.css
+++ b/web/assets/stylesheets/application.css
@@ -713,38 +713,37 @@ div.interval-slider input {
     position: relative;
     top: auto;
   }
-}
 
-.redis-url, .redis-namespace {
-  overflow: hidden;
-  white-space: nowrap;
+  .navbar-fixed-bottom .navbar-text {
+    max-width: 100%;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    line-height: normal;
+  }
 }
 
 @media (min-width: 768px) {
-  .redis-url {
-    max-width: 200px;
+  .navbar-fixed-bottom .navbar-text {
+    max-width: 22.5%;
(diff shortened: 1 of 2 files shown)


mperham:
I pushed this into Sidekiq 3 also, although I'm not sure I should have.

jcavalieri:
:-)

mperham:
I reverted this as it looked bad with the stock Pro / Enterprise setup.

screen shot 2015-11-06 at 09 03 50

mperham:
What do you think about this instead?

screen shot 2015-11-06 at 09 15 20

jcavalieri:
That looks good. I think the .preX added more than I was expecting.

jcavalieri:
My only concern might be when namespace is added.

mperham:
Yeah, obvious that version string is about as long as its ever going to get.

jcavalieri:
I calculated the percentages with some trial and error. It looked to me as long as the width percentages added up to 90% it seemed to not wrap.

jcavalieri:
22.5
+22.5
+27
+18
=90

so more or less can be given to each one, as all four add up to 90

jcavalieri:
Let me know if you want me to play around with it. I can mock it up on my local with the long '.preX' string.

mperham:
I'm happy with it as is now.  Let me know how it works for you if you have the time to try it out.  4.0+ only.

jcavalieri:
Looks good to me.

A couple tweaks if you want.

https://github.com/mperham/sidekiq/pull/2647

jcavalieri:
wait, one moment

jcavalieri:
Ok, PR is ready. Realized I wasn't consistent with the naming. Also added a screen shot.

PR #2647, opened 2015-11-06
footer tweaks, added title and ellipsis to url and namespace

jcavalieri:


jcavalieri:
screen shot 2015-11-06 at 3 24 34 pm

lib/sidekiq/web_helpers.rb
```diff
@@ -245,5 +245,10 @@ def environment_title_prefix
     def product_version
       "Sidekiq v#{Sidekiq::VERSION}"
     end
+
+    def redis_connection_and_namespace
+      namespace_suffix = namespace.blank? ? '' : "##{namespace}"
+      "#{redis_connection}#{namespace_suffix}"
```
mperham: Could you memoize this so its not constructed twice?

jcavalieri: Sure. I'm away from my desk right now but will do.

_______________
John Cavalieri

jcavalieri:
not sure why the CI build seems to taking a while

@mperham, I memoized the string as requested

PR #2657, opened 2015-11-13
Set redis-namespace gem as dev dependency

simpl1g:
Actually redis-namespace is still a dependency, move it dev dependency

* `redis-namespace` has been removed from Sidekiq's gem dependencies. If
  you want to use namespacing ([and I strongly urge you not to](http://www.mikeperham.com/2015/09/24/storing-data-with-redis/)), you'll need to add the gem to your Gemfile:
```ruby
gem 'redis-namespace'


seuros:
Oh thanks!

Issue #2562, opened 2015-09-19
Fake testing mode requires Worker class to exist


brandonhilkert:
Hey Mike!

I'm implementing a smaller app and looking to use Sidekiq a poor man's message bus. Ideally, I'd like the following to happen:

- Application 1 enqueues a Job on a queue the application's worker processes are not listening to
- Application 2 has a worker process that ONLY listens on that isolated queue and contains the worker class to process the job

In Application 1, I ended up adding a bare worker class to support testing:

```ruby
class SpecialWorker
  include Sidekiq::Worker
  sidekiq_options queue: :special

  def perform(connection_id)
  end
end
```

```ruby
    assert_difference 'SpecialWorker.jobs.size', 2 do
      # invoke the code that enqueues a job
    end
```

Without the job class I could use the raw client:

```ruby
        Sidekiq::Client.push(
          "queue" => "special",
          "class" => "SpecialWorker",
          "args" => [connection.id]
        )

```

However, now with the tests, I can't easily tell whether the job's been enqueued unless I rely on mocks (which i was hoping to avoid), I think because of the call to [`constantize` on the Worker class](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/testing.rb#L71), which fails because the class doesn't exist in Application 1 (`NameError: uninitialized constant SpecialWorker`).

I'm wondering if there would be any downfalls to using a hash for the stored jobs in the fake implementation and allow access to the class jobs (eg `SpecialWorker.jobs`) if it exists. Otherwise, create an alternative API to access jobs from that global testing hash. 

I'd be happy to submit a PR for it, but wanted to get your thoughts/feedback first. It's totally cool if this is too special a use case.

aprescott:
Why not define the class for the purpose of the test in app 1? Either something like RSpec's `stub_const` or simply defining `SpecialWorker` to be a constant that exists only in the test suite. What's the issue with using mocks?

brandonhilkert:
Feels tangential to value of the tests to me. Why would I add classes that
don't need to be in production to the test suite?

In general, mocking would work, but I'd rather have a direct interface to
the queues without needing to define the worker.

On Saturday, September 19, 2015, Adam Prescott <notifications@github.com>
wrote:

> Why not define the class for the purpose of the test in app 1? Either
> something like RSpec's stub_const or simply defining SpecialWorker to be
> a constant that exists only in the test suite. What's the issue with using
> mocks?
>
> —
[...]

PR #2659, opened 2015-11-13
Implements queue-based testing API

brandonhilkert:
When using the Sidekiq::Client API to push jobs on to the queue, it's
not ideal to assert the size of the queue from the perspective of a
worker because the worker may not exist in the application.

This API implements a testing API from the perspective of a queue. The
existing Worker-based testing API remains unchanged, but leverages the
job hash implemented through the Sidekiq::Queues class.

Examples:

    assert_equal 1, Sidekiq::Queues["default"].size
    assert_equal "SpecialWorker", Sidekiq::Queues["default"].first["class"]
    Sidekiq::Queues["default"].clear
    Sidekiq::Queues.clear_all

ref: https://github.com/mperham/sidekiq/issues/2562


mperham:
Aren't you calling `constantize` in the code, therefore requiring the worker class anyways?

brandonhilkert:
True fact. Fixed!

mperham:
Would you update the 4.0.0 changes with a summary and your code sample?

brandonhilkert:
This broke the `push_bulk` method when using the `inline` testing method. I need to revisit to figure out how to get `perform_one` to work with jobs that don't come from the fake queue.

Once the tests are passing, I'll add the above as suggested.

brandonhilkert:
That should fix the inline testing method. I seem to remember there's another `attr_accessor` for batches. Do you think there's any risk for that breaking with https://github.com/mperham/sidekiq/commit/19f9e05d88aacc9bd7b86b8c323aea8d41bc6dab#diff-6a5d6caa0b2fdc38b0b6ef9bc4fe40bdR78

I explicitly set `jid`, but not sure if there's more of these like that for other features I'm familiar with in the paid packages.

mperham:
I think if your change is equivalent to the old mechanism, it's fine.  We didn't set BID before so no need to set it now.

lib/sidekiq/testing.rb
```diff
@@ -68,15 +68,16 @@ class Client
...
         end
         true
       elsif Sidekiq::Testing.inline?
         payloads.each do |job|
-          job['jid'] ||= SecureRandom.hex(12)
           klass = job['class'].constantize
-          klass.jobs.unshift Sidekiq.load_json(Sidekiq.dump_json(job))
-          klass.perform_one
+          worker = klass.new
+          worker.jid = job['jid'] || SecureRandom.hex(12)
```
mperham: Weird that inline testing was pushing to the queue and then immediately popping it off.  I see we can't call perform_one anymore since we don't push to the queue.

lib/sidekiq/testing.rb
```diff
@@ -143,28 +202,36 @@ module Worker
     #
     module ClassMethods
 
+      # Queue for this worker
+      def queue
+        self.sidekiq_options["queue"]
```
mperham: Any worries about symbol vs string for this value?

lib/sidekiq/testing.rb
```diff
@@ -183,18 +250,22 @@ def execute_job(worker, args)
...
-        jobs.clear
+        Queues.clear_all
       end
 
       # Drain all queued jobs across all workers
       def drain_all
-        until jobs.values.all?(&:empty?) do
-          jobs.keys.each(&:drain)
+        while jobs.any?
+          worker_classes = jobs.map { |job| job["class"] }
```
mperham: Do you need to `uniq` the results?

lib/sidekiq/testing.rb
brandonhilkert: I think we're safe b/c of the underlying hash: https://github.com/mperham/sidekiq/blob/queue-testing-api/lib/sidekiq/worker.rb#L81

lib/sidekiq/testing.rb
brandonhilkert: Definitely. Updated.

lib/sidekiq/testing.rb
brandonhilkert: I updated the test to specify a symbol queue and all still passed, so that confirms and should ensure that condition is tested going forward.

4.0-Upgrade.md
```diff
@@ -25,6 +25,16 @@ gem 'redis-namespace'
   `concurrency + 2` connections in your pool or Sidekiq will exit.
   When in doubt, let Sidekiq size the connection pool for you.
 
+* There's a new testing API based off the `Sidekiq::Queues` namespace. All
+  assertions made against the Worker class still work as expected.
+```ruby
+assert_equal 0, Sidekiq::Queues["default"].size
+HardWorker.perform_async(:something)
```
mperham: :something is not a legal argument since symbols aren't native JSON.  Make sure the sample code is valid and supported.

brandonhilkert: Whoops! Updated.

mperham:
LGTM but build failure.

brandonhilkert:
Pushed a fix. Think we're good.

mperham:
Can you add a section on the Testing wiki page too (make sure to note it's 4.0+)?

brandonhilkert:
Will do.

mhenrixon:
This broke all my tests for `expect(UntilExecutedJob.jobs.size).to eq(2)` what do I need to change to make them pass against Sidekiq 4?

mhenrixon:
Might not be the reason after all nothing ends up in either array for queues or the worker class jobs. Not sure what changed but I supposed it could be something else. It is only the specs for sidekiq fake testing that broke. Will keep investigating. Quite unexpected considering nothing in the testing overrides changed.

mhenrixon:
```ruby 
# Test broken in SidekiqUniqueJobs
it 'does not push duplicate messages', :focus do
  param = 'work'
  expect(UntilExecutedJob.jobs.size).to eq(0)
  expect(UntilExecutedJob.perform_async(param)).to_not be_nil
  expect(UntilExecutedJob.jobs.size).to eq(1)
  expect(UntilExecutedJob.perform_async(param)).to be_nil
  expect(UntilExecutedJob.jobs.size).to eq(1)
end

# Test working in SidekiqUniqueJobs
it 'does not push duplicate messages', :focus do
  param = 'work'
  expect(UntilExecutedJob.jobs.size).to eq(0)
  expect(UntilExecutedJob.perform_async(param)).to_not be_nil
  expect(Sidekiq::Queues['working'].size).to eq(1)
  expect(UntilExecutedJob.perform_async(param)).to be_nil
  expect(Sidekiq::Queues['working'].size).to eq(1)
end
```

Was there an intent to break backwards compatibility with `WorkerClass.jobs.size` or am I just doing something wrong? I thought the jobs array should still contain the jobs pushed to it regardless of the Queue stuff or not?

brandonhilkert:
The intent was that it would work without change, but that doesn't seem to be the case. Sorry about that :). Local tests using the worker API pass. I'll try to dig in in the next day or so.

mhenrixon:
No worries @brandonhilkert I already fixed the problem for me in the related pull request but didn't spend any time investigating a failing example before I did. I saw the test using a symbol for name of the queue in the worker and then fetching by a string which seemed odd since I couldn't find any code that would change this. If you take a look at https://github.com/mperham/sidekiq/blob/master/test/test_testing_fake.rb#L289 it seems like it should be working I just couldn't find the implementation that actually changes the symbol to a string. The changes I made does just that but maybe I am overriding or missing something in `SidekiqUniqueJobs`?

PR #2661, opened 2015-11-14
Ensure comparison between strings

mhenrixon:
Fixes the issue with comparing symbol to strings caused by #2659

I have a lot of symbols in my code and so do others. This pull request ensure that testing queus and jobs (just like production) are only ever compared as strings.

### DIFF
--- a/lib/sidekiq/testing.rb
+++ b/lib/sidekiq/testing.rb
@@ -130,7 +130,7 @@ module Sidekiq
     #
     class << self
       def [](queue)
-        jobs[queue]
+        jobs[queue.to_s]
       end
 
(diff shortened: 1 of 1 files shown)


brandonhilkert:
Yup, just found this to be the case. Thanks for digging it out.

mhenrixon:
I agree on this being a valid approach. If you take a look at the production implementation it does `stringify_key` and dumps to JSON which when returned from redis is then a string key.

Would love a more generic way than `.to_s` but am afraid it will be hard to find one that doesn't involve `ActiveSupport::HashWithIndifferentAccess`.

Issue #2663, opened 2015-11-17
Testing mode no longer working


sethvargo:
After upgrading to 4.0, any job that uses the following format is failing:

```ruby
expect {
  MyWorker.perform_async("arg")
}.to change(MyWorker.jobs, :size).by(1)
```

Is this a regression?

I'm going to go code diving, but I suspect that the value of `job` is somehow being cached, because this works:

```ruby
expect {
  MyWorker.perform_async("arg")
}.to change { MyWorker.jobs }
```

petergoldstein:
@sethvargo @brandonhilkert This appears to be due to the queue-based jobs change that was recently introduced.

Specifically, the worker class' `jobs` method used to return a reference to the underlying array of jobs.  That array gets updated when new jobs are added.  Now the `jobs` method returns an array derived from the underlying array at the time the method is called, and this derived array is not updated as jobs are added to the queue.  See below:

           def jobs
    -        Worker.jobs[self]
    +        Queues.jobs[queue].select { |job| job["class"] == self.to_s }
           end

mperham:
@brandonhilkert I'm pretty busy at rubyconf.  Can you fix and/or revert, whatever you're comfortable with?  I'll get a 4.0.1 release out asap with whatever you decide.  I'm ok with killing the feature if it's half-baked.  Likely no one is using it yet.

sethvargo:
@mperham oh have fun at RubyConf!

In my testing and exploration, it looks like RSpec is quietly moving away from the `change(thing, :method)` toward `change { }`. None of the documentation recommends that pattern anymore. While it hasn't been deprecated, I wonder if it would just be worth calling out in the changelog as a breaking change (it was a major version bump after all), and updating the wiki?

mperham:
@sethvargo There's millions of lines of Ruby with that pattern.  I don't want to break everyone's code just for this feature that slipped in late.

brandonhilkert:
PR for the revert: https://github.com/mperham/sidekiq/pull/2664

mperham:
4.0.1 is out.

brandonhilkert:
@sethvargo Any chance you could run your tests against the updated `queue-api` branch?

sethvargo:
@brandonhilkert I refactored all our tests to use the other RSpec syntax when we were blocked by the original issue. So, yes, I can run them, but it won't be a good indicator if the problem is fixed :smile:. Sorry!

brandonhilkert:
[...]

PR #2676, opened 2015-11-25
Queue Testing API

brandonhilkert:
When using the Sidekiq::Client API to push jobs on to the queue, it's
not ideal to assert the size of the queue from the perspective of a
worker because the worker may not exist in the application.

This API implements a testing API from the perspective of a queue. The
existing Worker-based testing API remains unchanged, but leverages the
job hash implemented through the Sidekiq::Queues class.

Examples:

    assert_equal 1, Sidekiq::Queues["default"].size
    assert_equal "SpecialWorker", Sidekiq::Queues["default"].first["class"]
    Sidekiq::Queues["default"].clear
    Sidekiq::Queues.clear_all

ref: https://github.com/mperham/sidekiq/issues/2562

There was a [regression](https://github.com/mperham/sidekiq/issues/2663) involving the `Rspec` syntax `change(Queueworker.jobs, :size).by(1)` . This has been solved keeping a reference to a real array based on the original fake implementation. 

I created a test Rails app and [added a test to verify the issue is fixed](https://github.com/brandonhilkert/sidekiq-queue-rails/blob/master/spec/controllers/posts_controller_spec.rb#L21). This test fails with the original code and passes with the latest changes include here in this PR.

Additionally, a test was added to attempt to simulate what Rspec is doing under the hood: https://github.com/mperham/sidekiq/commit/c4330cb32628627045287b22442cf71600218108#diff-db88da5a19a4fee1b469d961e1d38a81R333

When I ran this test WITHOUT the fix, i get:

```
[10:14:46] bhilkert [~/Dropbox/code/sidekiq] (old-queue-api) $ be rake
Run options: --seed 26376

# Running:

................................................................................................................................................................................................................................F...................................

   1) respects underlying array changes
      Expected: 1
        Actual: 0
      # test/test_testing_fake.rb:343

Finished in 1.545262s, 168.2563 runs/s, 477.5889 assertions/s.
260 runs, 738 assertions, 1 failure, no errors, no skips
```

With the changes here, that test now passes. 

Thoughts?

### DIFF
--- a/4.0-Upgrade.md
+++ b/4.0-Upgrade.md
@@ -25,6 +25,16 @@ gem 'redis-namespace'
   `concurrency + 2` connections in your pool or Sidekiq will exit.
   When in doubt, let Sidekiq size the connection pool for you.
 
+* There's a new testing API based off the `Sidekiq::Queues` namespace. All
+  assertions made against the Worker class still work as expected.
+```ruby
+assert_equal 0, Sidekiq::Queues["default"].size
+HardWorker.perform_async("log")
+assert_equal 1, Sidekiq::Queues["default"].size
+assert_equal "log", Sidekiq::Queues["default"].first['args'][0]
+Sidekiq::Queues.clear_all
+```
+
 ## Upgrade
 
 First, make sure you are using Redis 2.8 or greater. Next:
--- a/Changes.md
+++ b/Changes.md
@@ -13,6 +13,16 @@
   and to remove dependencies.  This has resulted in major speedups, as
   [detailed on my blog](http://www.mikeperham.com/2015/10/14/optimizing-sidekiq/).
 - See the [4.0 upgrade notes](4.0-Upgrade.md) for more detail.
+- There's a new testing API based off the `Sidekiq::Queues` namespace. All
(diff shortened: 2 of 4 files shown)


mperham:
Looks pretty slick.  You'll need to move the Changes.md content to 4.0.2.

brandonhilkert:
Donzo

mperham:
Thanks, I'm going to wait a while for 4.0 to settle so this might not be released until Dec sometime.

brandonhilkert:
Sounds good!

On Wednesday, November 25, 2015, Mike Perham <notifications@github.com>
wrote:

> Thanks, I'm going to wait a while for 4.0 to settle so this might not be
> released until Dec sometime.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mperham/sidekiq/pull/2676#issuecomment-159679306>.
>


-- 

------------------------------------
*Build a Ruby Gem is available!
<http://brandonhilkert.com/books/build-a-ruby-gem/?utm_source=gmail-sig&utm_medium=email&utm_campaign=gmail>*

http://brandonhilkert.com

PR #2690, opened 2015-12-04
Refactor the daemonize method

matugm:
I refactored the `daemonize` method into smaller methods to make it easier to read. 

The test doesn't seem to cover this functionality at all, but I didn't want to add new test since this is a private method. I did some manual testing and everything seems to be ok.

### DIFF
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -165,22 +165,31 @@ module Sidekiq
     def daemonize
       return unless options[:daemon]
 
+      reopen_files_in_readwrite_mode
+      redirect_standard_output
+      initialize_logger
+    end
+
+    def reopen_files_in_readwrite_mode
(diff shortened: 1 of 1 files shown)


davydovanton:
LGTM :+1:

lib/sidekiq/cli.rb
```diff
@@ -165,22 +165,31 @@ def print_banner
...
-          file.reopen file.path, "a+"
-          file.sync = true
-        rescue ::Exception
-        end
-      end
+      files_to_reopen.each { |file| reopen_file(file) }
+    end
+
+    def files_to_reopen
+      ObjectSpace.each_object(File).reject { |file| file.closed? }
```
davydovanton: I'm prefer `to_proc` method here :blush:

mperham:
@matugm Thank you for your PR but I find the original code easier to read.  I find lots of little methods turn into a maze that is tougher to understand in some cases.

Issue #2691, opened 2015-12-04
perform_in fails when passed a string for the interval


selenetelenav:
```
 :001 > SomeJob.perform_in '60'
TypeError: can't convert String into an exact number
	from /Users/selenet/.rvm/gems/ruby-2.2.2@ventura/gems/activesupport-4.1.9/lib/active_support/core_ext/time/calculations.rb:208:in `+'
	from /Users/selenet/.rvm/gems/ruby-2.2.2@ventura/gems/activesupport-4.1.9/lib/active_support/core_ext/time/calculations.rb:208:in `plus_with_duration'
	from /Users/selenet/.rvm/gems/ruby-2.2.2@ventura/gems/sidekiq-4.0.1/lib/sidekiq/worker.rb:60:in `perform_in'
	from (irb):1
```

This was introduced in https://github.com/mperham/sidekiq/pull/2198 .

mperham:
What's the bug?  This seems like correct behavior.

selenetelenav:
Prior to #2198, passing a numeric string to .perform_in worked. e.g. `perform_in('60')` would schedule the job for 60 seconds in the future.

mperham:
The intent has always been that the first argument must be a timestamp, numeric or something that acts numeric (like an AS time interval).  Sorry if that intent was under-documented but no plans to support a String argument.

PR #2692, opened 2015-12-04
perform_in accepts strings for the interval

selenetelenav:
Fixes https://github.com/mperham/sidekiq/issues/2691

### DIFF
--- a/lib/sidekiq/worker.rb
+++ b/lib/sidekiq/worker.rb
@@ -57,7 +57,7 @@ module Sidekiq
       def perform_in(interval, *args)
         int = interval.to_f
         now = Time.now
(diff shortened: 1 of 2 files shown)


lib/sidekiq/worker.rb
```diff
@@ -57,7 +57,7 @@ def perform_async(*args)
       def perform_in(interval, *args)
         int = interval.to_f
         now = Time.now
-        ts = (int < 1_000_000_000 ? (now + interval).to_f : int)
+        ts = (int < 1_000_000_000 ? ((interval.is_a? String) ? now + int : now + interval).to_f : int)
```
davydovanton: Maybe replace first ternary operator to separate variable for improving code  readability, what do you think?

PR #2698, opened 2015-12-08
Prevent the footer from covering the delete and retry buttons.

Wildebeest:
If the Redis URL or namespace is too long, the footer will creep up and cover the delete and retry buttons. This change will hide the namespace in this case, rather than hide the buttons.

Before:
screen shot 2015-12-08 at 12 55 22 am

After:
screen shot 2015-12-08 at 12 58 52 am

Also delete some unused CSS.

### DIFF
--- a/web/assets/stylesheets/application.css
+++ b/web/assets/stylesheets/application.css
@@ -258,31 +258,6 @@ table .table-checkbox label {
   display: none;
 }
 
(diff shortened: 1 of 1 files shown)


mperham:
There's already a fix in 4.0.

Wildebeest:
@mperham Sweet! Thanks for taking a look at this. :) I'll close this one then.

PR #2701, opened 2015-12-08
fix typos

scalp42:


mperham:
Is that French change legit?

scalp42:
Nobody use "fil" to describe threads in French, it sounds like a translation straight from Google or something as in "sewing".

"Processus" sounds weird too but thats the actual translation. French is not Canadian French and has a lot of anglicism in it.

PR #2742, opened 2015-12-31
Freeze all string literals in source, fixes #2741

### LINKED ISSUE #2741: Support Ruby's frozen string mode
Ruby 2.3 allows a new frozen string magic comment.  Add the comment to the top of all Sidekiq source files, remove all `.freeze` calls, get the test suite running and smoke test the system.

mperham:
```
The snow glows white
On the mountain tonight
Not a footprint to be seen
A kingdom of isolation
And it looks like I’m the Queen

The wind is howling
Like this swirling storm inside
Couldn’t keep it in
Heaven knows I tried…

Don’t let them in
Don’t let them see
Be the good girl you always have to be
Conceal
Don’t feel
Don’t let them know…
Well, now they know!

Let it go, let it go
Can’t hold it back anymore
Let it go, let it go
Turn away and slam the door!
I don’t care what they’re going to say
Let the storm rage on
The cold never bothered me anyway

It’s funny how some distance
Makes everything seem small
And the fears that once controlled me
Can’t get to me at all!

It’s time to see
What I can do
To test the limits and break through
No right, no wrong
No rules for me
I’m free!

Let it go! Let it go!
I am one with the wind and sky!
Let it go! Let it go!
You’ll never see me cry!
Here I stand and here I’ll stay
Let the storm rage on…

My power flurries through the air into the ground
My soul is spiraling in frozen fractals all around
And one thought crystallizes like an icy blast
I’m never going back
The past is in the past!

Let it go! Let it go!
And I’ll rise like the break of dawn!
Let it go! Let it go!
That perfect girl is gone!

Here I stand in the light of day…
Let the storm rage on!!!
The cold never bothered me anyway
```


b264:
:+1:

PR #2744, opened 2016-01-01
[RFC] Reuse worker objects (drop object allocation count)

vassilevsky:
Hi 👻

It seems that something like this can help reduce object allocation count, GC pauses, and overall memory usage as well as give a small speed bump in cases where there are very many small jobs to process and the jobs do not become unusable-dirty after the first run.

What do you think?

### DIFF
--- a/lib/sidekiq/processor.rb
+++ b/lib/sidekiq/processor.rb
@@ -35,6 +35,7 @@ module Sidekiq
       @job = nil
       @thread = nil
       @strategy = (mgr.options[:fetch] || Sidekiq::BasicFetch).new(mgr.options)
+      @reusable_workers = {}
     end
 
     def terminate(wait=false)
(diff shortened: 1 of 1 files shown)


mperham:
This removes one allocation but introduces caching which can lead to subtle bugs.  My gut says it's not worth the work.

vassilevsky:
It also came to my mind that I can override a worker class' `new` to return the same object all the time. That would not require any changes in Sidekiq.

Thanks for taking the time for the reply! 🙏

mperham:
I'd suggest benchmarking before doing any performance tuning.  I suspect you'll find that optimizing object allocation by overriding `new` is brittle with little gain.

vassilevsky:
Yes, definitely, this requires measurements and all.

PR #2747, opened 2016-01-04
Fix Japanese translation for `NoRetriesFound`

mizoR:
"再試行できません" means "It can not be retried.", so it has negative (or dangerous) nuance.

"再試行するジョブはありません" is better, it means "There is no jobs to be retried."


davydovanton:
Thanks for contribution, @mizoR! :+1:

mizoR:
:smile:

PR #2753, opened 2016-01-06
Fix locale definition for testing locale

### LINKED ISSUE #2752: Non-deterministic build?
The build is passing and failing randomly.  Investigate.

davydovanton:
Fixed #2752
I found the bug. It was in locale defenition before running test. If you look to [`locale_files`](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/web_helpers.rb#L20-L24) method you can find that `@@locale_files` variable cache after first method call. And in test you call this method, and after you update `settings.locales` value.

### DIFF
--- a/test/test_web.rb
+++ b/test/test_web.rb
@@ -29,6 +29,8 @@ class TestWeb < Sidekiq::Test
       end
     end
 
+    Sidekiq::Web.settings.locales << File.join(File.dirname(__FILE__), "fixtures")
+
     it 'can show text with any locales' do
       rackenv = {'HTTP_ACCEPT_LANGUAGE' => 'ru,en'}
       get '/', {}, rackenv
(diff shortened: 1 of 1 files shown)


davydovanton:
Now I think all tests will be passed
/cc @mperham

mperham:
The tests are still order-dependent.  I'd prefer to see the tests explicitly add the locale and clear the cache inside the `it` block or in a `before` block.

davydovanton:
No, the problem is not in the order of the tests, the problem is in trying to update the locale files on the fly.  Therefore I consider that we need to replace all initial code to the top of file.  
But if you want to update locale files in the fly - I can offer other way. We can change `locale_files` value in before clock.
What do you think?

mperham:
I'd like any global state changes (like the list of locales) to be local to each test.  That means if a test adds a locale before the test, it should remove the locale after the test.  And any locale caching needs to be cleared too, like @@strings and @@locale_files.

mperham:
BTW I'm happy to do this work if my requirements aren't clear to you.

davydovanton:
no, I understood you. I'm working on this now.

davydovanton:
@mperham I added `before` and `after` blocks and if it solution is good for you - I can squash commits before a merge.

mperham:
I developed my own solution which isn't quite as low-level as yours.  Hope you don't mind. https://github.com/mperham/sidekiq/commit/f8baa8af37274cacd3ef0437e9cede5ad5ecbe0d

PR #2757, opened 2016-01-07
Add "quiet" label to UI when process is quieted"

jcarlson:
I'd like to see a visual indicator in the web UI when a process has been quieted (or is stopping).

Our shop uses a blue/green deployment strategy, so the inactive processes are "quieted" after a deployment, but they continue to heartbeat and appear in the web UI. We'd like to be able to easily identify and confirm that they are "quiet".

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -738,6 +738,10 @@ module Sidekiq
     def dump_threads
       signal('TTIN')
     end
+    
+    def stopping?
+      self['stopping']
+    end
 
     private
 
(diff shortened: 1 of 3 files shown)


jcarlson:
@mperham I guess I answered my own question. It doesn't look like it was that hard.

jcarlson:
Well, I botched the implementation at first, but this looks workable, now.

mperham:
:+1: Really nice work.  Sometimes it takes a while to get there but you got it where it needed to go.

Issue #2755, opened 2016-01-07
Sidekiq not properly working when rails is already loaded


TiagoCardoso1983:
With Sidekiq 3.5.3:

```ruby
[1] pry(main)> require './config/environment'                                                                         
Sidekiq::VERSION                                                                                                      
=> true                                                                                                               
[2] pry(main)> Sidekiq::VERSION                                                                                       
=> "3.5.3"                                                                                                            
[3] pry(main)> load Bundler.which('sidekiq')                                                                          
         m,                                                                                                           
         `$b                                                                                                          
    .ss,  $$:         .,d$                                                                                            
    `$$P,d$P'    .,md$P"'                                                                                             
     ,$$$$$bmmd$$$P^'                                                                                                 
   .d$$$$$$$$$$P'                                                                                                     
   $$^' `"^$$$'       ____  _     _      _    _                                                                       
   $:     ,$$:       / ___|(_) __| | ___| | _(_) __ _                                                                 
   `b     :$$        \___ \| |/ _` |/ _ \ |/ / |/ _` |                                                                
          $$:         ___) | | (_| |  __/   <| | (_| |                                                                
          $$         |____/|_|\__,_|\___|_|\_\_|\__, |                                                                
        .d$$                                       |_|                                                                
                                                                                                                      
2016-01-07T10:47:10.354Z 728863 TID-oswu3a6e8 INFO: Running in ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-linux]
2016-01-07T10:47:10.354Z 728863 TID-oswu3a6e8 INFO: See LICENSE and the LGPL-3.0 for licensing details.
2016-01-07T10:47:10.354Z 728863 TID-oswu3a6e8 INFO: Upgrade to Sidekiq Pro for more features and support: http://sidekiq.org                                                                                                                
2016-01-07T10:47:10.355Z 728863 TID-oswu3a6e8 INFO: Starting processing, hit Ctrl-C to stop
```

With Sidekiq 4.0.1:

```ruby
[1] pry(main)> require './config/environment'                                                                         
Sidekiq::VERSION                                                                                                      
=> true                                                                                                               
[2] pry(main)> Sidekiq::VERSION                                                                                       
=> "4.0.1"                                                                                                            
[3] pry(main)> load Bundler.which('sidekiq')                                                                          
         m,                                                                                                           
         `$b                                                                                                          
    .ss,  $$:         .,d$                                                                                            
[...]

PR #2759, opened 2016-01-08
lazy loading the redis pool 

TiagoCardoso1983:
Related with https://github.com/mperham/sidekiq/issues/2755

This bypasses redis pool creation until it is needed. It's a minor improvement for the normal usage of sidekiq, but improves CoW support when switching to server mode after fork. 

This is however not enough in the standard case due to the configure_client and configure_server blocks which will usually be called before fork (if one wants to eager load a rails app before forking, for example), if one wants to define different error handlers and other stuff you may do in your configure blocks (although one could take the same approach as in middlewares where one defines them separately).  

This is not related to the swarm feature from enterprise mode, instead it's for the use case of preloading a rails app in a master process, fork it and initialize sidekiq server.

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -91,15 +91,19 @@ module Sidekiq
     end
   end
 
+  REDIS_MUTEX = Mutex.new
+
   def self.redis_pool
-    @redis ||= Sidekiq::RedisConnection.create
+    @redis || REDIS_MUTEX.synchronize do
+      @redis ||= Sidekiq::RedisConnection.create(@redis_params || {})
+    end
   end
 
   def self.redis=(hash)
-    @redis = if hash.is_a?(ConnectionPool)
-      hash
+    if hash.is_a?(ConnectionPool)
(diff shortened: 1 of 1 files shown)


mperham:
Sorry, not a chance I'm putting a mutex there. You'll need to monkey patch.

TiagoCardoso1983:
Understand your concern. Just gonna say that the mutex block part will be run at-most-once, after which you'll no longer need to go through it to access the redis variable. I'm already monkey-patching.

mikegee:
@mperham Could you add a word about why a mutex is bad there for causal issue readers like me? Is it because that is a frequently executed method that you don't want to add overhead to?

mperham:
@mikegee In this case I was mostly concerned about readability.  I already have a plan for building similar functionality which doesn't require hacking those core methods.

I was concerned about a Mutex in a performance critical method but he's right that his double checked locking mostly fixes that issue.

mikegee:
Thank you!

PR #2770, opened 2016-01-14
Sidekiq::Worker.drain_all loops endlessly when queue name is symbolized

srecnig:
The queue-based testing hash (@jobs_by_queue) stores the queue name as a string. Therefore, when a worker is given a symbol as queue name, deleting the job from the @jobs_by_queue silently fails, leaving the job in @jobs_by_queue (even though deleting succeded from the @jobs_by_worker). 

Sidekiq::Worker.drain_all gets the jobs through @jobs_by_queue, but since the jobs are never removed from the queue-based hash, calling drain_all will result in an endless loop.

Without my patch in Queues.delete_for, my added test will result in the described endless loop. If my suggested patch/test needs more work, please let me know, i'm more than happy to do that.

thanks!

### DIFF
--- a/lib/sidekiq/testing.rb
+++ b/lib/sidekiq/testing.rb
@@ -170,7 +170,7 @@ module Sidekiq
       end
 
       def delete_for(jid, queue, klass)
-        jobs_by_queue[queue].delete_if { |job| job["jid"] == jid }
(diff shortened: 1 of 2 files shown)


davydovanton:
:+1:

PR #2771, opened 2016-01-14
Avoid recomputing range_size everytime in API

ddrmanxbxfr:
This PR is a minor optimization in the API.

We should avoid recomputing the range size 2 times as we already computed it for range_start.

Thanks!


davydovanton:
thanks :+1:

PR #2772, opened 2016-01-14
Apply range_start optimization in SortedSet class

ddrmanxbxfr:
Same optimization as the queue class.

Compute the range size once.

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -227,7 +227,7 @@ module Sidekiq
 
       while true do
         range_start = page * page_size - deleted_size
-        range_end   = range_start + (page_size - 1)
+        range_end   = range_start + page_size - 1
         entries = Sidekiq.redis do |conn|
           conn.lrange @rname, range_start, range_end
         end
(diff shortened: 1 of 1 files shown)


lib/sidekiq/api.rb
```diff
@@ -502,7 +502,7 @@ def each
 
       while true do
         range_start = page * page_size + offset_size
-        range_end   = page * page_size + offset_size + (page_size - 1)
+        range_end   = range_start + (page_size - 1)
```
krzysiek1507: @ddrmanxbxfr `( )` unnecessary ;)

ddrmanxbxfr: Uh uh, indeed. Fixing it in both classes

krzysiek1507: :+1:

davydovanton:
@krzysiek1507 could you fix `( )` in #2771 too?

ddrmanxbxfr:
I just did in this PR if you look at the changes.

Thanks!

krzysiek1507:
@davydovanton you mean @ddrmanxbxfr ;)

ddrmanxbxfr:
Both method we're patched in this PR. The one in #2771 and a new one in
sorted set class.

davydovanton:
thanks!

PR #2777, opened 2016-01-17
Speed up page rendering on sidekiq.org

nateberkopec:
+ Fix canonical links on some pages
+ Defer script loading to DOMContentLoaded with defer attribute.
+ Use 1 CDN (cloudflare) instead of 2
+ Glyphicons -> woff2 on cloudflare
+ Use https on Google Fonts, allowing HTTP/2 connection

@mperham, you've got a lot of network blocking happening because of all of the PNG files on the homepage. You can only download the browser's maximum connection limit at a time, which I think is usually like 6. The image downloads sometimes also crowd out more important JS requests served from your domain. I suggest you sign up for Cloudflare free and pipe everything through them - that'll serve all of sidekiq.org over HTTP/2, which will allow those images to download in parallel more efficiently.


nateberkopec:
I should note the defer attribute is buggy on IE8 and below, LMK if you care and I can remove it. 

Also, should mention that my local testing shows this speeds up initial render by ~30-40%. The page is really fast already so that's only like a couple hundred milliseconds.

PR #2783, opened 2016-01-20
gemspec celluloid ~> 0.17 to match rails5 dependency

twairball:
updated celluloid gem dependency to match rails 5.0.0.beta1 dependency (from actioncable which depends on celluloid ~> 0.17.2)


mperham:
This branch is ancient and no longer relevant.  You don't want to use it.

twairball:
eek! sorry about that. should I be testing master branch, for rails5 upgrades ?

mperham:
Yes, master is the latest stuff.

PR #2784, opened 2016-01-20
Change relative_time method behaviour

maxehmookau:
This commit updates the `relative_method` on `Sidekiq::WebHelpers` so that it responds with something more specific rather than a Rails-y style *"3 years from now"* as it seems to do now (although, only sometimes?) I'm guessing there's some magic in the Rails Time class which overrides the `to_s` method in certain circumstances?
screen shot 2016-01-20 at 14 29 51

It also adds a test for this method as there weren't any before.

### DIFF
--- a/lib/sidekiq/web_helpers.rb
+++ b/lib/sidekiq/web_helpers.rb
@@ -146,7 +146,7 @@ module Sidekiq
     end
 
     def relative_time(time)
-      %{<time datetime="#{time.getutc.iso8601}">#{time}</time>}
+      %{<time datetime="#{time.getutc.iso8601}">#{time.to_formatted_s(:long)}</time>}
     end
 
     def job_params(job, score)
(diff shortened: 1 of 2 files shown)


mperham:
Yes, I can see **what** you did.  I can't see **why** you did it.  Why do you think Sidekiq's 1000s of users would prefer your behavior?

maxehmookau:
Ah sorry, that was quite an oversight.

Let's say you've got 100 jobs on that page, all in the very distant future but all very close (within a few minutes). The UI would report that all of these events are essentially happening at the same time "3 years from now" which IMO could be confusing. 
While admittedly "3 years from now" is clearer to read, during development it's quite helpful to know exactly when a job is scheduled for or at the very least, the day it's scheduled for.

mperham:
If you mouse over the link, it shows the precise time and the entries are sorted by time.

mperham:
Relative times are useful because they don't require any timezone handling.  If we showed precise times, people would want to configure their display timezone, complicating the Web UI.

maxehmookau:
Ah fair enough then, thanks for reviewing anyway.

PR #2787, opened 2016-01-23
Pass exception object to sidekiq_retries_exhausted block

Nowaker:
Behold! I'm taking a challenge of making an **extremely** compelling case for why Sidekiq's thousands of users would want that change. :-)

1. Finally have a dead easy way to handle an exception when retries are exhausted. Currently it's doable only through global error handlers (`config.error_handlers`). A lot of custom code is needed to implement a per-job handler. Example use cases:
  - Suspend account when no money could be collected after retries.
  - Report to Bugsnag after the job really fails. (Instead of the default way of doing that, every job retry)
2. Backwards compatible. Block parameters are optional in Ruby. Thanks to that `sidekiq_retries_exhausted do |msg|` and `sidekiq_retries_exhausted do |msg, e|` just work.
3. No performance penalty. No extra computing is needed to create an object. No extra garbage created. Exception object is there, just pass it.

### DIFF
--- a/lib/sidekiq/middleware/server/retry_jobs.rb
+++ b/lib/sidekiq/middleware/server/retry_jobs.rb
@@ -130,17 +130,17 @@ module Sidekiq
             end
           else
             # Goodbye dear message, you (re)tried your best I'm sure.
-            retries_exhausted(worker, msg)
+            retries_exhausted(worker, msg, exception)
           end
 
           raise exception
         end
 
-        def retries_exhausted(worker, msg)
+        def retries_exhausted(worker, msg, exception)
           logger.debug { "Dropping message after hitting the retry maximum: #{msg}" }
           begin
             if worker.sidekiq_retries_exhausted_block?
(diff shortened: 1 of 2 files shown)


mperham:
Compelling enough.  Please add test assertion(s) to verify.

Nowaker:
Thank you, will do.

Nowaker:
Here you go. One test checks if `sidekiq_retries_exhausted do |msg|` still works. The rest is actual testing of `sidekiq_retries_exhausted` block which wasn't tested before. Code coverage metrics will grow with this commit. ;)

mperham:
Would you update the changelog with a summary and give yourself credit?

Nowaker:
Looks like you already did it: https://github.com/mperham/sidekiq/blob/master/Changes.md. Thanks!

Would you like to have this feature backported to 3.x too? (If we can even say "backport", the patch applies perfectly on the previous codebase)

mperham:
My policy is no new features or API changes in 3.x.  Bug fixes only.

PR #2802, opened 2016-01-29
Update redis_connection.rb

zedtux:
Improve error message when the `redis-namespace` gem is missing with a namespace parameter.


zedtux:
My English is not that good, so if someone can review it, it would be appreciated :)

mperham:
You English seems fine to me, thanks!

zedtux:
Awesome, thank you @mperham

PR #2809, opened 2016-02-02
Refactoring all pages

seuros:
This was in my todo list for long time.  

ping @davydovanton


mperham:
Whoa, this is awesome!  Thank you, I will test shortly. :heart:

davydovanton:
Nice work! :star2: 
I don't prefer 4 spaces instead 2 but everything else looks fine :+1:

js/skq-global.js
```diff
@@ -1,13 +1,15 @@
...
     var body = $('html, body');
-    var windowHeight = $(window).height()
-    var scrollPos = document.body.scrollTop
-    var mobileBreak = 992
-
+    var windowHeight = $(window).height();
+    var scrollPos = document.body.scrollTop;
+    var mobileBreak = 992;
+    var logoCarousel = $(".logo-carousel");
+    var navBar = navbar;
```
mperham: This breaks for me.

Run `bundle exec jekyll s`, open `http://localhost:4000` and look at the JS console.

mperham:
@seuros Any chance you can look into that broken JS, see comment above?

js/skq-global.js
seuros: Oh, it should be navBar = $(".navbar"). I will fix it now.

mperham:
Better.  I noticed that the corp logos are cut off.

screen shot 2016-02-06 at 10 55 53

mperham:
The support page is linked incorrectly when on the Pro and Enterprise pages.

mperham:
Images don't work on the footer.  This is pretty buggy still, please do a few minutes of QA and let me know when it's better.

mperham:
You can test the site with `bundle exec jekyll s` and `open http://localhost:4000`

seuros:
Which browser did you use to have the images cut ? It work fine in chrome and firefox.

mperham:
I was using Chrome/OSX.  Really nice work, everything looks good now.

PR #2842, opened 2016-02-20
[ci-skip] Add ruby 2.3 to the README

tejasbubane:
Since ruby 2.3 is supported and present in [travis config](https://github.com/mperham/sidekiq/blob/master/.travis.yml), I think it should be mentioned in the README.


tejasbubane:
Looks like `[ci-skip]` does not work, should be `[ci skip]` :disappointed:

mperham:
readme has been updated

Issue #2850, opened 2016-02-25
Incompatibility with servers that rename-command INFO (since 4.1.0)


lwe:
First of all, thanks a lot for giving us sidekiq and contributing so much to the Ruby ecosystem.

Our workers are running on a CloudFoundry instance and a Redis service also running on CloudFoundry. The operator decided to disable `INFO` command due to security reasons, i.e. they basically added the following to their configuration file:

```ini
rename-command INFO ""
```

This week we switched from Sidekiq 3.x to Sidekiq 4.1.0 and our workers stopped booting with the following error `Redis::CommandError: ERR unknown command 'info'`. Checking to sidekiq source revealed that commit d285f194803780b639dd6095a785f638b627720f added an additional check:

```ruby
  ver = conn.info['redis_version']
  raise "Oops, you are using Redis v#{ver}, Sidekiq requires Redis v2.8.0 or greater" if ver < '2.8'
```

to ensure that it's really running against Sidekiq 2.8 or greater. We contacted our operator and they don't seem too inclined to make security related changes. Currently there are two remaining options for us

1. **Monekypatching**, we already did that as a quick-fix, but not really sure if this might break something else down the line: https://gist.github.com/lwe/590fbf613284d85cf1a3
2. **Creating a PR** on sidekiq to skip it or find a decent alternative (e.g. checking for a specific command added in 2.8.0?)

However before investing our time in a PR, do you think there is any possibility that this skip / check would make sense, if so any proposals or favorites to pursue?

### Environment

```
Redis server v=3.0.1 sha=00000000:0 malloc=libc bits=64 build=bf58331b4c8133f5
Sidekiq version (v4.1.0): ac6c23521e4f64fb47de77d2adbb96ea0536059f
ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin15]
Rails 4.2.5.1
```

### Stack trace

```
Redis::CommandError: ERR unknown command 'info'
    /_path_/gems/redis-3.2.2/lib/redis/client.rb:114:in `call'
    /_path_/gems/redis-3.2.2/lib/redis.rb:211:in `block in info'
    /_path_/gems/redis-3.2.2/lib/redis.rb:57:in `block in synchronize'
    /_path_/ruby/2.3.0/monitor.rb:214:in `mon_synchronize'
    /_path_/gems/redis-3.2.2/lib/redis.rb:57:in `synchronize'
[...]

PR #2851, opened 2016-02-25
Added Sidekiq#redis_info to encapsulate info calls

lwe:
As discussed in #2850 this adds the method `redis_info` to `Sidekiq`, to allow easier monkey patching for servers that disable `INFO`.

### Open items

-  [x] Had to use `Sidekiq.redis_info` and was not able to write it in `Utils.redis_info`, as the `Utils` are not included in the WebHelpers, okay?
- [x] Don't know sidekiq / redis enough, but is the call to `redis.info` for `:namespace` actually required, I've only seen `web_helpers.rb` actually using it like that
- [x] Added a test to check that the call to `redis_info` returns a few keys, maybe we should test more keys or specific values - or are there any keys that return a fixed value in tests?

### Proposal for a monkey-patch-free-version...

One more thing, if possible I'd really like to change `self.redis_info` to the following, but I agree that having to support all crazy redis configuration variants might not really be viable.

```ruby
  def self.redis_info
    redis do |conn|
      begin
        # admin commands can't go through redis-namespace starting
        # in redis-namespace 2.0
        if conn.respond_to?(:namespace)
          conn.redis.info
        else
          conn.info
        end
      rescue Redis::CommandError => ex
        # 2850 assume redis 2.8 when INFO has been renamed
        raise unless ex.message =~ /Unknown command/
        { 'redis_version' => '2.8' }
      end
    end
  end
```

Otherwise is it possible to add this to the wiki, or where would be the place to add a "supported" monkeypatch? Maybe on https://github.com/mperham/sidekiq/wiki/Problems-and-Troubleshooting, can I freely edit this page?

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -90,6 +90,18 @@ module Sidekiq
     end
   end
 
+  def self.redis_info
+    redis do |conn|
+      # admin commands can't go through redis-namespace starting
+      # in redis-namespace 2.0
+      if conn.respond_to?(:namespace)
+        conn.redis.info
+      else
+        conn.info
+      end
+    end
+  end
+
   def self.redis_pool
     @redis ||= Sidekiq::RedisConnection.create
   end
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -37,6 +37,14 @@ module Sidekiq
     'queue' => 'default'
   }
 
+  FAKE_INFO = {
+    "redis_version" => "9.9.9",
+    "uptime_in_days" => "9999",
+    "connected_clients" => "9999",
+    "used_memory_human" => "9P",
+    "used_memory_peak_human" => "9P"
+  }.freeze
+
   def self.❨╯°□°❩╯︵┻━┻
     puts "Calm down, yo."
   end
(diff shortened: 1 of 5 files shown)


mperham:
1. I think it's fine to put it in `Sidekiq`.  It's pretty core.
2. I don't follow.
3. Anything you return in lieu of real data should be obviously fake.  Don't return version "2.8", return version "9.9".

I'm fine with the monkeypatch-free version as long as the hash data is obviously fake.  Maybe make all of the numeric values 9s?  version => 9.9, connected_clients => 99, etc.

lwe:
1. :+1: 
2. Nevermind, `web_helper.rb` was the only place where it did check to make a distinction between `conn.redis.info` and `conn.redis`. I did check the source of redis-namespace and the answer is _yes, we should make this distinction_.
3. Makes sense. Would it be fine to only return `redis_version: 9.9` and nothing else?

mperham:
RE: 3. return all elements required by the Web UI.  I want it to show the nonsense rather than nothing at all so the user is made aware that something odd is happening.

You can automate the happy path test and test the edge case manually.

lwe:
Added a second commit with obvious fake info. When running the tests manually, there is one test case that fails:

```
   1) reconnects if connection is flagged as readonly
      Redis::CommandError: ERR unknown command 'info'
          /_path_/sidekiq/test/test_sidekiq.rb:92:in `block (3 levels) in <class:TestSidekiq>'
```

should code be added to skip this spec in that case or simply assume as the edge case is tested manually that it's okay that this fails?

PS: Will check any feedback tomorrow, I'm on CET, so thanks for all the input and good night.

mperham:
Thank you!

lwe:
:ok_hand: you are welcome.

PR #2891, opened 2016-03-19
Fixed SystemCallError for Windows path

nrakochy:
Updated logging path to fix SystemCallError on Windows environment, as '/dev/null' will break gem loading

### DIFF
--- a/lib/sidekiq/logging.rb
+++ b/lib/sidekiq/logging.rb
@@ -47,8 +47,7 @@ module Sidekiq
     end
 
     def self.logger=(log)
-      null_path = (File.exists?('/dev/null') ? '/dev/null' : 'nul')
-      @logger = (log ? log : Logger.new(null_path))
+      @logger = (log ? log : Logger.new(File::NULL))
     end
 
     # This reopens ALL logfiles in the process that have been rotated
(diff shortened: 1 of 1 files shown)


davydovanton:
I restarted jruby build

lib/sidekiq/logging.rb
```diff
@@ -47,7 +47,8 @@ def self.logger
     end
 
     def self.logger=(log)
-      @logger = (log ? log : Logger.new('/dev/null'))
+      null_path = (File.exists?('/dev/null') ? '/dev/null' : 'nul')
```
mperham: Could you make a platform check rather than a filesystem check?

nrakochy: The platform check gets a little verbose with the different types of platforms. New update removes boolean and simply calls File::NULL.

nrakochy:
Hmm... Not sure why the JRuby build keeps failing on Travis. Having a hard time reproducing on Windows/Mac/JRuby locally and I can't see the error log, so sorry to spam with non-working patches.

Would this work?

    Logger.new(File.open(File::NULL, "w"))

mperham:
Great idea, didn't know about File::NULL and doesn't appear to be documented.

PR #2906, opened 2016-03-29
allow sidekiq::client.via to point to same redis

junghlee90:
this fixes the incorrect conditional statement 

`if x = Thread.current[:sidekiq_via_pool] && x != pool`

which evaluates as if it was written as 

`if x = (Thread.current[:sidekiq_via_pool] && x != pool)`


mperham:
thank you!

PR #2930, opened 2016-04-15
Dump the context as JSON for proper logging

aiionx:
Dump the context as JSON for proper logging

### DIFF
--- a/lib/sidekiq/exception_handler.rb
+++ b/lib/sidekiq/exception_handler.rb
@@ -5,7 +5,7 @@ module Sidekiq
 
     class Logger
       def call(ex, ctxHash)
-        Sidekiq.logger.warn(ctxHash) if !ctxHash.empty?
+        Sidekiq.logger.warn(Sidekiq.dump_json(ctxHash)) if !ctxHash.empty?
         Sidekiq.logger.warn "#{ex.class.name}: #{ex.message}"
         Sidekiq.logger.warn ex.backtrace.join("\n") unless ex.backtrace.nil?
       end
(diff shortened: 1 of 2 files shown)


mperham:
What problem are you trying to solve?

aiionx:
Hi.

The Log prints the context as a ruby structure:
```
2016-04-15T02:45:00.240Z 9316 TID-osw3ap2is WARN: {"class"=>"ProjectWebHookWorker", "args"=>[...], "retry"=>true, "queue"=>"project_web_hook", "jid"=>"64f5e0416015381122462373", "created_at"=>1460671848.3572097, "enqueued_at"=>1460688290.1475492, "error_message"=>"Net::ReadTimeout", "error_class"=>"Net::ReadTimeout", "failed_at"=>1460671867.5247047, "retry_count"=>10, "retried_at"=>1460688300.2396612}
```

Its better if is printed as JSON

```
2016-04-15T02:45:00.240Z 9316 TID-osw3ap2is WARN: {"class": "ProjectWebHookWorker", ...}
```

That way it can be easily parsed.

Sorry I pushed without testing in local env.
If this is approved I can fix, otherwise it can be dropped

mperham:
It's not really designed for parsing, just reading by a developer.  Ideally you have another error handler which is pushing errors to an error service, like Honeybadger, Bugsnag, etc.  That said, I don't have a problem with your change, presuming you fix it.

aiionx:
Thanks,

I'm parsing your logs with logstash.
It makes it easier if the output is JSON and not a raw ruby data structure.

Test passes.

PR #2969, opened 2016-05-11
Edge case where 'args' is a string passed to display_args

suddenrushofsushi:
Why:
In rare cases the UI at /sidekiq/morgue will throw a 500 error with: undefined method `map' for "value":String; especially when third party tools can submit to a queue.

What:
This PR handles the edge case where a job's arguments are passed through to the display helper as a string, instead of an array. It also adds some minimal tests for the display_args helper.

### DIFF
--- a/lib/sidekiq/web_helpers.rb
+++ b/lib/sidekiq/web_helpers.rb
@@ -166,6 +166,7 @@ module Sidekiq
     end
 
     def display_args(args, truncate_after_chars = 2000)
+      args = [args] unless args.respond_to?(:map)
       args.map do |arg|
         h(truncate(to_display(arg), truncate_after_chars))
       end.join(", ")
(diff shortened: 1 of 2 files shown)


lib/sidekiq/web_helpers.rb
```diff
@@ -166,6 +166,7 @@ def truncate(text, truncate_after_chars = 2000)
     end
 
     def display_args(args, truncate_after_chars = 2000)
+      args = [args] unless args.respond_to?(:map)
```
davydovanton: may be is better to use `Array()` instead conditional? I.e. `Array(args).map {}`

suddenrushofsushi: Yes, I agree. I'll update that, thank you.

mperham:
Args must be an array. That's the job payload spec. Garbage in, garbage out.

suddenrushofsushi:
Alright, thanks. Closing.

PR #2978, opened 2016-05-18
Set kill timeout to 15 sec. in upstart config

vicentllongo:
Upstart "kill timeout" is 5 seconds by default (http://upstart.ubuntu.com/cookbook/#kill-timeout). 

This will cause that when using the provided upstart/sidekiq.conf, upstart will kill the sidekiq process while is waiting for any running workers to end, and even before the process enqueue them into the retries queue, losing any running jobs.

This PR sets a higher timeout that should be safe to use with the default sidekiq "kill_timeout" which is 10 seconds.

### DIFF
--- a/examples/upstart/sidekiq.conf
+++ b/examples/upstart/sidekiq.conf
@@ -42,6 +42,10 @@ normal exit 0 TERM
 # this commented out.
 reload signal USR1
 
+# Upstart waits 5 seconds by default to kill the a process. Increase timeout to
+# give sidekiq process enough time to exit.
+kill timeout 15
+
 instance $index
 
 script


mperham:
Does this work on earlier versions of Upstart?  Will it work on CentOS 6, for example?

vicentllongo:
It seems the setting has been around for a while (there's no mention it has been added in a recent version): http://upstart.ubuntu.com/cookbook/#stanzas-by-category

I have tested this with Ubuntu 14.04 which uses upstart 1.12.1. CentOS 6 seems to use v0.6.5, and by the changelog it seems is supported for that version, but maybe someone else using CentOS can confirm it.

mperham:
:metal:

Issue #2988, opened 2016-05-26
Namespace key must be a symbol when configuring Redis


clouvet:
We ran into a problem where Sidekiq was not using the namespace we'd specified in a yml config file: 
```yaml
development:
  sidekiq_redis:
    url: 'redis://127.0.0.1:6379/1'
    namespace: sidekiq
```
We configure Sidekiq by loading the hash from that yml file:
```ruby
# sidekiq.rb initializer
require 'sidekiq'

Sidekiq.configure_client do |config|
  config.redis = YAML.load_file(Rails.root.join('config', 'redis.yml'))[Rails.env.to_s]['sidekiq_redis']
end

Sidekiq.configure_server do |config|
  config.redis = YAML.load_file(Rails.root.join('config', 'redis.yml'))[Rails.env.to_s]['sidekiq_redis']
end
```
After looking through the source, it became clear from the `RedisConnection` class that I need to set the `namespace` key explicitly to be a symbol:
```yaml
development:
  sidekiq_redis:
    url: 'redis://127.0.0.1:6379/1'
    :namespace: sidekiq
```
That change solved our problem, but since that's not exactly intuitive and since other keys can be strings, I've created this issue. I'm happy to make a pull request which adds `options = options.symbolize_keys` to the `RedisConnection.create` method and then tests for `namespace` as a string too. It works on both Sidekiq 3.4.1 and 4.1.2 with all tests passing. 

Incidentally, this issue led me to [the blog post about not using namespaces](http://www.mikeperham.com/2015/09/24/storing-data-with-redis/). It seems a refactor is in our near future. 

----
Versions:
Ruby: 2.3.0
Rails: 3.2.22
Sidekiq: 3.4.1, 4.1.2

PR #2989, opened 2016-05-26
Symbolize keys in options hash for RedisConnection.create

clouvet:
Fixes an edge case where a stringified `namespace` key is ignored. A `namespace` key may be a string if the Sidekiq Redis config is loaded from a yml file.

This pull request also includes minor, syntactical improvements to error messaging related to redis-namespace. 

All tests pass.

Resolves https://github.com/mperham/sidekiq/issues/2988.


mperham:
Perfect, thank you!

PR #3001, opened 2016-06-04
Make process string more informative

vassilevsky:
When there are many Sidekiq processes running on a server each processing different queues, it is hard to say which is which 😰


mperham:
Use -g

vassilevsky:
Manual work is the worst 😭

PR #3004, opened 2016-06-07
Update Pro-Changes.md

allcentury:
Fix semantic version


mperham:
Nope, it is correct as-is.

Issue #2245, opened 2015-03-19
Job Transaction API


mperham:
Today Sidekiq performs a number of operations after a job is complete.  Each one of those operations is a round trip and independent from each other, meaning one could fail and leave data in a "half" state if the network happens to die at that moment.

I'm thinking about a possible API for middleware to register pipeline and multi blocks to be run as part of one "transaction" at the end of each job, so all operations can be performed with a max of two round trips and the MULTI block assures all or nothing semantics:

```ruby
def call(worker, message, queue)
  yield
  worker.on_finish_multi do |conn|
    conn.incr("somekeya", 1)
    conn.decr("somekeyb", 1)
  end
  worker.on_finish_pipeline do |conn|
    conn.incr("somekey", 1)
  end
end
```

Notably this would ensure Sidekiq can update statistics, reliably acknowledge jobs and perform batch logic all together.

PR #3022, opened 2016-06-17
Prototype of transactional API, #2245

mperham:
Note breaking changes in BasicFetch API.  This will require Sidekiq 5.0.


mperham:
Prototyping this has shown me that it can't be used to implement Pro's batch feature so it's a lot of change with some serious limitations.  Punting.

PR #3036, opened 2016-06-28
Proper jquery.timeago Norwegian locale shortname

thomey:
Fixes (No route matches [GET] "/sidekiq/javascripts/locales/jquery.timeago.nb.js").


mperham:
I think the proper fix is to create a copy for no and nb.  Isn't no a proper Norwegian locale code?

thomey:
It depends on the standard. We have to official languages; bokmål and nynorsk. According to ISO 639-1 the inclusive code is no and the individual codes are nb and nn. It's common to use nb-no and nn-no, but the locale reported by I18n (Sinatra, Rails etc.) is either nb or nn (which is why you have nb.yml). You could keep a copy for no but as long as you're using I18n it will probably never be used (and you would have to maintain two indentical files).

Side note: You should probably catch missing jquery.timeago localizations on line 10 in layout.erb as well ;)

mperham:
Thanks for the info.  If `nb.yml` works today, then nb is ok with me.

PR #3052, opened 2016-07-07
Introspect what sidekiq is working on at any time

acoravos:
We added an additional event for the TTIN signal, which we use to track which jobs are running. It lets us introspect what sidekiq is working on at any time. For example:

```
  # ...
  config.on(:status) do
    Thread.list.each do |thread|
      next unless thread[:job].present?
​
      Sidekiq.logger.warn "Interrupted Thread TID-#{thread.object_id.to_s(36)} #{thread[:job]}"
    end
  end
  
  # ...
end
```

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -26,6 +26,7 @@ module Sidekiq
     lifecycle_events: {
       startup: [],
       quiet: [],
+      status: [],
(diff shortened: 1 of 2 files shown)


mperham:
Where is `thread[:job]` coming from?  I don't think that's part of base Sidekiq.

Issue #2910, opened 2016-04-04
Getting the dreaded `Forbidden` message


hobakill:
`ruby 2.1.8p440 (2015-12-16 revision 53160) [x86_64-linux]`
`sidekiq 4.1.1`
`Phusion Passenger 5.0.26` (Standalone)
`Rails 4.1.1`

```
require 'sidekiq/web'
Sidekiq::Web.use Rack::Session::Cookie, secret: ENV['RACK_SESSION_COOKIE']
map '/sidekiq' do
  use Rack::Auth::Basic, 'Protected Area' do |username, password|
    md5_of_password = Digest::MD5.hexdigest(password)
    username == 'user' && md5_of_password == ENV['SIDEKIQ_HASH']
  end

  run Sidekiq::Web
end
```

Since the update to 4.1.1 I'm getting a `Forbidden` message from what looks like nginx when trying to retry jobs, delete jobs, kill jobs, etc. Login and everything else seems to work fine. Not sure what else to do. I've tried my best to follow the suggestions from people having similar issues but nothing has worked so far. 

Of note we're using AWS Elasticbeanstalk.

hobakill:
I've removed the
`Sidekiq::Web.use Rack::Session::Cookie, secret: ENV['RACK_SESSION_COOKIE']`
line and it's still giving me `Forbidden`

mperham:
There's several closed issues that show how to share the session between Rails and Sinatra.

```ruby
Sidekiq::Web.set :session_secret, Rails.application.secrets[:secret_token]
Sidekiq::Web.set :sessions, Rails.application.config.session_options
```

hobakill:
@mperham thanks for the consolidation. i really thought i put in a rather deep-dive on the closed issues. i didn't come up with such a concise answer. this seems to be working. again thank you for your help.

epintos:
I'm having the same issue with Elastic Beanstalk 64bit Amazon Linux 2016.03 v2.1.0 running Ruby 2.2.4 (Passenger Standalone) and Sidekiq 4.1.1.

PFStein:
@epintos I am a dev that works with @hobakill. I was investigating this issue after he opened a pull request into our application to add the new sidekiq configuration for 4.1. I noticed that the monitoring documentation and the code provided by @mperham uses `Sidekiq::Web.set :session_secret, Rails.application.secrets[:secret_token]` which is not actually standard for a rails 4 application. 

[...]

PR #3069, opened 2016-07-26
Migrate Sidekiq::Web to a pure Rack application

badosu:
Migrate Sidekiq::Web to a pure Rack application to avoid sinatra as
dependency. rack-protection is still needed.

The application is mounted on top of Rack::Builder, mantaining all of
the previous http interface.

Rack apps being used:

- Rack::File to serve assets
- Rack::Session::Cookie, the secret can be configured via
  Sidekiq::Web.session_secret
- Rack::Protection, same as before when using sinatra
- Sidekiq::WebApplication, described below.

Sidekiq::WebApplication is a very simple rack application composed of a
Sidekiq::WebRouter and a Sidekiq::WebAction dispatcher. This terminology
was adopted to be able to mantain Sidekiq::Web as a Rack app.

The Router is heavily inspired on Rack::Router[0] (and in many parts
identical, I don't know if this can be an issue), however not being retrocompatible.

The Action is a wrapper to provide convenience, DRY code and maintain
the old interface.

I tried to mantain most of the old application structure so that
customizations and monkey-patches are easily adjustable or even
further work be done to enforce retrocompatibility.

Testing welcome!

0: https://github.com/pjb3/rack-router

### DIFF
--- a/lib/sidekiq/web.rb
+++ b/lib/sidekiq/web.rb
@@ -1,26 +1,30 @@
 # frozen_string_literal: true
 require 'erb'
 require 'yaml'
-require 'sinatra/base'
 
 require 'sidekiq'
 require 'sidekiq/api'
 require 'sidekiq/paginator'
-require 'sidekiq/web_helpers'
+
+require 'sidekiq/web/router'
+require 'sidekiq/web/application'
+
+require 'rack/protection'
+
+require 'rack/builder'
+require 'rack/static'
+require 'rack/session/cookie'
 
 module Sidekiq
-  class Web < Sinatra::Base
-    include Sidekiq::Paginator
+  class Web
+    REQUEST_METHOD = 'REQUEST_METHOD'.freeze
+    PATH_INFO = 'PATH_INFO'.freeze
 
-    enable :sessions
-    use ::Rack::Protection, :use => :authenticity_token unless ENV['RACK_ENV'] == 'test'
-
-    set :root, File.expand_path(File.dirname(__FILE__) + "/../../web")
-    set :public_folder, proc { "#{root}/assets" }
-    set :views, proc { "#{root}/views" }
-    set :locales, ["#{root}/locales"]
-
-    helpers WebHelpers
+    ROOT = File.expand_path(File.dirname(__FILE__) + "/../../web")
+    VIEWS = "#{ROOT}/views"
+    LOCALES = ["#{ROOT}/locales"]
+    LAYOUT = "#{VIEWS}/layout.erb"
+    ASSETS = "#{ROOT}/assets"
 
     DEFAULT_TABS = {
       "Dashboard" => '',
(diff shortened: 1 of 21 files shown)


badosu:
Hmm.. It looks like `Binding#local_variable_set` is not available on Ruby 2.0.0. I am using it to set local variables when rendering partials.

Let me know if you have any suggestions, I can enforce this using eval, but I would like to avoid that.

mperham:
Holy cow, this is amazing.

mperham:
This is a huge change so a couple of notes:

0. Thank you for this work - it's really awesome.
1. This should probably go out as Sidekiq 5.0, which dovetails nicely with Rails 5.0.
2. Have you tested any Web UI extensions, like sidekiq-failures, sidekiq-cron, sidetiq, etc?  I'm ok with minimal breaking changes if necessary but we should make an effort not to break 3rd party gems.
3. The major area where people tweak Sinatra is for the session store: https://github.com/mperham/sidekiq/issues/2910#issuecomment-206978212.  Not sure what to do about that.

Looks like the test suite is passing so that's a huge step forward.

badosu:
Hi Mike, thanks for the kind words.

2. Where is this work located?
3. No, not yet, I mostly ever used `sidetiq` so I'll have to take a look at this later.
4. I had to tweak this as well in the past, sharing a session with a `Roda` application. I'll see where I can expose more options.

There are 2 important things to consider as well:

- I noticed there are no tests with regard to pagination and I could not test this in the UI yet, so it is the most fragile area.
- As you already pointed out, a lot of people customize their Sidekiq Web options so I'll pay special attention with these use cases

That said: the ERB templates should be cached and the configuration flexibility is still lacking. I just pulled this in an all-nighter so I'll take a look at these refinements later on.

mperham:
Sidekiq 5.0 does not exist yet.  This branch + the rails5 branch can be its start.

mperham:
pagination testing can be as simple as `100.times {|idx| HardWorker.perform_async(idx) }` then navigate to the default queue.

mperham:
rack-router is MIT licensed.  If a non-trivial amount of rack-router code is included, we need to add a Credits.txt with its MIT license.

mperham:
Sinatra's [hooks for registering extensions](http://www.sinatrarb.com/extensions.html) are used by Sidekiq Pro, Sidekiq Enterprise and a lot of 3rd party extensions.  We'll need to add support for them.  I can help but that appears to be the biggest blocker right now.

/Users/mike/src/pro/lib/sidekiq/pro/web.rb:119:in `<top (required)>': undefined method `register' for Sidekiq::Web:Class (NoMethodError)

```ruby
::Sidekiq::Web.register Sidekiq::Pro::Web
```

mperham:
I've merged these changes into the badosu-remove-sinatra branch.

mperham:
@badosu If you want committer access so you can polish these changes more, just let me know.

badosu:
@mperham Could you give me that access? I'll continue on those points tonight.

badosu:
@mperham I've made some improvements and the sidetiq page is already accessible. Some points though:

- It is not needed anymore to include explicitly `home_path` on redirect calls, as seen [here](https://github.com/tobiassvn/sidetiq/blob/master/lib/sidetiq/web.rb)
- I had to reinclude code to set local variables inside partials, however It is not compatible with ruby 2.0.0 and I can't see a way to do it without using `eval`. 
- `params` includes only the querystring parameters accessible via String keys instead of symbols
- `route_params` includes those that come from the route, accessible via symbols

We will have to decide what to deprecate and what not, fortunately it looks pretty straightforward to upgrade old extensions to the new model.

mperham:
Could you open up a new PR for the branch in this repo so we can see commits and discuss?

lib/sidekiq/web.rb
```diff
@@ -41,226 +45,56 @@ def custom_tabs
...
-      halt 404 unless params['key']
-      job = Sidekiq::DeadSet.new.fetch(*parse_params(params['key'])).first
-      retry_or_delete_or_kill job, params if job
-      redirect_with_query("#{root_path}morgue")
+      attr_accessor :app_url, :session_secret
+      attr_writer :locales
     end
 
+    def initialize
+      secret = Web.session_secret
```
mperham: This requires the user to set the session_secret manually.  Could we pull it out of the Rails app to make Sidekiq work without any user intervention, e.g.

```ruby

def retrieve_secret
  sec = nil
  sec ||= Web.session_secret
  sec ||= Rails.application.session_secret if defined?(::Rails) # not sure of the syntax in Rails 3.2, 4.x, 5.x...
  sec ||= SecureRandom.hex(64)
  sec
end
```

PR #3075, opened 2016-07-27
Migrate from Sinatra to a Rack application

badosu:
Continuing discussion from https://github.com/mperham/sidekiq/pull/3069

About extensions:

- It is not needed anymore to include explicitly `home_path` on redirect calls, as seen [here](https://github.com/tobiassvn/sidetiq/blob/master/lib/sidetiq/web.rb)
- I had to reinclude code to set local variables inside partials, however It is not compatible with ruby 2.0.0 and I can't see a way to do it without using `eval`. 
- `params` includes only the querystring parameters accessible via String keys instead of symbols
- `route_params` includes those that come from the route, accessible via symbols
- The `halt` call should be replaced to `next`, e.g. `halt 404 -> next 404`

We will have to decide what to deprecate and what not, fortunately it looks pretty straightforward to upgrade old extensions to the new model.

### DIFF
--- a/lib/sidekiq/web.rb
+++ b/lib/sidekiq/web.rb
@@ -1,26 +1,30 @@
 # frozen_string_literal: true
 require 'erb'
 require 'yaml'
-require 'sinatra/base'
 
 require 'sidekiq'
 require 'sidekiq/api'
 require 'sidekiq/paginator'
-require 'sidekiq/web_helpers'
+
+require 'sidekiq/web/router'
+require 'sidekiq/web/application'
+
+require 'rack/protection'
+
+require 'rack/builder'
+require 'rack/static'
+require 'rack/session/cookie'
 
 module Sidekiq
-  class Web < Sinatra::Base
-    include Sidekiq::Paginator
+  class Web
+    REQUEST_METHOD = 'REQUEST_METHOD'.freeze
+    PATH_INFO = 'PATH_INFO'.freeze
 
-    enable :sessions
-    use ::Rack::Protection, :use => :authenticity_token unless ENV['RACK_ENV'] == 'test'
-
-    set :root, File.expand_path(File.dirname(__FILE__) + "/../../web")
-    set :public_folder, proc { "#{root}/assets" }
-    set :views, proc { "#{root}/views" }
-    set :locales, ["#{root}/locales"]
-
-    helpers WebHelpers
+    ROOT = File.expand_path(File.dirname(__FILE__) + "/../../web")
+    VIEWS = "#{ROOT}/views"
+    LOCALES = ["#{ROOT}/locales"]
+    LAYOUT = "#{VIEWS}/layout.erb"
+    ASSETS = "#{ROOT}/assets"
 
     DEFAULT_TABS = {
       "Dashboard" => '',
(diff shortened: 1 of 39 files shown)


mperham:
* If you can't do it with Ruby 2.0, how does Sinatra do it?  Does it use eval?
* Ok about halt.  You should add a halt method that raises an informative error.

mperham:
I believe Sinatra uses throw/catch to implement halt.  Would it be possible to keep the same pattern?

badosu:
Do you have a list of known Sidekiq extensions so that I can take a look at what I need to do to support them all?

mperham:
You could look thru the Related Projects wiki page.

mperham:
That's great, do what you feel is right.  A change this big deserves a major version bump, since it'll for sure break someone, but do what you can to minimize the upgrade pain.

badosu:
Okay, I am almost finished now, besides all those points above I had to:

- Implement `content_type`
- Implement matchers on before and after hooks
- Change `add_to_head`
- `erb` implementation was tricker than I thought, as it was used for templates with and without layouts depending if it was called inside a template or note
- Add `settings`

These web extensions are passing the tests and are looking fine:

- [sidetiq](https://github.com/tobiassvn/sidetiq)
- [sidekiq-failures](https://github.com/mhfs/sidekiq-failures)
- [sidekiq-cron](https://github.com/ondrejbartas/sidekiq-cron)
- [sidekiq-monitor-stats](https://github.com/harvesthq/sidekiq-monitor-stats)
- [sidekiq-throttled](https://github.com/sensortower/sidekiq-throttled)
- [sidekiq-status](https://github.com/utgarda/sidekiq-status)

Needs further testing and adjust:

- [sidekiq-statistic](https://github.com/davydovanton/sidekiq-statistic) Tests not passing
- [sidekiq-scheduler](https://github.com/moove-it/sidekiq-scheduler) Uses `redirect to`
- [sidekiq_status](https://github.com/cryo28/sidekiq_status) Uses `redirect to`
- [sidekiq-enqueuer](https://github.com/richfisher/sidekiq-enqueuer)

Not supported (requires extensive changes):

- (sidekiq-benchmark)[https://github.com/kosmatov/sidekiq-benchmark] Uses sinatra-assetpack

Not tested yet:

- [sidekiq-prioritized_queues](https://github.com/publitas/sidekiq-prioritized_queues)

mperham:
Wow, you're incredibly thorough.  I'm impressed. 🍰 

I'll review soon and see what else needs to be done to make it work with Sidekiq Pro and Enterprise's UI extensions.

mperham:
Can you benchmark a few different pages so we can see how performance looks vs Sinatra?

badosu:
@mperham I don't know of It'll be a great difference as there aren't any optimizations and also I've sacrificed performance in favour of compatibility with `sinatra` (see `#params` implementation)

mperham:
Updated PR with changes to make Sidekiq Pro and Sidekiq Enterprise run green.  Notably it took me about an hour to figure out this little tidbit, necessary to run multiple Sidekiq::Web instances in the same process:

http://stackoverflow.com/questions/5643907/what-is-x-cascade-header

badosu:
Surprisingly enough, peformance is worse after this change.

Using a config.ru just mounting the application and thin, with `ab -l -c 4 -n 2000 http://0.0.0.0:3000/`

- ~400 req/s with Sinatra
- ~200 req/s new impl.
- ~300 req/s with a simple ERB cache

Possible causes:

- The Router uses a regexp match for even when it is not needed
- Tilt can have a better template caching system than the one I am implementing
- Rack::Builder

badosu:
I just updated the PR with 2 changes:

- Set content-length header, as I could not even test the application with ab without it and it is a tiny implementation
- Implement simple caching of erb templates

mperham:
An easy optimization to make is to split the routes based on method.

lib/sidekiq/web/action.rb
```diff
@@ -77,14 +80,20 @@ def json(payload)
     def initialize(env, block)
       @env = env
       @block = block
+      @@files ||= {}
     end
 
     private
 
     def _erb(file, locals)
+      # TODO Use Binding#local_variable_set if/when support for 2.0.0 is dropped
```
mperham: Can we switch based on Ruby version?  Seems like this might help performance.

lib/sidekiq/web/application.rb
```diff
@@ -278,8 +279,12 @@ def call(env)
           WebAction::TEXT_HTML
         end
 
-        [200, headers, [resp]]
+        [200, type_header, [resp]]
       end
+
+      resp[1][CONTENT_LENGTH] = resp[2].inject(0) { |l, p| l + p.bytesize }.to_s
```
mperham: You're mutating a shared data structure, not thread-safe.  Need to use `merge`.

lib/sidekiq/web/action.rb
badosu: I've benchmarked this already, the difference is not enough.

badosu:
I tweaked many things, and applied some optimizations, but I guess they are not as significant as the templating engine.

I am taking a look at how Tilt does it, I'll probably try using it for a POC to see if the bottleneck is indeed the rendering.

badosu:
Great, I changed the templating system to cache the compiled template, it now achieves ~ 490 req/sec. Considerably higher than Sinatra.

mperham:
I'm blown away by your work here.  Really impressive.

mperham:
We can merge this with the rails5 branch and call it Sidekiq 5.0 but I want to make a second pass next week to see if I can get it 100% compat with Pro and Enterprise so I don't need to cut new major versions of them - that would be huge plus.

mperham:
Since we're adding a bunch of new code, we need a few tests.

1. Verify the [BasicAuth](https://github.com/mperham/sidekiq/wiki/Monitoring#rails-http-basic-auth-from-routes) middleware works and stops unauthenticated requests.
2. Verify the CSRF protection stops POSTs without a proper authenticity_token.

badosu:
@mperham This should go in the settings hash, to use the same interface as sinatra sessions.

badosu:
@mperham I changed it so the user can specify it's own Rack::Session options if the secret alone is not enough.

I don't understand how this is related to github oauth.

badosu:
@mperham Feel free to add this to the examples section or update the wiki page when you want https://gist.github.com/badosu/5a6bb42cc0c0a7ae6e9b01313c61a5b9

badosu:
With the last change, `sidekiq-statistic` should work as well now

badosu:
@mperham I am pretty confident on the status to this implementation.

Let me know if there are any pending issues.

mperham:
I'm not aware of any issues.  It'll take me a week or two to get everything tested as this is a major release so be patient.

mperham:
To test this branch, put this in your Gemfile:

```ruby
gem 'sidekiq', github: 'mperham/sidekiq', branch: 'badosu-remove-sinatra'
```

lib/sidekiq/web.rb
```diff
@@ -41,227 +55,94 @@ def custom_tabs
...
-      )
-    end
+      unless using?(::Rack::Protection) || ENV['RACK_ENV'] == 'test'
+        middlewares.unshift [[::Rack::Protection, { use: :authenticity_token }], nil]
+      end
 
-    get '/stats/queues' do
-      queue_stats = Sidekiq::Stats::Queues.new
+      unless using? ::Rack::Session::Cookie
+        unless secret = Web.session_secret
```
badosu: @mperham I guess you wanted to comment here?

I am not sure what is exactly the use case, but if sharing the session is required I am not sure this will work at all, because since Rails 4 it uses it's own middleware.

Using `Sidekiq::Web::use` users can follow the approach used in these tutorials: https://robots.thoughtbot.com/how-to-share-a-session-between-sinatra-and-rails http://stderr.timfischbach.de/2013/09/14/rails-4-and-sinatra-session-sharing.html

However, they should have a way to prevent usage of the `::Rack::Session::Cookie` middleware, that is not implemented today.

If just reusing the same session secret is the use case, then I guess this would be a more straightforward way to configure it, as it enables setting more options directly:

```ruby
### condiftion if Rails
Sidekiq::Web.use Rack::Session::Cookie, secret: 'v3rys3cr31', host: 'nicehost.org'
###
```

We might use this other approach as well: http://ap4y.github.io/2013/11/03/integrating-multiple-ruby-web-applications.html

Let me know what are the use cases for this that I'll take a look at what we can do and how to best integrate with Rails.

johnathanludwig:
I'm finding two issues in my testing.

1. Passing the redis pool doesn't seem to be working. We mount two web instances. In development each instance points at a different DB on redis. Using this example:
  ```ruby
  mount Sidekiq::Pro::Web.with(redis_pool: ConnectionPool.new { Redis.new(url: 'redis://localhost:6379/1') }), at: "/sidekiq_primary", as: "/sidekiq_primary"
  mount Sidekiq::Pro::Web.with(redis_pool: ConnectionPool.new { Redis.new(url: 'redis://localhost:6379/2') }), at: "/sidekiq_secondary", as: "/sidekiq_secondary"
  ```
  When I go to `sidekiq_primary` or `sidekiq_secondary`, both show the redis instance is `redis://localhost:6379/0` instead of 1 or 2.

2. Chrome is caching the pages sometimes displaying an old response. I think this is because the ETag header is coming back the same even after making changes. In some cases the ETag is being refreshed though. Here are the steps I went through:

  1. Go to Queues tab.
  2. Queue job in one queue.
  3. Refresh page. Queue should show 1 job queued.
  4. Click on queue to view job.
  5. Delete job. Page should be refreshed and show job removed.
  6. Go back to queues. Should say 0 jobs.
  7. Delete queue. Page should be refreshed with queue removed.

  Performing those steps resulted should result in at least one of those expectations being incorrect due to the caching.

mperham:
The sharding issue is fixed on Pro master. I'll coordinate the gem releases to make sure this isn't a problem.

lib/sidekiq/web.rb
mperham: Thanks for those links, that's really helpful.  [mounting in config/routes.rb is the main use case](https://robots.thoughtbot.com/how-to-share-a-session-between-sinatra-and-rails#rails-routes) I wanted to handle; sounds like this is automatically handled by Rails.

mperham: Oh, maybe we shouldn't set up any session middleware at all by default.  Rails is the 90% use case and should provide one to us.  We can document that `config.ru` users need to set up a session handler and we could provide a simple method to do it for them, e.g. `Sidekiq::Web.default_session_handler!(secret)`.

mperham:
@badosu Any comments on (2) above?

mperham:
We shouldn't be loading any ETag middleware at all.  I don't really see much benefit for this specific webapp.

badosu:
@mperham If you're asking about the ETag issue, I don't know how it can happen if we control all headers and we're not passing any ETag response headers at all.

Only if this is something being setup by Pro or a middleware customization.

mperham:
Could this be part of the rails middleware stack when mounted in a rails app?

mperham:
Ok, gonna merge this as is and let it sit on master for a few weeks.

mperham:
Thank you @badosu for your incredible work on this PR.  It's a huge change and you've put an amazing amount of work into polishing it.  I hope it proves stable!

badosu:
@johnathanludwig Do you have any idea how we could reproduce this problem (the Etag one)?

johnathanludwig:
@badosu These were the steps I ran. Along the way something would end up being cached by the browser and not display the expected results:

1. Go to Queues tab.
2. Queue one job.
3. Refresh page. Queues should show 1 job queued.
4. Click on queue to view job.
5. Delete job. Page should be refreshed and show job removed.
6. Go back to queues. Should say 0 jobs.
7. Delete queue. Page should be refreshed with queue removed.

Issue #3070, opened 2016-07-26
Monitoring : no css/js/images because of violation of Content Security Policy directives


nicolasmlv:
Ruby version: 2.3.0
Sidekiq free version : 4.1.2
Sinatra 2.0.0.pre.alpha (master)
rack-protection / rack : 2.0

In my Rails app, I wanted to add the monitoring (https://github.com/mperham/sidekiq/wiki/Monitoring)

But there is some violation of "Content Security Policy directives", and the result is there is no JS, CSS, images.

But 👍 because it still works (except live polling).

<img width="1340" alt="screen shot 2016-07-26 at 18 01 50" src="https://cloud.githubusercontent.com/assets/1242610/17146222/e610e90e-535d-11e6-9b86-68dad5603cae.png">

<details>

Refused to load the stylesheet 'https://www.xxx.xx/sidekiq/stylesheets/bootstrap.css' because it violates the following Content Security Policy directive: "style-src self".

retries:7 Refused to load the stylesheet 'https://www.xxx.xx/sidekiq/stylesheets/application.css' because it violates the following Content Security Policy directive: "style-src self".

retries:8 Refused to load the image 'https://www.xxx.xx/sidekiq/images/favicon.ico' because it violates the following Content Security Policy directive: "default-src none". Note that 'img-src' was not explicitly set, so 'default-src' is used as a fallback.

retries:1 Refused to load the script 'https://www.xxx.xx/sidekiq/javascripts/application.js' because it violates the following Content Security Policy directive: "script-src self".

retries:1 Refused to load the script 'https://www.xxx.xx/sidekiq/javascripts/locales/jquery.timeago.en.js' because it violates the following Content Security Policy directive: "script-src self".

retries:394 Refused to apply inline style because it violates the following Content Security Policy directive: "style-src self". Either the 'unsafe-inline' keyword, a hash ('sha256-RplG6L4DS7tinB+kpxygtAVbINLXXNahDRzdfk1+cwY='), or a nonce ('nonce-...') is required to enable inline execution.

retries:1 Refused to load the image 'https://www.xxx.xx/sidekiq/images/favicon.ico' because it violates the following Content Security Policy directive: "default-src none". Note that 'img-src' was not explicitly set, so 'default-src' is used as a fallback.
</details>

mperham:
Sounds like your top-level website is setting some CSP rules that the Sidekiq Web UI does not follow, maybe?  I don't know much about CSP.

mperham:
[...]

PR #3079, opened 2016-07-28
Fix CSP errors on Sinatra / Rack 2

jamesdabbs:
See #3070 

Now that https://github.com/sinatra/rack-protection/pull/111 has been merged in, this should work for users pulling `sinatra` and `rack-protect` from master (e.g. edge Rails folks). It _should_ be backwards compatible, though I've only tested against Rails 4.2 and 5 personally.

@mperham - it seemed like you thought that Sinatra 2 should already be backwards compatible, so let me know if you'd prefer to fight this fight upstream. Also - is adding a method here a reasonable way to allow folks to configure if needed, or would you prefer something else?

### DIFF
--- a/lib/sidekiq/web.rb
+++ b/lib/sidekiq/web.rb
@@ -12,8 +12,21 @@ module Sidekiq
   class Web < Sinatra::Base
     include Sidekiq::Paginator
 
+
     enable :sessions
-    use ::Rack::Protection, :use => :authenticity_token unless ENV['RACK_ENV'] == 'test'
+
+    # N.B. until https://github.com/sinatra/rack-protection/pull/111 lands,
+    #  you'll need to uncomment the next line:
+    Rack::Protection::ContentSecurityPolicy::KEYS.push :img_src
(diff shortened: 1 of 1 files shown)


mperham:
This is something that should be pushed upstream.  I assume it will affect all Sinatra applications.  Sidekiq's Web UI is a basic webapp that needs to serve local static assets - if rack-protection breaks that model, there's a fundamental bug.

jamesdabbs:
That's fair - I've opened https://github.com/sinatra/rack-protection/pull/112 and https://github.com/sinatra/sinatra/pull/1156, which would resolve _most_ of this, so we'll see how those go.

That said, I doubt that `style-src: 'unsafe-inline'` would fly as an upstream default. Would you be open to a PR that removes the couple inline styles that you're currently using? Or, alternatively, one that adds `unsafe-inline` as a default `rack-protect` setting here?

mperham:
Perfect.

I would love a PR that removes inline CSS.  I want Sidekiq to work with rack-protection's defaults and best practices.

PR #3082, opened 2016-08-01
Fix typos in client test descriptions

davydovanton:


mperham:
Redi is my funny plural of Redis, so technically the spelling is correct.  If you want to change it to Redises, that would be fine.

davydovanton:
hahaha, no, It's so cute. may be I can add comment about this plural for other developers? What do you think?

mperham:
Sure

PR #3091, opened 2016-08-04
Removing Inline CSS

maryhowell:
Based on this [pull request ](https://github.com/mperham/sidekiq/pull/3079) this replaces the inline CSS with classes.


mperham:
🎉  Thanks!

PR #3107, opened 2016-08-18
Fixed the translation of "idle" for the de locale

bjjb:
"Inaktiv" has a specific connotation in German, which I suspect is not what we're looking to convey here.


mperham:
Thanks!

PR #3114, opened 2016-08-29
Raise error if required filed was not loaded

Tonkonozhenko:
When I was reviewing another's code why Sidekiq does not run, I found issue that it is launched:
`bundle exec sidekiq -r sidekiq.rb` instead of `bundle exec sidekiq -r ./sidekiq.rb`. It took some time because it raised an error about Redis but not launch configuration.

### DIFF
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -244,7 +244,7 @@ module Sidekiq
         end
         options[:tag] ||= default_tag
       else
-        require options[:require]
+        require options[:require] or raise ArgumentError, "#{options[:require]} was not required"
       end
     end
 
--- a/sidekiq.rb
+++ b/sidekiq.rb
@@ -1,0 +1,1 @@
+# Is used for testing purposes
--- a/test/test_cli.rb
+++ b/test/test_cli.rb
@@ -31,6 +31,14 @@ class TestCli < Sidekiq::Test
       assert @cli.valid?
     end
 
+    it 'does not require existing file' do
+      @cli.parse(['sidekiq', '-r', 'sidekiq.rb'])
+
+      assert_raises ArgumentError do
+        @cli.run
+      end
+    end
+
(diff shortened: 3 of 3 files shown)


sidekiq.rb
```diff
@@ -0,0 +1 @@
+# Is used for testing purposes
```
mperham: Don't put test files at the top of the project.

lib/sidekiq/cli.rb
```diff
@@ -244,7 +244,7 @@ def boot_system
         end
         options[:tag] ||= default_tag
       else
-        require options[:require]
+        require options[:require] or raise ArgumentError, "#{options[:require]} was not required"
```
mperham: I don't use and/or in Ruby.  Please use ||

sidekiq.rb
Tonkonozhenko: It can be reproduced only if required file is in root.

mperham: Ok, then remove the test.  I don't want the top-level mess.

lib/sidekiq/cli.rb
mperham: The error message should probably be more helpful in fixing the problem: "sidekiq.rb was not required, you should use an explicit path: ./sidekiq.rb or /path/to/sidekiq.rb"

mperham:
thank you!

devvit:
No, we cannot use 'require' like this.

Read http://ruby-doc.org/core-2.3.1/Kernel.html#method-i-require first, please.

```ruby
...
# "Loads the given name, returning true if successful and false if the feature is already loaded." from rubydoc
# You cannot decide file loaded or not with true|false
# So require(options[:require]) will be false forever
require(options[:require]) || raise(ArgumentError, not_required_message)
...
```

The correct way is still like before
```ruby
require(options[:require])
```

If you got errors about redis, you'd better check your code, but dont use 'require', :)

PR #3117, opened 2016-08-30
UI cleanup

mperham:
1. [x] Upgrade bootstrap to latest.
1. [x] Simplify images.
1. [x] Serve images with Cache-Control to assure speeeeeed.


mperham:
I really want to kill all the jquery.timeago locale files.  So much mess for so little gain.

shaneog:
@mperham If you switch to [moment.js](http://momentjs.com/), you can use a CDN with the locales 😄 

For example: https://cdnjs.com/libraries/moment.js/

Sadly, I don't know of any CDN that provides the jQuery.timeago locales.

mperham:
@shaneog I'd rather not depend on a CDN - that's another dependency that might break N years from now.

shaneog:
You can also just import one file (`moment-with-locales.min.js`) into the project instead, and update just that one file as needed.

mperham:
65k minimized to display "12 minutes ago".  Ugh, I don't know what to do.

shaneog:
[disTime](https://github.com/SimonWaldherr/disTime.js) is **tiny** and would only require two local files - but it doesn't have the broad language support that jQuery.timeago has.

mperham:
Looks like you just need disTime.min.js, as the locales seem to be included in it.  Looks pretty good.  I'd love a PR which switches to it.

shaneog:
Working on it 😉 
Although it'll probably be Monday before I can finish it, if that's ok?

mperham:
I'm on sort-of-vacation right now so take your time.  I'll be looking more at these cleanups later next week.

shaneog:
I've started it in #3123.

PR #3123, opened 2016-09-02
Replace jQuery.timeago

shaneog:
Replace the heavy jQuery.timeago plugin with a smaller library.

- [x] remove jQuery.timeago
- [x] add [timeago.js](https://github.com/hustcc/timeago.js)
- ~~update with new translations ([PR open](https://github.com/hustcc/timeago.js/pull/5) on source library)~~
- [x] use `timeago.js` build with all locales included - waiting on [this issue](https://github.com/hustcc/timeago.js/issues/23)
- [x] set JS locale based on Rails locale

The list of included locales can be found [here](https://github.com/hustcc/timeago.js/tree/master/locales).


mperham:
I had assumed the JS locale is provided by the browser on the client-side?

hustcc:
Use `register` to load new locales, then you can use the locales translation~

shaneog:
@mperham I could use the [navigator.language](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language) API but I figured that it would be better to align fully with Rails/web app.

mperham:
Good call.  Being consistent is more important than being right.

web/views/layout.erb
```diff
@@ -29,5 +28,10 @@
     </div>
     <%= erb :_footer %>
     <%= erb :_poll_js %>
+    <script type="text/javascript">
```
connorshea: Please don't use inline scripts, this'd make enforcing Content Security Policies a lot more difficult.

shaneog: Do you have any suggestions as to how I can pass the value to JavaScript without using an inline script?

connorshea: Is there a reason why adding this exact code to the application.js (or equivalent) doesn't work?

shaneog: We pass a variable from Rails to the JavaScript app to load the correct locale (`<%= locale %>`).
I don't see how we can have the JS app use the same locale as Rails without passing this through.

shaneog: Also, it's worth noting that [`_poll_js`](https://github.com/mperham/sidekiq/blob/master/web/views/_poll_js.erb) already uses inline JavaScript.

mperham: You could set an attribute on <body> and then have the JS use that attribute.

connorshea: Could you just add a data attribute to the body element and get the locale from that?

e.g. `<body class="admin" data-locale="<%= locale %>">` and then

```javascript
var locale = $('body').data('locale');
window.addEventListener('load', function() {
  timeago().setLocale(locale);
});
```

shaneog: Perfect - I'll do that, thanks!

mperham: If someone wants to send a PR changing the poll_js logic in the same manner, I'd love it.

connorshea: Can I open an issue to make sure it doesn't get forgotten in the mean time? :)

mperham: Absolutely, that's what it's there for!

shaneog:
OK, this PR is ready for review now. Sorry it took so long!

mperham:
I don't like the "live update" at all.  Can we turn it off?  It will lead to bug reports: "Next retry: 10 seconds ago".

mperham:
https://github.com/hustcc/timeago.js/pull/73

shaneog:
Absolutely. PR incoming

mperham:
Just did it.

PR #3153, opened 2016-09-20
Configurable :fetch strategy

antstorm:
Hi,

I'm using a custom fetch class in some of my Sidekiq processes and thought that it might be a good idea to expose it in CLI.

So this PR introduces a new `--fetch` CLI option that accepts a class name to be used for job fetching and also removes duplication around deciding which fetch class to use.

I haven't added the shorthand option because `-f` is usually used for files… it might make sense to rename this option to `strategy` with an `s` shorthand.

### DIFF
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -296,6 +296,10 @@ module Sidekiq
           opts[:environment] = arg
         end
 
+        o.on '--fetch FETCH', 'Custom fetch strategy class to use for picking up jobs from queues' do |arg|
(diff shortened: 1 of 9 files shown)


mperham:
Thank you but fetch isn't designed to change from the CLI but from Ruby.

antstorm:
@mperham in that case would it make more sense to remove it from `options` and make it a part of `Sidekiq` class?

mperham:
Sounds like busy work.  I see no appreciable difference or improvement.

PR #3158, opened 2016-09-22
Redirect from no trailing slash on Web UI

### LINKED ISSUE #3151: New Web UI Rack App doesn't respond to root URL without trailing slash
This is silly, but annoying only because I have to get my browser to re-learn a new favorite URL. Feel free to file this under a "whine" label!

Because of the new internal router on the Rack-ified Web UI, if you have it mounted on a URL like `/sidekiq`, it will not match for `http://whatever.com/sidekiq`. It instead needs the trailing slash: `http://whatever.com/sidekiq/`

Pretty silly and pedantic. Probably a quick fix in there somewhere (maybe a special case for the `/` route?). But it's the only annoyance I've had with the new code. Otherwise, it's pretty solid!

timdorr:
Fixes #3151

If you mount at `/sidekiq` and visit the `/sidekiq` URL, it will now redirect over to `/sidekiq/`, rather than 404ing.

### DIFF
--- a/lib/sidekiq/web/application.rb
+++ b/lib/sidekiq/web/application.rb
@@ -28,6 +28,10 @@ module Sidekiq
     def self.set(key, val)
       # nothing, backwards compatibility
     end
+    
+    get "" do
+      redirect(root_path)
+    end
 
     get "/" do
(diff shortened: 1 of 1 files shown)


mperham:
Thank you!

Issue #3145, opened 2016-09-16
Possible to unlock rack-protection version?


apolzon:
Ruby version:
Sidekiq / Pro / Enterprise version(s): 4.2.x

Was really happy to see the ability to remove the sinatra dependency and still use the sidekiq web-ui. However, I have a test-environment gem which still needs edge sinatra, and the updated sidekiq dependency of rack-protection ~> 1.5 causes a dependency conflict.

Edge sinatra currently has a minimum dependency of rack-protection 2.0.0.beta2. Is it possible that the new sidekiq web-ui code will be updated to support rack-protection 2.x? I totally understand if it doesn't happen before rack-protection moves to an official release, but it would be a bummer if any Rails 5+ app that needed sinatra was stuck on Sidekiq 4.1.x.

Thanks.

mperham:
If someone wants to test it, that's fine.  I don't generally support gems until they are out of beta but if it works, we can unlock the dependency.

PR #3162, opened 2016-09-27
Remove gemspec version constraint for rack-protection

bbulpett:
Addresses #3145 
Tested and currently in production use.
Deals with rack-protection Sinatra version conflict by removing version constraint from gemspec.


mperham:
Would you mind changing this to be ">= 1.5.0" so we don't try to work with ancient versions?

bbulpett:
Change has been made, we should be all set. Thanks

mperham:
thanks!

Issue #3157, opened 2016-09-22
Rails 5 Reloader Threading Issues


sj26:
We upgraded to Rails 5.0.0.1 and Sidekiq 4.2.1 + Sidekiq Pro 3.3.3 last week. :tada:

After the upgrade we started noticing threading issues in our Sidekiq workers.

Turns out [the new Rails reloader](https://github.com/mperham/sidekiq/blob/v4.2.1/lib/sidekiq/cli.rb#L243) does a lot of funny business which isn't entirely obvious including [ActiveRecord connection management](https://github.com/rails/rails/blob/v5.0.0.1/activerecord/lib/active_record/query_cache.rb#L46-L50), [QueryCache enable/disable](https://github.com/rails/rails/blob/v5.0.0.1/activerecord/lib/active_record/query_cache.rb#L26-L41) around each job, and [code reloading](https://github.com/rails/rails/blob/v5.0.0.1/railties/lib/rails/application/finisher.rb#L115-L175) with mutexes and shared locks. Some of these things are done by the Executor, not the Reloader, but they seem pretty interlinked with one calling the other. With several threads starting similar jobs with similar autoloaded dependencies we discovered workers were easily becoming deadlocked on code reloading, and we've also been experiencing stale QueryCache results within worker threads.

As a fix for now we've added a `:require: ./config/sidekiq.rb` to our `config/sidekiq.yml` with `config/sidekiq.rb` containing [the old Rails 4 behaviour](https://github.com/mperham/sidekiq/blob/v4.2.1/lib/sidekiq/cli.rb#L233-L239):

```ruby
require "./config/application.rb"
::Rails::Application.initializer "sidekiq.eager_load" do
  ::Rails.application.config.eager_load = true
end
require "sidekiq/rails"
require "./config/environment.rb"
```

The actual root cause of the threading issues is so far unclear, but I suspect lies within Rails 5's concurrency primitives. I found [ActiveSupport::Concurrency::ShareLock](https://github.com/rails/rails/blob/v5.0.0.1/activesupport/lib/active_support/concurrency/share_lock.rb) at the top of most locked threads' stack traces.

mperham:
Ugh, not surprising considering how complex this is but there's not much we can do about it in Sidekiq.  Please do open a rails/rails issue so that the core team knows this, especially if you have a sample app/script which can repro the behavior.

sj26:
[...]

PR #3166, opened 2016-09-28
Resolve Rails Reloader and ActiveRecord middleware incompatibility

sj26:
The new Rails reloader enables the query cache within every yield and expects to have access to the current connection that it enabled the query cache for at the end of the block to clear the cache and then release the connection. The Sidekiq ActiveRecord middleware which is added automatically releases the connection earlier, the query cache of another connection may instead be cleared, and the dirty query cache might be used by another worker.

This change makes sure that the ActiveRecord middleware cannot be used in tandem with the new Rails reloader (in case folks have added it to their middleware stacks explicitly) and removes it from the default middleware stack when the reloader is in play.

I couldn't figure out a way to raise the error from the middleware during startup. It seems the middleware is instantiated per job and so the raise only happens when a job is run? Not sure if there's a better way.

Fixes the query cache issue in #3157.

/cc @mperham @matthewd

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -150,10 +150,6 @@ module Sidekiq
     Middleware::Chain.new do |m|
       m.add Middleware::Server::Logging
       m.add Middleware::Server::RetryJobs
-      if defined?(::ActiveRecord::Base) && !(defined?(Sidekiq::Rails::Reloader) && Sidekiq.options[:reloader].is_a?(Sidekiq::Rails::Reloader))
-        require 'sidekiq/middleware/server/active_record'
-        m.add Sidekiq::Middleware::Server::ActiveRecord
-      end
     end
   end
 
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -150,7 +150,7 @@ module Sidekiq
     Middleware::Chain.new do |m|
       m.add Middleware::Server::Logging
       m.add Middleware::Server::RetryJobs
-      if defined?(::ActiveRecord::Base)
+      if defined?(::ActiveRecord::Base) && (!defined?(Sidekiq::Rails::Reloader) || Sidekiq.options[:reloader].is_a?(Sidekiq::Rails::Reloader))
         require 'sidekiq/middleware/server/active_record'
         m.add Sidekiq::Middleware::Server::ActiveRecord
       end
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -150,7 +150,7 @@ module Sidekiq
     Middleware::Chain.new do |m|
       m.add Middleware::Server::Logging
       m.add Middleware::Server::RetryJobs
-      if defined?(::ActiveRecord::Base) && (!defined?(Sidekiq::Rails::Reloader) || Sidekiq.options[:reloader].is_a?(Sidekiq::Rails::Reloader))
+      if defined?(::ActiveRecord::Base) && !(defined?(Sidekiq::Rails::Reloader) && Sidekiq.options[:reloader].is_a?(Sidekiq::Rails::Reloader))
         require 'sidekiq/middleware/server/active_record'
         m.add Sidekiq::Middleware::Server::ActiveRecord
       end
--- a/lib/sidekiq/middleware/server/active_record.rb
+++ b/lib/sidekiq/middleware/server/active_record.rb
@@ -7,7 +7,8 @@ module Sidekiq
           #
           # The reloader needs the active connection to clear its query cache
           # before releasing it otherwise other workers will see dirty query
-          # cache data.
+          # cache data. The reloader will then take care of clearing
+          # ActiveRecord connections as well.
           #
           # We need to make sure folks remove this middleware if they've added
           # it themselves.
--- a/lib/sidekiq/middleware/server/active_record.rb
+++ b/lib/sidekiq/middleware/server/active_record.rb
@@ -2,6 +2,21 @@ module Sidekiq
   module Middleware
     module Server
       class ActiveRecord
+        def initialize
+          # We can't use this middleware with the Rails reloader.
+          #
+          # The reloader needs the active connection to clear its query cache
+          # before releasing it otherwise other workers will see dirty query
+          # cache data.
+          #
+          # We need to make sure folks remove this middleware if they've added
+          # it themselves.
+          #
+          if defined?(Sidekiq::Rails::Reloader) && Sidekiq.options[:reloader].is_a?(Sidekiq::Rails::Reloader)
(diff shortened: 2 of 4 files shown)


lib/sidekiq/middleware/server/active_record.rb
```diff
@@ -2,6 +2,21 @@ module Sidekiq
...
+          # We can't use this middleware with the Rails reloader.
+          #
+          # The reloader needs the active connection to clear its query cache
+          # before releasing it otherwise other workers will see dirty query
+          # cache data.
+          #
+          # We need to make sure folks remove this middleware if they've added
+          # it themselves.
+          #
+          if defined?(Sidekiq::Rails::Reloader) && Sidekiq.options[:reloader].is_a?(Sidekiq::Rails::Reloader)
```
matthewd: Might be simpler to just throw this condition around the `clear_active_connections!` call... not sure.

lib/sidekiq/middleware/server/active_record.rb
```diff
@@ -2,6 +2,21 @@ module Sidekiq
...
+            raise ArgumentError, "Your are usign the Sidekiq ActiveRecord middleware and the new Rails 5 reloader which are incompatible. Please remove the ActiveRecord middleware from your Sidekiq middleware configuration."
```
matthewd: Typos: Your, usign

lib/sidekiq/middleware/server/active_record.rb
```diff
@@ -2,6 +2,21 @@ module Sidekiq
...
       class ActiveRecord
+        def initialize
+          # We can't use this middleware with the Rails reloader.
+          #
+          # The reloader needs the active connection to clear its query cache
+          # before releasing it otherwise other workers will see dirty query
+          # cache data.
+          #
+          # We need to make sure folks remove this middleware if they've added
+          # it themselves.
```
matthewd: Probably worth an explicit acknowledgement that this middleware is unnecessary because the reloader will do the job for us -- that it's not "incompatible so you're trading one feature for the other", it's "incompatible because duplication"

sj26:
The test failure looks like an unrelated jruby concurrency issue?

lib/sidekiq.rb
```diff
@@ -150,7 +150,7 @@ def self.default_server_middleware
     Middleware::Chain.new do |m|
       m.add Middleware::Server::Logging
       m.add Middleware::Server::RetryJobs
-      if defined?(::ActiveRecord::Base)
+      if defined?(::ActiveRecord::Base) && (!defined?(Sidekiq::Rails::Reloader) || Sidekiq.options[:reloader].is_a?(Sidekiq::Rails::Reloader))
```
matthewd: Missing a `!`, I think?

lib/sidekiq/middleware/server/active_record.rb
sj26: I did that initially but moved it here hoping that it might raise earlier in the Sidekiq boot, but sadly it only raises during a job run. Happy to move it back if it seems more sensible.

lib/sidekiq/middleware/server/active_record.rb
sj26: Ugh.

lib/sidekiq.rb
sj26: Yep, despite my testing that slipped through!

lib/sidekiq/middleware/server/active_record.rb
sj26: Yep, good call.

lib/sidekiq.rb
sj26: I've combined the conditional to match the one in the middleware and negated the whole thing.

lib/sidekiq/middleware/server/active_record.rb
sj26: Moved it back.

lib/sidekiq/middleware/server/active_record.rb
sj26: (I'd put it in the commit message but not in the comment!)

mperham:
I think at boot time the code should either add the middleware to the default server-side list or activate the Reloader.  Mutually exclusive.

sj26:
@mperham that's what we're doing?

Although I've just realised that initialize has already happened by the time the reloader gets set, so if the user customises their middleware stack which lazily resolves the default server midddleware then it will have included the ActiveRecord middleware anyway.

There's a chicken and egg. We need to change initialization based on the presence of the reloader, but the reloader can't be present until we've initialized. I'll investigate.

mperham:
Yep, all this logic should go in sidekiq/rails.rb or cli.rb.

sj26:
How about this version using railtie hooks within `sidekiq/rails.rb`?

It's split because we need to make sure the default middleware is in place as soon as possible before user code which might change middleware has run, and the reloader needs to come quite late in case cache_classes is changed (although the only caveat there should be a missed log entry).

This setup should allow users to change middleware in `config/environments/*` or `config/initializers/*` just fine. If they're changing it in `config/application.rb` then we're buggered.

mperham:
Looks pretty good to me now.  Is the runtime check in the middleware still necessary?  That's constant overhead that the user will have to pay for every job executed.

sj26:
I've changed the middleware hook to even earlier which hopefully might even cover cases where sidekiq configuration is performed in config/application.rb.

The runtime check in the middleware is a just-in-case for folks who might have explicitly added the ActiveRecord middleware themselves. If you think this is a remote enough chance then maybe we could drop it?

Oh, _or_ we could move the check to when we setup the reloader and explode there if we detect the ActiveRecord middleware in the middleware chain. I'll draft that up.

lib/sidekiq/rails.rb
```diff
@@ -59,8 +59,13 @@ class Rails < ::Rails::Engine
     # from config/environment.rb back to sidekiq/cli.rb.
     config.after_initialize do
       if ::Rails::VERSION::MAJOR >= 5
-        # The reloader also takes care of ActiveRecord
-        Sidekiq.options[:reloader] = Sidekiq::Rails::Reloader.new
+        # The reloader also takes care of ActiveRecord but is incompatible with
+        # the ActiveRecord middleware so make sure it's not in the chain already.
+        if defined?(Sidekiq::Middleware::Server::ActiveRecord) && Sidekiq.server_middleware.exists?(Sidekiq::Middleware::Server::ActiveRecord)
+          raise ArgumentError, "You are using the Sidekiq ActiveRecord middleware and the new Rails 5 reloader which are incompatible. Please remove the ActiveRecord middleware from your Sidekiq middleware configuration."
```
sj26: I'm not sure if this should `raise`, `Sidekiq.logger.warn`, `Rails.logger.warn`, or `ActiveSupport::Deprecation.warn`. Thoughts?

PR #3170, opened 2016-09-30
Fix the error in worker#perform_in about interval type

tzwm:
## Situation:
We are using Sidekiq Pro with ActiveSupport.
When I passed a object which is ActiveSupport::TimeWithZone to perform_at,
`TypeError: can't convert ActiveSupport::TimeWithZone into an exact
number` has occurred.

## Problem:
The object of Time class can't plus a ActiveSupport::TimeWithZone object.

## Solution:
We can transform any Time object to float, and use it to compare and
calculate.


mperham:
Thank you!

Issue #3180, opened 2016-10-06
Sidekiq::Web + Wildcard domains regression


inkstak:
Upgrading from Sidekiq 4.1 to Sidekiq 4.2 induce regression of #2730 

To refresh your memory - We're using wildard domains in Rails
```Dummy::Application.config.session_store :cookie_store, domain: :all```

after authentication (with Devise), Sidekiq::Web create two cookies:
* _foo @ .localhost.dev
* _foo @ jobs.localhost.dev

Since Sinatra dependency has been removed, the suggested fix in #2730 doesn't work.

I've also suggested a fix in https://github.com/mperham/sidekiq/issues/2555#issuecomment-149594650 which doesn't work too.

@mperham you tried to broach the subject in #3147, but the issue is already closed.

So what's the proper solution today to use authentication ?
I'm still trying to fix it on my side, but advices are welcome :)

**Comment by inkstak** [comment]
I've digged into my problem and have few ideas.

In my case, I use Devise to authenticate :
```
Jobs::Engine.routes.draw do
  authenticate :user, lambda { |u| u.super_hero? } do
    mount Sidekiq::Web, at: '/'
  end
end
```

Devise relies on Warden to authenticate and use the session provided by Rails.
It doesn't care about the session initiated by Sidekiq::Web in [lib/sidekiq/web.rb#L116](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/web.rb#L116).
This second session becomes useless, unless it causes problems by creating a second cookie.

To fix this, I monkeypatched the `build` method and removed references to `Rack::Protection` and `Rack::Session::Cookie`. It works like a charm.

I'm preparing a pull request to disable sessions like Sinatra.

mperham:
I suspected that adding our own session always wouldn't work sometimes but it's hard to auto-detect this case.  Looking forward to your PR.

mperham:
I'll fix this on Monday if you haven't sent a PR by then.  It's blocking a 4.2.3 release.

badosu:
@inkstak Fixed?

Issue #2487, opened 2015-08-13
Forbidden when clicking on retry


Micka33:
# The use case

My sidekiq web interface, and sidekiq process, are 2 different rails projects (both connecting to the same redis instance obviously).  

When in `production mode` a weird behavior happens.
If one of the jobs fails, and I want to manually retry it, using any of these buttons ![screen shot 2015-08-13 at 11 14 18](https://cloud.githubusercontent.com/assets/11553456/9246556/b0e60f2a-41ac-11e5-9fa8-1235d650ae36.png) ![screen shot 2015-08-13 at 11 14 36](https://cloud.githubusercontent.com/assets/11553456/9246558/b0e662d6-41ac-11e5-8c34-55b8f61eea08.png), I am redirect on a white page with only `Forbidden` on it, and my devise session is destroyed.
This does not happen in `development mode`.

```
URL when clicking retry: http://localhost:3000/sidekiq/retries/1439457271.0547059-2294ccdf2a6c8594801f1dfc
page content: Forbidden
```

## Environment
```ruby
Ubuntu 14.04.2 LTS
ruby 2.2.1p85
rails (4.2.3)
sidekiq (3.4.2)
  celluloid (~> 0.16.0)
  connection_pool (~> 2.2, >= 2.2.0)
  json (~> 1.0)
  redis (~> 3.2, >= 3.2.1)
  redis-namespace (~> 1.5, >= 1.5.2)
devise (3.5.2)
   bcrypt (~> 3.0)
   orm_adapter (~> 0.1)
   railties (>= 3.2.6, < 5)
   responders
   thread_safe (~> 0.1)
   warden (~> 1.2.3)
```

## Initializer
```ruby
template = ERB.new File.new(File.join(Rails.root, 'config', 'sidekiq.yml')).read
raw_yaml = YAML.load template.result(binding)
SIDEKIQ_CONFIG = raw_yaml.merge(raw_yaml[Rails.env])

Sidekiq.default_worker_options = { 'backtrace' => true }

Sidekiq.configure_client do |config|
  config.client_middleware do |chain|
    chain.add Sidekiq::Status::ClientMiddleware
  end
  config.redis = {
      :url       => SIDEKIQ_CONFIG[:url],
      :namespace => SIDEKIQ_CONFIG[:namespace],
      :size      => SIDEKIQ_CONFIG[:client_connections]
  }
end

require 'sidekiq/web'
require 'sidetiq/web'
require 'sidekiq-status/web'
```

[...]

PR #3183, opened 2016-10-08
Allow to disable sessions

inkstak:
According to #3180, this pull request allows to disable session.

The pull request is not ready to be merged : the test I've added won't pass, when executed along with the full test suite.
I'm not yet familiar with rack tests and looking for isolated sessions between tests.

### DIFF
--- a/lib/sidekiq/web.rb
+++ b/lib/sidekiq/web.rb
@@ -105,6 +105,32 @@ module Sidekiq
       @app ||= build
     end
 
+    def enable(*opts)
+      opts.each {|key| set(key, true) }
+    end
+
+    def disable(*opts)
+      opts.each {|key| set(key, false) }
+    end
+
+    def set(attribute, value)
+      send(:"#{attribute}=", value)
+    end
+
+    # Default values
+    set :sessions, true
+
+    attr_writer :sessions
+
+    def sessions
+      unless instance_variable_defined?("@sessions")
+        @sessions = self.class.sessions
+        @sessions = @sessions.to_hash.dup if @sessions.respond_to?(:to_hash)
+      end
+
+      @sessions
+    end
+
     def self.register(extension)
       extension.registered(WebApplication)
     end
@@ -119,7 +145,7 @@ module Sidekiq
 
     def build_sessions
       middlewares = self.middlewares
-      sessions = Web.sessions
+      sessions = self.sessions
 
       return if sessions === false
 
--- a/lib/sidekiq/web.rb
+++ b/lib/sidekiq/web.rb
@@ -133,7 +133,7 @@ module Sidekiq
           secret = SecureRandom.hex(64)
         end
 
-        options[:secret] = secret
+        options = { secret: secret }
         options = options.merge(sessions.to_hash) if sessions.respond_to? :to_hash
 
         middlewares.unshift [[::Rack::Session::Cookie, options], nil]
--- a/lib/sidekiq/web.rb
+++ b/lib/sidekiq/web.rb
@@ -63,6 +63,22 @@ module Sidekiq
         @views ||= VIEWS
       end
 
+      def enabled_options
+        @enabled ||= %w(sessions)
+      end
+
+      def enabled?(key)
+        enabled_options.include?(key.to_s)
+      end
+
+      def enable(key)
+        enabled_options.push(key.to_s) unless enabled?(key)
+      end
+
+      def disable(key)
+        enabled_options.delete(key.to_s)
+      end
+
       # Helper for the Sinatra syntax: Sidekiq::Web.set(:session_secret, Rails.application.secrets...)
       def set(attribute, value)
         send(:"#{attribute}=", value)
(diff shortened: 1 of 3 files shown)


mperham:
Ideally your API is backwards compatible with < 4.2, like [this comment shows](https://github.com/mperham/sidekiq/issues/2487#issuecomment-205262001).  Maybe something like:

```ruby
Sidekiq::Web.set :sessions, false
```

badosu:
@inkstak Let me know if you need any help

inkstak:
So, you can disable sessions 
```
Sidekiq::Web.disable :sessions
```
or
```
Sidekiq::Web.set :sessions, false
```
You can also customize session options:
```
Sidekiq::Web.set :sessions, { domain: 'all' }
```

test/test_web.rb
```diff
@@ -663,4 +663,46 @@ def app
...
   end
+
+  describe 'sidekiq web sessions options' do
+    include Rack::Test::Methods
+
+    def app
+      Sidekiq::Web.new
+    end
+
+    after { Sidekiq::Web.enable(:sessions) }
```
badosu: @inkstak For tests that change the state of the application I recommend changing the `app` directly, as you can see it is an instance of `Sidekiq::Web`.

You can port these methods for usage on the instance and test them here. If the implementation on the instance level works and we assume that the instance implementation is a proxy of the class one (or vice-versa) then it's fine.

lib/sidekiq/web/helpers.rb
```diff
@@ -173,7 +173,7 @@ def display_args(args, truncate_after_chars = 2000)
     end
 
     def csrf_tag
-      "<input type='hidden' name='authenticity_token' value='#{session[:csrf]}'/>"
+      "<input type='hidden' name='authenticity_token' value='#{session[:csrf]}'/>" if session
```
mperham: I don't believe this should ever be disabled.  Can you explain why this is necessary?

mperham:
Looking much better, let me know when you consider the PR complete.

test/test_web.rb
inkstak: I'm not comfortable with that. At the end it's still `Sidekiq::Web` to mount.
Adding instance methods adds complexity for the sole purpose to be tested.

That's my opinion. If you prefer tests with instance methods, the commit is ready.

inkstak: .. the commit is ready, one push behind !

lib/sidekiq/web/helpers.rb
inkstak: Test artefact, removed.

inkstak:
The PR looks great for me, the last question is about refactoring the tests or not.

test/test_web.rb
badosu: @inkstak I agree with you, but they are not for the sole purpose of testing.

The ideal design is to have only instance methods actually, the class methods only exist to provide retrocompatibiity with the old implementation, maybe we can change that in the future, depending on how @mperham sees it.

The issue with configuring `Sidekiq::Web` directly is that this change will replicate to the subsequent tests, unless you undefine the constant and reload the code.

mperham: I'm always in favor of instance methods over class-level.  The Sinatra/Rack API is bad in that regard.  I'd much prefer to see something like:

```ruby
web = Sidekiq::Web.new
web.something = ...
run web
```

badosu: @mperham Is up to you if you want to introduce this change, my initial design prioritized retrocompatibility so that users of the old Sinatra API would'nt suffer with the migration.

badosu: @inkstak Feel free to push, and thanks for all the attention.

mperham: No, you are right.  Compatibility is #1.  But that doesn't mean we can't have instance methods for testing.

lib/sidekiq/web.rb
```diff
@@ -97,6 +105,32 @@ def app
...
+
+    def sessions
+      unless instance_variable_defined?("@sessions")
+        @sessions = self.class.sessions
+        @sessions = @sessions.to_hash.dup if @sessions.respond_to?(:to_hash)
+      end
+
+      @sessions
+    end
+
```
inkstak: If you want to add settings on instance level, which inherits from class,
you can use that :



```
    %w(sessions session_secret app_url redis_pool).each do |method_name|
      attr_writer method_name

      define_method(method_name) do
        var = "@#{method_name}"

        unless instance_variable_defined?(var)
          value = self.class.send(method_name)
          value = value.to_hash.dup if value.respond_to?(:to_hash)
          instance_variable_set(var, value)
        end

        instance_variable_get(var)
      end
    end
```

mperham:
This looks great - I'm inclined to merge as is.  Any other opinions?

inkstak:
I'm waiting for it. Thanks for your support !

badosu:
:tada: @inkstak Thanks!

mperham:
This has the side effect that if sessions are disabled, CSRF protection is disabled.  Users should not be able to disable rack-protection.

mperham:
One more issue: subclasses of Sidekiq::Web are not inheriting the value set for sessions, etc.

devlinzed:
This should be in the FAQ

mperham:
Then add it.

PR #3193, opened 2016-10-16
Update the fuzzy timestamps during polling

shaneog:
The timestamps were not updated during live polling since
cd3d18b81e9bd81fdbec440adb0d6709128c412a.
This causes the timestamps to be updated with fuzzy timestamp text when
live polling is turned on.


mperham:
ty!

Issue #3154, opened 2016-09-20
Scheduled jobs disappearing without a trace


dabrorius:
Ruby version: 2.2.4
Sidekiq / Pro / Enterprise version(s): Sidekiq 4.1.2 (switched to Sidekiq Pro 3.3.3. which does not seem to be affected)

For the last few weeks, we have been having problems with disappearing Sidekiq jobs. We’re using Rails v4.2.7 and Sidekiq 4.1.2.

We have implemented tracking of jobs through the database to help us identified our issues. We create a new record when a job is enqueued and then update it when executing a job. This is implemented using ActiveJob’s before_enqueue and around_perform callbacks.

From 13th September to 16th of September we have logged 7 cases of disappearing jobs (out of over 10000 tracked jobs). In the Rails log, we can see that the job was successfully enqueued, however, Sidekiq log has no mention of the job with that ID anywhere in it.

Rails Log:
```
Sep 13 09:56:28 [166]: Started POST "/api/webhooks/google_calendar_push_notification" for x.x.x.x. at 2016-09-13 14:56:28 +0000
Sep 13 09:56:28 [166]: Processing by API::WebhooksController#google_calendar_push_notification as */*
Sep 13 09:56:28 [166]: Calendar Load (0.3ms)  SELECT  "calendars".* FROM "calendars" WHERE "calendars"."id" = $1 LIMIT 1  [["id", 163]]
Sep 13 09:56:28 [166]: [ActiveJob]    (0.2ms)  BEGIN
Sep 13 09:56:28 [166]: [ActiveJob]   SQL (0.3ms)  INSERT INTO "background_job_statuses" ("job_id", "job_name", "job_argument_id", "job_argument_type", "enqueued_at", "created_at", "updated_at") VALUES ($1, $2, $3, $4, $5, $6, $7) RETURNING "id"  [["job_id", "7b124d00-69bc-410d-a52c-4a3c2e0b7f3f"], ["job_name", "SyncWithCalendarChangesJob"], ["job_argument_id", 163], ["job_argument_type", "Calendar"], ["enqueued_at", "2016-09-13 14:56:28.948052"], ["created_at", "2016-09-13 14:56:28.949949"], ["updated_at", "2016-09-13 14:56:28.949949"]]
Sep 13 09:56:28 [166]: [ActiveJob]    (1.6ms)  COMMIT
[...]

PR #3212, opened 2016-10-29
Handle errors raised by the reloader

eugeneius:
The reloader might raise an error instead of yielding control. If this happens, the job will silently fail to run, without being reenqueued and without logging any indication of what happened. I'm pretty sure that's what was happening here: https://github.com/mperham/sidekiq/issues/3154#issuecomment-253125379.

This change doesn't prevent the job from being lost, but at least if the error handlers are called it will be possible to diagnose the problem.

### DIFF
--- a/lib/sidekiq/processor.rb
+++ b/lib/sidekiq/processor.rb
@@ -119,9 +119,9 @@ module Sidekiq
       jobstr = work.job
       queue = work.queue_name
 
-      @reloader.call do
-        ack = false
-        begin
+      ack = false
+      begin
+        @reloader.call do
           job = Sidekiq.load_json(jobstr)
           klass  = job['class'.freeze].constantize
           worker = klass.new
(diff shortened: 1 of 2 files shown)


mperham:
Do you have an example of an error the reloader might raise?

mperham:
thank you for the fix!

eugeneius:
[This line](https://github.com/rails/rails/blob/v5.0.0.1/activerecord/lib/active_record/query_cache.rb#L27) is run by the reloader before yielding, and will raise if the database is unavailable.

It's fairly straightforward to reproduce the problem:

1. Create a new Rails app that uses MySQL or Postgres
2. Add Sidekiq, and a worker that loads Active Record (just reference `ActiveRecord::Base`)
3. Run Sidekiq in the foreground, and enqueue some jobs from a Rails console
4. Observe that the first job runs fine, but subsequent jobs fail silently

The first job running successfully is weird, and arguably a Rails bug, but not directly relevant.

eugeneius:
Oh, I forgot step 0: stop (or don't start) the relevant database 😛

Issue #3223, opened 2016-11-06
Can't log exception raise in worker


john1king:
Ruby version: 2.3.1
Sidekiq / Pro / Enterprise version(s): 4.2.5

See here https://github.com/mperham/sidekiq/blob/v4.2.5/lib/sidekiq/processor.rb#L125, job variable is in the block and a method use same name. So when exception happened, job will be a Sidekiq::BasicFetch::UnitOfWork object.

I try to hack the logger class

```ruby
class Sidekiq::ExceptionHandler::Logger

  def call(ex, ctxHash)
    puts ctxHash
    Sidekiq.logger.warn(Sidekiq.dump_json(ctxHash)) if !ctxHash.empty?
    Sidekiq.logger.warn "#{ex.class.name}: #{ex.message}"
    Sidekiq.logger.warn ex.backtrace.join("\n") unless ex.backtrace.nil?
  rescue Exception => error
    puts error
  end

end
```

and output 

```ruby
# ctxHash
{:context=>"Job raised exception", :job=>#<struct Sidekiq::BasicFetch::UnitOfWork queue="queue:default", job="{\"class\":\"TestWorker\",\"args\":[],\"retry\":true,\"queue\":\"default\",\"jid\":\"172edc462c0e42df73da5b45\",\"created_at\":1478444732.198344,\"enqueued_at\":1478444774.4269462,\"error_message\":\"xxxxxx\",\"error_class\":\"RuntimeError\",\"failed_at\":1478444734.424051,\"retry_count\":0}">, :jobstr=>"{\"class\":\"TestWorker\",\"args\":[],\"retry\":true,\"queue\":\"default\",\"jid\":\"172edc462c0e42df73da5b45\",\"created_at\":1478444732.198344,\"enqueued_at\":1478444774.4269462,\"error_message\":\"xxxxxx\",\"error_class\":\"RuntimeError\",\"failed_at\":1478444734.424051,\"retry_count\":0}"}

# error
stack level too deep
```

mperham:
I don't understand this issue.  job should always be a Hash.

eugeneius:
Looks like I broke this in https://github.com/mperham/sidekiq/pull/3212 🙈 working on a fix now.

PR #3224, opened 2016-11-06
Pass job hash to handle_exception again

eugeneius:
Fixes https://github.com/mperham/sidekiq/issues/3223.

When I moved the reloader inside the block in https://github.com/mperham/sidekiq/pull/3212 so that any errors it raised would be handled properly, the `job` local variable was pushed into a nested scope, which meant it wasn't accessible from the rescue block any more. This changed the meaning of `job` in that rescue block from the local variable to [the `attr_reader` with the same name](https://github.com/mperham/sidekiq/blob/v4.2.5/lib/sidekiq/processor.rb#L30).

We don't need to reload the application before parsing the job payload, so we can move this work outside the reloader block so that the job hash is accessible in the rescue block again.

### DIFF
--- a/lib/sidekiq/processor.rb
+++ b/lib/sidekiq/processor.rb
@@ -121,8 +121,8 @@ module Sidekiq
 
       ack = false
       begin
+        job = Sidekiq.load_json(jobstr)
         @reloader.call do
-          job = Sidekiq.load_json(jobstr)
           klass  = job['class'.freeze].constantize
           worker = klass.new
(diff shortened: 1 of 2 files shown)


mperham:
Thanks, I see the problem now.  We should probably rename the local variable so there's no chance it will alias with the attribute in the future.

eugeneius:
I agree! 👍 https://github.com/mperham/sidekiq/pull/3225

PR #3234, opened 2016-11-09
Add fallback for 'retry' value server-side.

mlarraz:
If using a client other than Sidekiq to enqueue jobs,
the `retry` value set in the `sidekiq_options` hash will be ignored.

This is somewhat counterintuitive, so fall back to value if no value
is set in the message.


mperham:
Thank you!  I've integrated this into #3235 so it will be in 4.3.0.

PR #3235, opened 2016-11-09
Rework job processing in light of Rails 5's Reloader

mperham:
As mentioned in #3221, we need to redesign Sidekiq::Processor's job execution so it can integrate Rails 5's Reloader.  The Reloader must run in all environments as it handles autoloading, db connection management, etc.

This is ready for alpha testing if anyone wants to try it out:

```ruby
gem 'sidekiq', github: 'mperham/sidekiq', branch: 'redesign_dispatch'
```


mperham:
Status: I'm not sure I want to merge this.  4.2.7 has a hack that works and doesn't restructure the internals (which will break various 3rd party plugins).  Will leave here for a while and consider.

matthewd:
FWIW, I think you should do this. As you've noted, the middleware API gets passed the worker, which means all sorts of things have to happen / be loaded before that can occur. Even ignoring the specific/additional dangers of the Rails reloader, that's a fair bit of potential to leave a job silently dropped, without retry or log.

#3221 solves the specific problem of the reloader + production, but that doesn't help development mode (where load-time errors in worker classes aren't terribly surprising, and while job loss in development is hardly catastrophic, it can certainly be inconvenient)... or even other errors in production. (`cache_classes=true`/`eager_load=false` is not an unheard-of production configuration, which leaves that innocent-looking `constantize` call pretty busy.)

Not to mention the potential dangers of invoking the worker class's user-defined `initialize` without any retry protection.

mperham:
@matthewd Well said.  👍

lastobelus:
Is the hack in 4.2.X a source of fsevent_watch zombies by any chance? My rails 5 apps with sidekiq steadily produce fsevent_watch zombies and I have to kill 9 them and restart spring every few hours.

jamesgecko:
Are the [Middleware docs](https://github.com/mperham/sidekiq/wiki/Middleware) still valid since this change?

mperham:
No, that's all 4.x specific but the API is still the same. 5.0 no longer has any middleware by default.

Issue #3047, opened 2016-06-30
Cron jobs can stop firing when Redis goes down


mperham:
When Redis goes down, the periodic job structure can lose loops due to destructive operations on the structure in between network failures.

PR #3255, opened 2016-11-23
Put source encoding comment as line for [j]ruby 1.9 compatibility

nilbus:
In jruby 1.7.22 (1.9.3p551 compatibility mode), UTF-8 encoding is not properly
detected, because the encoding comment is not on the first line as required in ruby
1.9.

The frozen_string_literal magic comment did not come into existence until ruby 2.3,
and ruby 1.9 does not look past the first line for magic comments.

This results encoding-related syntax errors. Examples:

    SyntaxError: /home/nilbus/ws/rental_express/ROOT/rails/vendor/bundle/jruby/1.9/gems/sidekiq-4.2.6/lib/sidekiq.rb:52: Invalid char `\235' ('') in expression
      def self.â¨â¯Â°â¡Â°â©â¯ï¸µâ»ââ»
                ^

    SyntaxError: /home/nilbus/ws/rental_express/ROOT/rails/vendor/bundle/jruby/1.9/gems/sidekiq-4.2.6/lib/sidekiq/api.rb:269: Invalid char `\237' ('') in expression
        alias_method :ð£, :clear
                       ^

This patch should restore compatibility with ruby 1.9 and greater.

Can you release this fix quickly? It's blocking my ability to upgrade to
sidekiq-ent-1.2.3 to resolve our issue related to #3047. Thanks!

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -1,5 +1,5 @@
-# frozen_string_literal: true
 # encoding: utf-8
+# frozen_string_literal: true
 require 'sidekiq/version'
 fail "Sidekiq #{Sidekiq::VERSION} does not support Ruby versions below 2.0.0." if RUBY_PLATFORM != 'java' && RUBY_VERSION < '2.0.0'
 
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -1,5 +1,5 @@
-# frozen_string_literal: true
 # encoding: utf-8
+# frozen_string_literal: true
 require 'sidekiq'
 
 module Sidekiq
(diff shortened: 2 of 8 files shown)


mperham:
I'll accept this but you should know that sidekiq 4 does not officially support jruby 1.7 and has not supported Ruby 1.9 for years.

nilbus:
Thank you. We'll be on JRuby 9000 soon, thankfully.

nilbus:
Given the lack of support, should I expect to see a release for this? If not, I'd like to know so that I can install from GitHub rather than a released gem.

mperham:
It'll be in 4.2.7 which will be out Monday.

nilbus:
Excellent. Thanks.

Issue #3245, opened 2016-11-18
Add enqueued_at when fake Testing


Ferdy89:
Ruby version: 2.2.4
Sidekiq / Pro / Enterprise version(s): Sidekiq 4.1.2, Sidekiq Pro 3.3.3, Sidekiq Ent 1.3.2

This is a feature request, not a bug.

While running some tests, we realized queuing jobs while on fake testing mode doesn't add the `enqueue_at` attribute. It feels to us that it could be a simple change to make fake mode closer to the real one.

We're happy to implement this change if approved.

mperham:
:+1:

PR #3258, opened 2016-11-25
Add the enqueued_at attribute when fake Testing

Ferdy89:
Implements https://github.com/mperham/sidekiq/issues/3245

Trying to emulate the real behavior, the job obtains an `enqueued_at` attribute when it's placed in a queue.

Since using `perform_in` or `perform_at` does not put the job in the queue when run, the attribute is not set for those methods.

### DIFF
--- a/lib/sidekiq/testing.rb
+++ b/lib/sidekiq/testing.rb
@@ -69,7 +69,9 @@ module Sidekiq
     def raw_push(payloads)
       if Sidekiq::Testing.fake?
         payloads.each do |job|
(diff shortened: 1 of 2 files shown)


sergey-alekseev:
In tests I get `Redis::CannotConnectError` with `perform_in` while having the default testing strategy (just `require 'sidekiq/testing'`, so `Sidekiq::Testing.fake!`) and Redis server turned off. Doesn't happen with `perform_async`. Sidekiq v5.2.5.

PR #3277, opened 2016-12-13
dir should from options[:require], rails.root maybe only for developm…

houdl:
dir should from options[:require], rails.root maybe only for development use

### DIFF
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -253,7 +253,7 @@ module Sidekiq
     end
 
     def default_tag
-      dir = ::Rails.root
+      dir = File.expand_path(options[:require])
       name = File.basename(dir)
       if name.to_i != 0 && prevdir = File.dirname(dir) # Capistrano release directory?
         if File.basename(prevdir) == 'releases'


mperham:
Why?

houdl:
@mperham   when i only use grape framework, there is no anything about rails.

mperham:
`default_tag` is only called if Rails is detected.  You can set your tag with `-g xyz`.

houdl:
ok, thanks

houdl:
@mperham   i think that it should have a default tag even if it is not used in rails framwork.

mperham:
How?  What should it be?

houdl:
@mperham  bacause we always not set tag for sidekiq a lot of time.  in capistrano-sidekiq , the tag also is nil , so when we not use rails, there alway have a error message, something like 
```ruby
no implicit conversion of nil into String
sidekiq/cli.rb:257:in `basename'
```
i think sidekiq should have a default tag not only for rails.

mperham:
The only time it calls `default_tag` is when options[:require] is a directory, which is what happens when you don't give Sidekiq a boot script for your application, e.g. `-r ./boot.rb`.  Sidekiq assumes Rails in that case because that is what 90% of people are using.  Use `-r` so Sidekiq knows how to boot your non-Rails app.

houdl:
@mperham  thanks for your answer patiently. i have know what your mean. options[:require] directory only for rails, other framwork must appoint enviroment file.

PR #3293, opened 2016-12-25
Fixnum -> Integer replacement and Ruby 2.4 support for Travis.

marshall-lee:
Fixnum is deprecated in Ruby 2.4 so we should get rid of it.


marshall-lee:
Hmm, build fails only on `jruby-head` and it seems to be unrelated.

mperham:
thanks!

PR #3313, opened 2017-01-11
Mention the error handler context change in 4.2.3

aprescott:
I believe a custom handler added to `error_handlers` is supported functionality. The change in #3161 unknowingly broke some of our exception handling on retries, because we were expecting `context` to be `job`, whereas we now need to use `context[:job]` before pulling out job args.

```ruby
# before
context["retry_count"]

# after
context[:job]["retry_count"]
```


mperham:
Thank you and sorry for the breakage!  I shouldn't have put that in a patch release and didn't think it through.

aprescott:
It happens! No real damage done. Thanks for your ongoing awesomeness working on Sidekiq.

PR #3348, opened 2017-02-10
Show error class on retry page

stepantubanov:
Shows error class in corresponding table cell. Job's display class is still on the page, it is shown at the top of the page (in job details).


mperham:
Wow, how did we miss that all this time???  Thanks!

PR #3370, opened 2017-03-04
Auto-size redis pool on clients with MAX_THREADS

nateberkopec:
Thoughts Mike? I can add a test if you think it's a good idea.

### DIFF
--- a/lib/sidekiq/redis_connection.rb
+++ b/lib/sidekiq/redis_connection.rb
@@ -12,7 +12,7 @@ module Sidekiq
 
         options[:url] ||= determine_redis_provider
 
-        size = options[:size] || (Sidekiq.server? ? (Sidekiq.options[:concurrency] + 5) : 5)
+        size = options[:size] || default_pool_size
 
         verify_sizing(size, Sidekiq.options[:concurrency]) if Sidekiq.server?
 
@@ -101,6 +101,13 @@ module Sidekiq
         ENV[ENV['REDIS_PROVIDER'] || 'REDIS_URL']
       end
 
+      def default_pool_size
+        if Sidekiq.server?
+          Sidekiq.options[:concurrency] + 5
+        elsif ENV['MAX_THREADS'] || ENV['RAILS_MAX_THREADS']
+          Integer(ENV['MAX_THREADS'] || ENV['RAILS_MAX_THREADS']) || 5
+        end
+      end
     end
   end
 end


mperham:
Why not just use the Rails max threads variable?

nateberkopec:
tbh I don't follow Rails development closely enough. @schneems are we all supposed to use RAILS_MAX_THREADS now? All I know is that "in the wild" most people still use MAX_THREADS, though I realize that may be outdated/deprecated.

mperham:
Yes, RAILS_MAX_THREADS is supposed to hint to the process the amount of parallelism.  puma and sidekiq should both respect the variables.  Rails will size the DB connection pool based on it.  Conventions are the best, use it, love it.

nateberkopec:
@mperham updated to go with RAILS_MAX_THREADS, not sure why you closed.

nateberkopec:
Does GH not update closed PRs? can open another.

nateberkopec:
updated patch would be https://github.com/nateberkopec/sidekiq/commit/f9a7bfb7b1fcfa7923d174315ecc942b27a5a5b3

lib/sidekiq/redis_connection.rb
```diff
@@ -101,6 +101,13 @@ def determine_redis_provider
         ENV[ENV['REDIS_PROVIDER'] || 'REDIS_URL']
       end
 
+      def default_pool_size
+        if Sidekiq.server?
+          Sidekiq.options[:concurrency] + 5
+        elsif ENV['MAX_THREADS'] || ENV['RAILS_MAX_THREADS']
```
mperham: You're missing an else clause.

mperham:
I can't reopen due to a force push?

mperham:
Sidekiq already adjusts the concurrency based on RAILS_MAX_THREADS and then sets the pooi size based on the concurrency so I don't think this PR is necessary, but correct me if I misunderstand.

nateberkopec:
@mperham I want to make sure client pool sizes aren't unneccessarily large on, say, single-threaded apps (RAILS_MAX_THREADS=1). AFAICT in that case right now you'd set the pool size to 5, even though the client will only ever use 1. Server-side concurrency and pool sizes are not affected (they look good right now).

mperham:
Rails's connection pool and the connection_pool gem are both lazy.  If there's only one thread, only one connection will be created.  But if your one Unicorn thread happens to spawn off other threads (for instance, with sucker_punch), then the gem will allow up to 5 connections by default.  Seems like a good tradeoff.

nateberkopec:
Works for me! Forgot about the sucker punch use case too.

schneems:
`RAILS_MAX_THREADS` is the best guess for thread count unless they configured manually.

👍

PR #3379, opened 2017-03-10
Add farsi language to locals file

mm580486:


mperham:
Very cool, first new language in a year or so!

web/locales/fa.yml
```diff
@@ -0,0 +1,79 @@
+# elements like %{queue} are variables and should not be translated
+en: # <---- change this to your locale code
```
squarism: Should this be `fa`?

seuros:
The issue is that farsi is RTL language.

seuros:
@mm580486 can you should a screenshoot of the dashboard ?

squarism:
RTL = right to left for anyone wondering  🌻

Issue #3402, opened 2017-03-22
4.2.10 is broken for us because of #3384


eigilsagafos:
@mperham Issue #3384 breaks Sidekiq for us. We are not using Rails, but we are using ActionView (5.0.2) which depends on rails-html-sanitizer. Rails HTML sanitizer lives within the Rails namespace (Rails::Html::Sanitizer) so `!Rails.env` crashes with a no method error.

PR #3403, opened 2017-03-22
Ensure Rails module has #env

eigilsagafos:
This addresses #3402

### DIFF
--- a/lib/sidekiq/testing.rb
+++ b/lib/sidekiq/testing.rb
@@ -317,7 +317,7 @@ module Sidekiq
   end
 end
 
-if defined?(::Rails) && !Rails.env.test?
+if defined?(::Rails) && Rails.respond_to?(:env) && !Rails.env.test?
   puts("**************************************************")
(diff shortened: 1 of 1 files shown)


lib/sidekiq/testing.rb
```diff
@@ -317,7 +317,7 @@ def drain_all
   end
 end
 
-if defined?(::Rails) && !Rails.env.test?
+if defined?(::Rails) && Rails.respond_to?(:env) && !Rails.env.test?
```
seuros: if defined?(::Rails::VERSION)

eprothro:
@mperham personally, I'd love to see a release for this fixup, I would wager this breaks more than eigilsagafos and our gems dependent on sidekiq but not rails.

mperham:
I don't like to issue quick patch releases unless a large percentage of users are affected.  You can put this in your initializer to work around:

```ruby
# Remove once running Sidekiq 4.2.11+
module Rails
  def self.env
    self
  end
  def self.test?
    true
  end
end
```

eprothro:
Yep, already have. Thanks, figured I would check.

PR #3418, opened 2017-04-04
Modify `drain_all` to respect enqueued order

ggilder:
Currently `Sidekiq::Worker.drain_all` effectively reorders the queue by
grouping job classes together. This change maintains the enqueued order
of jobs in the queue when using `drain_all`.

### DIFF
--- a/lib/sidekiq/testing.rb
+++ b/lib/sidekiq/testing.rb
@@ -306,11 +306,9 @@ module Sidekiq
       # Drain all queued jobs across all workers
       def drain_all
         while jobs.any?
-          worker_classes = jobs.map { |job| job["class"] }.uniq
-
-          worker_classes.each do |worker_class|
-            worker_class.constantize.drain
-          end
+          next_job = jobs.first
+          Queues.delete_for(next_job["jid"], next_job["queue"], next_job["class"])
+          next_job["class"].constantize.process_job(next_job)
         end
       end
     end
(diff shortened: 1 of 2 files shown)


mperham:
It is a smell to depend on execution ordering in an async system.

ggilder:
I agree in principle, but you already have deterministic ordering here — it's just not the natural ordering that jobs were enqueued in.

mperham:
Good point, we should make it random. :-D

ggilder:
Haha... ok, so maybe to make this more concrete I can present the problem that led me to stumble on this.

I have an app where I use Cucumber scenarios to test complex flows. They make fairly heavy use of tables to concisely document the state of the database before and after certain actions are performed, so something like this simplified example:

```
Then expect the incidents table to contain:
  | id  | source | case_id | state       | amount |
  | 801 | abc01  | 10001   | active_case | 4000   |
  | 802 | abc01  | 10002   | active_case | 1600   |
  | 803 | abc02  | 10003   | active_case | 1600   |
  | 804 | abc02  | 10004   | active_case | 2300   |
```

So there's an implicit assumption that associated models (in this example, cases and incidents) are created in order because we enqueue jobs that process each incident in order.

Now, you could certainly rewrite that to not specify the foreign key ids, but it would be much more cumbersome. So from my perspective, maintaining the queue order is a convenience that allows my test scenarios to be simpler and easier to maintain.

If this is not something you want to support in sidekiq, that's fine... just thought I'd give some context in case that sways you at all. :)

mperham:
You can assert the ID is non-nil, you can assert it is within a range but I still maintain it's a bad idea to depend on precise code ordering within your integration tests unless you explicitly execute the code yourself.

Issue #3437, opened 2017-04-18
Web: Localization in wrong language in Chrome & Safari (Mac)


johanlunds:
**Ruby version:** 2.4
**Sidekiq / Pro / Enterprise version(s):** sidekiq (4.2.10), sidekiq-pro (3.4.5)

**Chrome version:** 57.0.2987.133 (stable)
**Safari version:** 10.0.3
**OS version:** 10.12.3 (macOS Sierra)

~~Please include your initializer and any error message with the full backtrace.~~

**Are you using an old version?** No
**Have you checked the changelog to see if your issue has been fixed in a later version?** Yes

## Description

**Reproduce:** Set OS localization settings like mine (see screenshot below). I have it set as English primary and Swedish secondary. All other apps in the OS will be localized in English with these settings. Open Chrome. Go then to the Sidekiq Web GUI. Do the same in Safari.

**Expected behavior:** In **both Chrome & Safari** Sidekiq Web GUI should be localized in **English.**

**Actual behavior:**  In **Chrome** Sidekiq Web GUI is localized in **Swedish**. In **Safari** it works as expected - it is in **English**.

**Investigation:**

Header sent in Chrome request is `Accept-Language: en-US,sv-SE;q=0.8,sv;q=0.6,en;q=0.4`.

Safari doesn't seem to send a `Accept-Language` header (?). Using the Ruby debugger I see that `env['HTTP_ACCEPT_LANGUAGE']` is `en-us` then.

**Screenshot of OS settings:** 

![language___region_and_new_issue_ _mperham_sidekiq](https://cloud.githubusercontent.com/assets/22144/25136281/e0eb9d8e-2454-11e7-8d82-f489f26b04bd.png)

johanlunds:
Further testing: switching the OS locale settings to be (1) Swedish preferred, (2) English secondary (and restarting Safari so it is now in Swedish), the Sidekiq Web GUI is still in English.

Expected behavior would be that it is in Swedish (since that is the primary, preferred language set in the OS settings).

Inspecting the request in Safari:

![_development__sidekiq_pro_and_web__localization_in_wrong_language_in_chrome__mac__ _issue__3437_ _mperham_sidekiq](https://cloud.githubusercontent.com/assets/22144/25137595/8960ab46-2458-11e7-8398-8e82399ca298.png)

[...]

PR #3448, opened 2017-04-30
Use Rack::Utils to parse HTTP_ACCEPT_LANGUAGE

natematykiewicz:
In https://github.com/mperham/sidekiq/issues/3437 @mperham asked for the locale parser to use Rack::Utils instead of the one he wrote. The goal was to fix the locale issue that @johanlunds was having. I added a test case for the header his browser was sending. Unfortunately, the current code was already parsing it to "sv", so this shouldn't actually fix the issues he was having. It does, however, use Rack::Utils, which might cover other problems that we don't yet know of.

### DIFF
--- a/lib/sidekiq/web/helpers.rb
+++ b/lib/sidekiq/web/helpers.rb
@@ -85,7 +85,12 @@ module Sidekiq
       @locale ||= begin
         locale = 'en'.freeze
         languages = env['HTTP_ACCEPT_LANGUAGE'.freeze] || 'en'.freeze
-        ::Rack::Utils.best_q_match(languages.downcase, available_locales) || locale
+
+        # Put our default locale in front, since '*' is going to match
+        # the first locale in our list.
+        locales = (available_locales - [locale]).unshift(locale)
+
+        ::Rack::Utils.best_q_match(languages.downcase, locales) || locale
       end
     end
 
--- a/lib/sidekiq/web/helpers.rb
+++ b/lib/sidekiq/web/helpers.rb
@@ -81,9 +81,8 @@ module Sidekiq
       @locale ||= begin
         locale = 'en'.freeze
         languages = env['HTTP_ACCEPT_LANGUAGE'.freeze] || 'en'.freeze
-        languages.downcase.split(','.freeze).each do |lang|
+        ::Rack::Utils.q_values(languages.downcase).map(&:first).each do |lang|
           next if lang == '*'.freeze
-          lang = lang.split(';'.freeze)[0]
           break locale = lang if find_locale_files(lang).any?
         end
         locale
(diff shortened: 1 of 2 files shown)


natematykiewicz:
In reading more about these q values, I realize that this PR is not correct. In fact, it would make the most sense that "en-US,sv-SE;q=0.8,sv;q=0.6,en;q=0.4" would be parsed to "en", since en is 1 (the default).

I'll update the PR, now.

natematykiewicz:
I see. So, the locales in order of q are "en-us", "sv-se", "sv", "en". "sv" is the first locale that actually has a definition, so that's the one that's currently being chosen. The current implementation feels fragile, which is why a lot of these libraries seem to sort on the q value, instead of assuming that they're already sorted.

https://github.com/rack/rack-contrib/blob/master/lib/rack/contrib/locale.rb#L26

Here's an interesting question. If English is primary, and Swedish is secondary, why does "sv" have a higher q value than "en"?

@mperham The i18n gem has fallbacks you can define for locales. If you were to define a fallback for "en-us" that was simply "en", then in this case, "en" would be the chosen locale instead of "sv".

What are your thoughts on that?

natematykiewicz:
At the moment, it's failing on this test:
(expecting 'ru,us' to be Russian, but getting English text back)
https://github.com/mperham/sidekiq/blob/master/test/test_web.rb#L61

If you look at http://guides.rubyonrails.org/i18n.html#choosing-an-implied-locale, Rails mentions two gems to help you determine the locale.

https://github.com/rack/rack-contrib/blob/master/lib/rack/contrib/locale.rb#L26
and
https://github.com/iain/http_accept_language/blob/master/lib/http_accept_language/parser.rb#L17

Interestingly, they both return different results when given 'ru,us'. Rack Contrib says 'us', where http accept language says 'ru'.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.4
The RFC doesn't mention having two different locales without a "q". Though, my gut says that "ru" makes sense to use if it were to occur. I do question if it actually happens, or if "ru,us;q=0.8" would actually be more realistic. Nonetheless, this would be a breaking change, so it's probably not acceptable as-is.

johanlunds:
Awesome Nate! Good research.

`Rack::Utils.best_q_match` does not seem to be a solution.

`Rack::Contrib` seems to simplistic. I don't think it would handle a request like `sv-SE,sv`.

`HttpAcceptLanguage` seems most interesting, especially [`#language_region_compatible_from`](https://github.com/iain/http_accept_language/blob/master/lib/http_accept_language/parser.rb#L94-L106).

The RFC was a bit difficult to understand.

johanlunds:
Hello @natematykiewicz! I continued your work in https://github.com/mperham/sidekiq/pull/3449. I hope that was ok!

PR #3449, opened 2017-04-30
Improve parsing of HTTP_ACCEPT_LANGUAGE (continue Nate's work)

johanlunds:
See: https://github.com/mperham/sidekiq/pull/3448 and https://github.com/mperham/sidekiq/issues/3437

cc @natematykiewicz @mperham 

I continued on @natematykiewicz's work. This PR passes all tests. It borrows some code from https://github.com/iain/http_accept_language/blob/master/lib/http_accept_language/parser.rb.

### DIFF
--- a/lib/sidekiq/web/helpers.rb
+++ b/lib/sidekiq/web/helpers.rb
@@ -85,7 +85,12 @@ module Sidekiq
       @locale ||= begin
         locale = 'en'.freeze
         languages = env['HTTP_ACCEPT_LANGUAGE'.freeze] || 'en'.freeze
-        ::Rack::Utils.best_q_match(languages.downcase, available_locales) || locale
+
+        # Put our default locale in front, since '*' is going to match
+        # the first locale in our list.
+        locales = (available_locales - [locale]).unshift(locale)
+
+        ::Rack::Utils.best_q_match(languages.downcase, locales) || locale
       end
     end
 
--- a/lib/sidekiq/web/helpers.rb
+++ b/lib/sidekiq/web/helpers.rb
@@ -102,7 +102,7 @@ module Sidekiq
             preferred_language == available.split('-', 2).first
           end
 
-          lang_group.find { |lang| lang == preferred } || lang_group.first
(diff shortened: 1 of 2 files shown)


mperham:
@johanlunds Does this solve your original issue?

@natematykiewicz Look good?

natematykiewicz:
@johanlunds, Looks awesome to me! I'm definitely not offended. Glad I could help!

I still don't understand how your original header that your browser was sending makes any sense, but I guess that's Chrome's issue.

If you said "English primary, Swedish secondary" in your settings, why would your header be:
`en-US,sv-SE;q=0.8,sv;q=0.6,en;q=0.4`
english, swedish, swedish, english.

Seems like it should be `en-US,en;q=0.8,sv-SE;q=0.6,sv;q=0.4` to me (specific english, general english, specific swedish, general swedish).

@mperham, should he freeze the 'en' on line 110 of lib/sidekiq/web/helpers.rb? Also, what do you want me to do with my PR? Should I hit the "Close Pull Request" button?

mperham:
No need to freeze, I generally freeze strings only in the hot codepath.  This isn't a high traffic codepath.

johanlunds:
> @johanlunds Does this solve your original issue?

Yes.

Thanks for merging this!

> @johanlunds, Looks awesome to me! I'm definitely not offended.

Good!

> I still don't understand how your original header that your browser was sending makes any sense, but I guess that's Chrome's issue.

I have more info on that actually. Turns out, things do not work the way I thought.

Neither Chrome nor Safari sends languages other than the Primary ([as noted in the original issue description](https://github.com/mperham/sidekiq/issues/3437#issue-222444923)) in the Accept-Language header. The Accept-Language header was instead constructed from Primary Language + Region. I have English + Swedish, which one can argue is a weird unusual combination.

Either way, with this PR everyone will get their desired behavior I think - even me with my weird OS locale settings.

johanlunds:
I found this now: https://www.w3.org/International/questions/qa-lang-priorities seems to have good documentation on the header and up-to-date info on how specific browsers behave. Good to have if this issue would arise again in the future.

PR #3453, opened 2017-05-02
Fix race condition in Sidekiq API

### LINKED ISSUE #3452: [Web UI] TypeError GET /busy
Ruby version: 2.2.0
Sidekiq / Pro / Enterprise version(s): Sidekiq 4.2.10 / Pro 3.5.0

I've encountered a bug with the Sidekiq Web UI that I'm unfortunately unable to reproduce. I was testing out the "Stop" button on the Busy tab while the browser had polling turned on in the Busy tab and got an error. As far as I can tell from looking at the stack trace, here's what happened:

1. The processes were retrieved from Redis: https://github.com/mperham/sidekiq/blob/13f746a7a72c257ff4448de71f13d24182efff66/lib/sidekiq/api.rb#L696
2. One of the processes was stopped via the Web UI.
3. Data for each process was pulled from Redis: https://github.com/mperham/sidekiq/blob/13f746a7a72c257ff4448de71f13d24182efff66/lib/sidekiq/api.rb#L702
4. The stopped process has no data in Redis, returning `nil` and causing the JSON parse to error out: https://github.com/mperham/sidekiq/blob/13f746a7a72c257ff4448de71f13d24182efff66/lib/sidekiq/api.rb#L709

I'll whip up a quick pull request to fix.

JacobEvelyn:
Fixes #3452.

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -725,6 +725,11 @@ module Sidekiq
         end
 
         result.each do |info, busy, at_s, quiet|
+          # If a process is stopped between when we query Redis for `procs` and
+          # when we query for `result`, we will have an item in `result` that is
+          # composed of `nil` values.
+          next if info.nil?
+
           hash = Sidekiq.load_json(info)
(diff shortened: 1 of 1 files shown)


mperham:
thank you!

PR #3457, opened 2017-05-04
Update footer UTC time in web dashboard polling refresh

fidelisrafael:
Hi there! First of all, thank you guys for the amazing work on Sidekiq, I really appreciate it.

I'm sending this small PR with a small change in Sidekiq Web Dashboard:

### Why this?

When I running my jobs, I keep an eye in dashboard to follow how it's going...and for debugging
purposes sometimes I want to know how much time a specific batch of jobs takes to run(from enquedued to processed) and
how many seconds it take...and I dont want to count the seconds by myself...So I thought updating the current time in footer would be a good ideia...even if I don't need to count the time...it's just a good thing so this reflect the current stats for current time.
I mean, this is a way to say "this stats are related to this current time"(if you let Sidekiq Dashboard opened for 1 hour for example, the date in dashboard will be one hour late...weird)

### DIFF
--- a/lib/sidekiq/web/application.rb
+++ b/lib/sidekiq/web/application.rb
@@ -234,7 +234,6 @@ module Sidekiq
     get '/stats' do
       sidekiq_stats = Sidekiq::Stats.new
       redis_stats   = redis_info.select { |k, v| REDIS_KEYS.include? k }
-
       json(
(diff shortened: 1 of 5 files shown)


mperham:
thank you!

Issue #1786, opened 2014-06-17
Show arguments passed to the worker on the logs


ezilocchi:
Currently the logs are being displayed like this:

```
2014-06-17T21:17:02Z 16566 TID-ospmsznzk MyWorker JID-474923f4986cb3dc0a00a007 INFO: start
2014-06-17T21:17:03Z 16566 TID-ospmsznzk MyWorker JID-474923f4986cb3dc0a00a007 INFO: done: 1.491 sec
```

I would like to show which are the arguments that have been passed to the worker on that execution. So it will be much easier to debug on production
Example:

```
2014-06-17T21:17:02Z 16566 TID-ospmsznzk MyWorker with { first: 'some value', second: 'another value' } JID-474923f4986cb3dc0a00a007 INFO: start
2014-06-17T21:17:03Z 16566 TID-ospmsznzk MyWorker JID-474923f4986cb3dc0a00a007 INFO: done: 1.491 sec
```

Based on what I understood from the wiki (https://github.com/mperham/sidekiq/wiki/Logging) the only way of modifying the logging message is replacing the Formatter: 

```
Sidekiq.logger.formatter = MyFormatter.new
```

But It doesn't receives the actual arguments, only `(severity, time, program_name, message)`:
`lib/sidekiq/logging.rb`

``` ruby
    class Pretty < Logger::Formatter
      # Provide a call() method that returns the formatted message.
      def call(severity, time, program_name, message)
        "#{time.utc.iso8601} #{Process.pid} TID-#{Thread.current.object_id.to_s(36)}#{context} #{severity}: #{message}\n"
      end

      def context
        c = Thread.current[:sidekiq_context]
        c ? " #{c}" : ''
      end
    end
```

Is there any way of doing this? I found that probably some work can be done on this other class `lib/sidekiq/middleware/server/logging.rb` where `item` contains exactly what I need, but it's not being passed to `Sidekiq::Logging` where I can override the `Sidekiq.logger.formatter`:

``` ruby
        def call(worker, item, queue)
          Sidekiq::Logging.with_context("#{worker.class.to_s} JID-#{item['jid']}") do
            begin
              start = Time.now
              logger.info { "start" }
              yield
              logger.info { "done: #{elapsed(start)} sec" }
            rescue Exception
              logger.info { "fail: #{elapsed(start)} sec" }
              raise
            end
          end
        end
```

Any suggestion?
Thanks

khoan:
[...]

PR #3463, opened 2017-05-08
Extract Processor#log_context to ease extending what's logged

Empact:
This was previously exposed in JobLogger and Middleware::Server::Logging
but was inlined into Processor.
https://github.com/mperham/sidekiq/commit/701e06224cebfeeb1f9c84e3abd9b96263defa46

Useful, e.g. in this case:
https://github.com/mperham/sidekiq/issues/1786

### DIFF
--- a/lib/sidekiq/processor.rb
+++ b/lib/sidekiq/processor.rb
@@ -119,21 +119,13 @@ module Sidekiq
       nil
     end
 
-    def log_context(job_hash)
-      # If we're using a wrapper class, like ActiveJob, use the "wrapped"
-      # attribute to expose the underlying thing.
-      klass = job_hash['wrapped'.freeze] || job_hash["class".freeze]
-      bid = job_hash['bid'.freeze]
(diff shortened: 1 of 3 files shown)


lib/sidekiq/processor.rb
```diff
@@ -119,18 +119,21 @@ def handle_fetch_exception(ex)
       nil
     end
 
+    def log_context(job_hash)
```
mperham: While we're doing this refactoring, it seems like this method is a logging concern that should go into Sidekiq::Logging.

Empact: I'm ambivalent on that point because that would mean teaching Logging about job_hashes rather than having a simple string-based interface. Maybe you have other ideas on implementation?

mperham: 👍

mperham:
Build failure.  Will merge once fixed.

Empact:
Fixed + added a changelog entry.

PR #3474, opened 2017-05-13
Remove core extensions

mperham:
Right now Sidekiq optionally loads a few ActiveSupport patches or monkeypatches a few Ruby core classes with its own helpers in sidekiq/core_ext.rb.  This shouldn't necessary.  Pull out the few necessary methods so we don't touch the core classes.

- [x] class_attribute
- [x] Hash#*_keys
- [x] Hash#deep_merge*
- [x] Kernel#silence_warnings
- [x] String#constantize


lib/sidekiq.rb
```diff
@@ -149,7 +149,11 @@ def self.default_server_middleware
   end
 
   def self.default_worker_options=(hash)
-    @default_worker_options = default_worker_options.merge(hash.stringify_keys)
+    # stringify
+    hash.keys.each do |key|
+      hash[key.to_s] = hash.delete(key)
+    end
+    @default_worker_options = default_worker_options.merge(hash)
```
solnic: I'd recommend **not mutating** `hash` here, just create a copy instead :)

mikegee: How about starting with a copy of the defaults, then updating it with the provided hash like this?

```ruby
@default_worker_options = default_worker_options.dup
hash.each do |key, value|
  @default_worker_options[key.to_s] = value
end
```

lib/sidekiq.rb
```diff
@@ -223,6 +227,16 @@ def self.on(event, &block)
...
 
+  def self.constantize(str)
+    names = str.split('::')
+    names.shift if names.empty? || names.first.empty?
+
+    constant = Object
+    names.each do |name|
+      constant = constant.const_defined?(name) ? constant.const_get(name) : constant.const_missing(name)
+    end
+    constant
```
mikegee: This way is more terse, but perhaps less obvious:

```ruby
def self.constantize(str)
  names = str.split('::')
  names.shift if names.empty? || names.first.empty?
  
  names.inject(Object) do |constant, name|
    constant.const_defined?(name) ? constant.const_get(name) : constant.const_missing(name)
  end
end
```

mperham:
This is complete if anyone wants to give it a final review.

lib/sidekiq.rb
```diff
@@ -149,7 +149,12 @@ def self.default_server_middleware
   end
 
   def self.default_worker_options=(hash)
-    @default_worker_options = default_worker_options.merge(hash.stringify_keys)
+    h = hash.dup
+    # stringify
+    h.keys.each do |key|
+      h[key.to_s] = h.delete(key)
+    end
+    @default_worker_options = default_worker_options.merge(h)
```
rosenfeld: @mperham I usually do this with something like:

```ruby
hash_with_string_keys = Hash[ hash.map{|k, v| [k.to_s, v] } ]
```ruby

**Comment by ryansch** [comment]
I see that you've duplicated your stringify code in a number of places.  Why not stick that code in a method somewhere and pass the hash in?  The method need not be a monkey patch.

File: lib/sidekiq.rb (hunk 3 above)
Comment (mperham) [inline lib/sidekiq.rb]: 🤷‍♂️ I find that less readable and allocates N Arrays + 1 Hash instead of 2 Hashes.

**Comment by mperham** [comment]
@ryansch There's no good place to put it and I didn't want to create a junk drawer for one method.  Sidekiq::Util is server-side only.

File: lib/sidekiq.rb (hunk 3 above)
Comment (rosenfeld) [inline lib/sidekiq.rb]: That's fine (I find it more readable but if you find it less readable this is a good enough argument to me). I just think that this kind of micro-optimization doesn't make much sense in this case. How many possible keys we're talking about after all? ;) It will happen just once :)

Comment (mperham) [inline lib/sidekiq.rb]: I agree, I'm 50/50 on it.  I like that yours is a one-liner and can be inlined.  It's not a hot path so I'm still considering it...

Comment (rosenfeld) [inline lib/sidekiq.rb]: If you're still worried about performance, I tried both approaches in IRB with 100 symbol keys in the array. The both seem to take something like a bit less than 10us... Just give it a try if you don't believe ;)

Comment (rosenfeld) [inline lib/sidekiq.rb]: I don't know why some people seem to have a hard time trying to understand why the existence of symbols is what I consider Ruby's worst part ;)

File: lib/sidekiq/processor.rb [hunk 4]
```diff
@@ -234,5 +234,14 @@ def cloned(thing)
       Marshal.load(Marshal.dump(thing))
     end
 
+    def constantize(str)
+      names = str.split('::')
+      names.shift if names.empty? || names.first.empty?
```
rosenfeld: May I ask you why you're shifting if `names.empty?`?

lib/sidekiq/testing.rb
```diff
@@ -55,6 +55,15 @@ def server_middleware
         yield @server_chain if block_given?
         @server_chain
       end
+
+      def constantize(str)
+        names = str.split('::')
+        names.shift if names.empty? || names.first.empty?
```
rosenfeld: @mperham in case you miss, the same happens here.

lib/sidekiq/processor.rb
mperham: This is a short circuit so the base case `"".split("::")` still works.

rosenfeld: But the first condition is doing nothing in this case, just the second one. Actually, for this specific case of an empty string before split, not even the second one, but I understand the second one is important for `'Class'.split("::")`.

rosenfeld: Actually I think none of them are actually useful, right? Maybe we should just delete that line...

rosenfeld: Or maybe the split behavior might have changed among Ruby versions. Is that possible. I didn't check in all Ruby supported versions...

rosenfeld: Oh, the second is required for `"::Class".split('::")`

lib/sidekiq/worker.rb
```diff
@@ -135,8 +140,68 @@ def get_sidekiq_options # :nodoc:
...
+        # stringify
+        item.keys.each do |key|
+          item[key.to_s] = item.delete(key)
+        end
+
+        Sidekiq::Client.new(pool).push(item)
+      end
+
+      def class_attribute(*attrs)
+        instance_reader = true
```
rosenfeld: I'm curious, why does the worker need `class_attribute`?

mperham: So you can have a BaseWorker with a default set of properties and BaseWorker subclasses that tune those properties.  It's quite common with Rails developers to do this.

rosenfeld: Oh, I see it's used to extend `base` in the `included` hook.

rosenfeld: Maybe that API could be simplified by using something like `set_property 'prop', value` and `get_property 'prop'` but at least I do understand now why it exists :)

lib/sidekiq/processor.rb
rosenfeld: Oh, I finally got it. The first condition is to avoid calling `nil.empty?`. I'd prefer it to be more explicit though:

```ruby
names.shift if !names.empty? && names.first.empty?
```

Or

```ruby
names.shift if names.first&.empty?
```

rosenfeld: @mperham, did you miss these comments?

mperham: Nope, I think it's fine as-is.

rosenfeld: Ok, thanks for merging this, looking forward to the new monkey patch free release 😜

mperham: Thanks for the comments, I really appreciate the feedback.

PR #3536, opened 2017-06-30
do not pass AR object

hooopo:
why?

look the wiki:

https://github.com/mperham/sidekiq/wiki/Best-Practices#1-make-your-job-parameters-small-and-simple


mperham:
Sidekiq never touches the database.  If you want to do this, use ActiveJob.

PR #3540, opened 2017-07-05
Update changelog for section 5.0.4

devality:
I spent some time finding out about issue. I think it would be better if changelog has it.


mperham:
Thank you!

PR #3548, opened 2017-07-12
Bump default timeout

iloveitaly:
Thanks for all the incredibly hard work you've put into Sidekiq!

[Heroku has extended the time](https://devcenter.heroku.com/articles/dynos#shutdown) workers have to terminate when a `TERM` is issued to 30 seconds. I ran into this issue on resque ([more info here](https://github.com/iloveitaly/resque-heroku-signals)).

I think it would be good to update the defaults in sidekiq to match Heroku's new worker timeout. What do you think?

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -6,7 +6,7 @@
 :verbose: false
 :concurrency: 10
 
-# Set timeout to 8 on Heroku, longer if you manage your own systems.
+# Set timeout to 28 on Heroku, longer if you manage your own systems.
 :timeout: 30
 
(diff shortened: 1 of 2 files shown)


mperham:
Lots of deploy scripts assume Sidekiq's default. I bet this change would suddenly cause a lot of kill -9's. WDYT?

iloveitaly:
@mperham Ah, that's a great point. Probably best not to modify the defaults and just update code comments and the wiki?

mperham:
Send me a PR?

PR #3573, opened 2017-08-11
Unified Job kill via Sidekiq::DeadSet#kill

Yurokle:
Currently there are three different places which contain exactly the same piece of code that kills a job and cleans the morgue:
```
now = Time.now.to_f		
Sidekiq.redis do |conn|		
  conn.multi do		
    conn.zadd('dead', now, payload)		
    conn.zremrangebyscore('dead', '-inf', now - DeadSet.timeout)		
    conn.zremrangebyrank('dead', 0, - DeadSet.max_jobs)		
  end		
end
```
Logic and magic words (`'dead'` for sorted set name) are spread around the codebase. Since all that code already depends on `DeadSet` class (`DeadSet.timeout` and `DeadSet.max_jobs`), we could just move the logic to `DeadSet#kill(job_payload)`.

\* I really wanted to also re-use `DeadSet#schedule` in `DeadSet#kill`, but it does `Sidekiq.dump_json(payload)` under the hood while we are already getting serialized data for all job kill calls :(

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -458,14 +458,7 @@ module Sidekiq
     # Place job in the dead set
     def kill
       remove_job do |message|
-        now = Time.now.to_f
-        Sidekiq.redis do |conn|
-          conn.multi do
-            conn.zadd('dead', now, message)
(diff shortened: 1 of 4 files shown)


mperham:
thank you!

PR #3574, opened 2017-08-11
When a middleware mutates arguments, retry with original args

magnusvk:
Currently when a server middleware mutates arguments and then the job
fails, it will get re-queued with the mutated arguments, not the
original arguments.

This is unexpected and (IMO) faulty behavior as that same middleware
will now see the already-mutated arguments when the job gets executed
again.

This changes the behavior to use the pristine, unmutated jobs hash when
requeueing the failed job. It also includes a failing test case for the
old code.

### DIFF
--- a/lib/sidekiq/processor.rb
+++ b/lib/sidekiq/processor.rb
@@ -126,7 +126,7 @@ module Sidekiq
       pristine = cloned(job_hash)
 
       Sidekiq::Logging.with_job_hash_context(job_hash) do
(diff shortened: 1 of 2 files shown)


mperham:
a bug in the 5.0 refactoring, nice catch. thanks!

magnusvk:
Going to pull this fix into our app via Github directly, but would love a new gem version released if you have the chance. :) Thanks.

PR #3575, opened 2017-08-11
Add docs for the class argument of Sidekiq::Client.push

malept:
Document the difference between using a class name string vs. a `Sidekiq::Worker`-based worker class. It wasn't obvious what the difference was without digging into the code.

I'm happy to change the wording, this was just a way to get the conversation started.


mperham:
thanks!

PR #3592, opened 2017-08-31
Make heartbeat TTL configurable

lygaret:
**What**
Make the worker's heartbeat TTL configurable, rather than hard-coded at 60 seconds

**Why**
*Note: we're evaluating pro, using `super_fetch!`*

When a worker process crashes the Ruby VM, `super_fetch!` helps us ensure that dead jobs are still in redis and can be requeued with `SuperFetch::Retriever.cleanup_the_dead` which is called as part of process startup. We think this'll work pretty well for us, but has the downside that we have to wait 60 seconds before restarting the process, because "the dead" in this case means the heartbeat has expired.

In our particular use-case, the one minute heartbeat expiration makes for a poor user experience: we have "real-time" jobs, and clients which are prone to worry if processing latency goes way up.

We'd like to be able to tune the heartbeat TTL to balance user-experience in this case with the possibility of network events etc. Obviously the TTL can't go too low (causing partitions, etc.) but we've got a pretty small and stable network and relatively few worker/threads at the moment, so we feel pretty confident we can bring the TTL down at least half if not more.

**Misc**

* this was a small enough set of changes that I figured I'd just open the PR, rather than go through an issue, apologies if this was incorrect.
* this PR is against 4.x which is what we needed for our experiments, but I'd be happy to bring these changes to master as well

---

Thanks for taking the time to look at this, and hopefully I'm not thinking of this 100% wrong :wink:

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -34,6 +34,7 @@ module Sidekiq
     dead_timeout_in_seconds: 180 * 24 * 60 * 60, # 6 months
     reloader: proc { |&block| block.call },
     executor: proc { |&block| block.call },
+    heartbeat_expiration: 60,
   }
 
   DEFAULT_WORKER_OPTIONS = {
@@ -185,6 +186,10 @@ module Sidekiq
   end
   def self.logger=(log)
     Sidekiq::Logging.logger = log
+  end
+
+  def self.heartbeat_expiration=(timeout)
+    self.options[:heartbeat_expiration] = timeout
   end
 
   # How frequently Redis should be checked by a random Sidekiq process for
--- a/lib/sidekiq/launcher.rb
+++ b/lib/sidekiq/launcher.rb
@@ -90,7 +90,7 @@ module Sidekiq
             Processor::WORKER_STATE.each_pair do |tid, hash|
               conn.hset(workers_key, tid, Sidekiq.dump_json(hash))
             end
-            conn.expire(workers_key, 60)
+            conn.expire(workers_key, @options[:heartbeat_expiration])
           end
         end
         fails = procd = 0
@@ -100,7 +100,7 @@ module Sidekiq
             conn.sadd('processes', key)
             conn.exists(key)
             conn.hmset(key, 'info', to_json, 'busy', Processor::WORKER_STATE.size, 'beat', Time.now.to_f, 'quiet', @done)
-            conn.expire(key, 60)
+            conn.expire(key, @options[:heartbeat_expiration])
             conn.rpop("#{key}-signals")
           end
         end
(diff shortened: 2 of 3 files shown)


mperham:
It sounds like you want super_fetch to recover jobs ASAP, ideally within seconds.  This does not handle poison pills: job that execute a codepath which leads to process crash.  Recovering that job will lead to a repeating cycle of process death and likely overflowing queues.

lygaret:
Yeah, you're right, but in our particular situation, that's not what's causing our crashes, rather it's just lots and lots of memory issues (somewhat flakey native extensions with high memory use + spiky load). We're working in parallel on fixing that, but quick job recovery is definitely what we'd like.

We've thought of other ways to go about this, but being able to tune the heartbeat is something that's been part of most of the ideas we've had. 

You updated the wiki yesterday to say "In summary, super_fetch might recover jobs in 5 minutes or 3 hours, there's no guarantee", which we totally understand, but we'd really like to figure out if there's a way to get that lower. Maybe we're just looking in the wrong place: most of our jobs wouldn't make sense 3 hours from now, and even 1 full minute would feel laggy.

lygaret:
Another idea I've had is to record the heartbeat identity to a file in a startup hook, and then manually expire that heartbeat on process restart with a wrapper script, somewhat like the PIDfile. 

Then we wouldn't have to do anything to the heartbeat TTL, and we wouldn't need a separate worker to run `cleanup_the_dead` (since the heartbeat would be expired at process start and the built in call would catch it.)

Basically, we really, really want this to work for our use case ;)

mperham:
If your jobs don't make sense in X hours, take a look at Pro's expiring jobs feature to auto-prune them after a deadline.

The tension here is that one of super_fetch's specific goals is to gracefully handle poison pills by NOT recovering crashed jobs quickly.  If you are using native extensions, poison pills can easily pop up at any time.  I can't accept a change which makes this easier to turn off.  Sidekiq is an async system and jobs can be delayed some amount.

I would be willing to extract that 60 into a constant but I won't go so far as to make it a configuration knob.  NB: lowering the TTL also means increasing possible false alarms around process death if the network partitions for ~30 seconds.

lygaret:
Well understood; I'm going to go ahead and close this at the moment then, I appreciate your time.

If hand-wave away the way this is currently implement though, I'm curious if you have any suggestions as to how to move forward with this use-case. One thing I've thought of is to quickly drain orphaned jobs to the failed queue, possibly via a watchdog on the process itself, in such a way that they can go through normal retry logic - that could at least help mitigate a poison pill situation as they'll end up in the morgue eventually rather than in a crasher loop.

We were really hoping that pro's reliability features were going to help us with lost jobs from crashes, and we're getting close, so I appreciate your work, and time, thank you.

PR #3636, opened 2017-10-12
Use pessimistic version operator instead of less-than in docs

will-in-wi:
Technically, the code as is allows any version less than the given version, which would include `0.0.1`. This is a little more common and easier to grok. It's also consistent with SemVer.

Not a big deal, just thought I'd suggest it. Thanks!


mperham:
thanks but I've had these docs for years and no one's had problems with it as is.  Let sleeping dogs lie.

PR #3644, opened 2017-10-19
Add queue latency to the dashboard

dsantosmerino:
In order to improve the visibility of our queues, we would like to get more details in the UI.


mperham:
Thanks, this is a great idea.

PR #3647, opened 2017-10-26
Fix support link in README.md

srushe:
The link to the Support page, for happy hour, is pointing to the incorrect url.


mperham:
danke!

PR #3649, opened 2017-10-26
Added warning to STDERR if we detect Rails 3.x (< 4)

amasses:
Due to Sidekiq dropping support for Rails 3.2 I have added a warning to STDERR (similar to the one in testing.rb) about no longer supporting 3.2 for Server/Workers.

### DIFF
--- a/lib/sidekiq/rails.rb
+++ b/lib/sidekiq/rails.rb
@@ -49,3 +49,9 @@ module Sidekiq
     end
   end if defined?(::Rails)
 end
+
+if defined?(::Rails) && ::Rails::VERSION::MAJOR < 4
+  $stderr.puts("**************************************************")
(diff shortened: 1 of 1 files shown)


mperham:
Did this cause problems?  If so, what did it do?

amasses:
Yeah, we had an upgrade in which the developer who updated the gem didn't test the `sidekiq` binary (which was terminating as an expected behaviour), and when this was deployed it resulted in the process (executed by Foreman/Upstart) to chew up CPU time due to unbounded respawning. 

We have an additional test around this now in our pipeline, but I thought it might be helpful to have a more obvious warning in the application output.

mperham:
Yikes! Thanks.

amasses:
No problems - wasn't a *huge* issue, and it did highlight some missing metrics on our systems as well, so in the end a good outcome :) 

Thanks for merging!

PR #3650, opened 2017-10-28
CI against Ruby 2.2.8/2.3.5/2.4.2

284km:


mperham:
I don't really see the point in bumping the patch numbers here for every Ruby release.  Seems fine as is.

PR #3652, opened 2017-11-01
JobSet#each goes through elements in descending score order

Ferdy89:
Sorry for not having submitted and issue before, this was simple enough to explain better through code.

The current implementation of the #each method uses Redis.zrange to
paginate the iteration and use multiple lightweight calls. It performs
this pagination in descending score order, but each page is returned
from Redis in ascending order. The result is that the final iteration
through the whole set is not sorted properly. Here's an example with a
page of size 3:

Redis set: 1, 2, 3, 4, 5, 6, 7, 8, 9
JobSet.to_a: 7, 8, 9, 4, 5, 6, 1, 2, 3

This fixes it with barely no performance cost (each page is reverted in
Ruby) and all the items are perfectly sorted in descending score order.

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -548,7 +548,7 @@ module Sidekiq
         end
         break if elements.empty?
         page -= 1
-        elements.each do |element, score|
+        elements.reverse.each do |element, score|
           yield SortedEntry.new(self, score, element)
         end
         offset_size = initial_size - @_size
--- a/test/test_api.rb
+++ b/test/test_api.rb
@@ -230,6 +230,16 @@ class TestApi < Sidekiq::Test
         assert_equal 0, q.size
       end
 
+      it 'enumerates jobs in descending score order' do
+        # We need to enqueue more than 50 items, which is the page size when retrieving
+        # from Redis to ensure everything is sorted: the pages and the items withing them.
+        51.times { ApiWorker.perform_in(100, 1, 'foo') }
+
+        set = Sidekiq::ScheduledSet.new.to_a
+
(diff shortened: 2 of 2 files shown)


mperham:
I think you might be confusing a bit of madness to my logic.  In the Web UI, if you go to a certain page, do you expect to see the elements in descending order of time?  I think you expect to see the youngest first, right?  This change would reverse that.

We go through pages in reverse order so to better deal with deletes while iterating.  But we go through elements within a page in normal order so they display on the Web UI in the expected order.  The user sees what they expect, the API iterates through each element once and handles deletes gracefully.  The only caveat: it's not the actual order of the elements in Redis.

WDYT?

Ferdy89:
My apologies, I should have explained my use case before talking about a "fix" without context. I don't have complaints about the Web UI, I'm using the API programmatically to fetch results from a `JobSet` and I want to gather the entries for a given day. When doing so, I'm iterating through the Set and I was expecting to be able to stop as soon as the current element is not the given date, because I was assuming the Set was sorted.

With that in mind, I don't think this is a bug but more like unexpected behavior, in my opinion.

I actually think the Web UI uses another piece of code to retrieve the info even more efficiently so the page size can be customized https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/paginator.rb. Not only that, but seems to get the items in ascending order (unless the `:reverse` option is present, which is only true for the morgue). In that sense, I don't think any users would be affected unless they're explicitly using the programmatic API. Even if they're using it, I don't think they're relying on the order of the elements because of the code on this PR.

mperham:
You're right!  Thanks for figuring this out.

PR #3658, opened 2017-11-07
Change "really needs" to "must" in redis_connection.rb

ethagnawl:
This is a semantic change to `verify_sizing`'s `ArgumentError` message in redis_connection.rb. 

In my mind, "must have" is more prescriptive than "really needs to have". The wording of the existing message implies that the program will continue run, but the user should do something differently in the future.


mperham:
fair enough, thanks!

PR #3666, opened 2017-11-15
update redis namespace

ota42y:
The redis-namespace 1.6.0 released.
And gemspec allow installing 1.6.0.
https://github.com/mperham/sidekiq/blob/master/sidekiq.gemspec#L24

So there aren't necessary to lock master branch already.


mperham:
thank you!

PR #3668, opened 2017-11-17
Update to reflect new Twitter character limit

jgodson:
Twitter now allows 280 characters, which still makes support over it difficult 👍


mperham:
ae6f4aa7

Issue #3673, opened 2017-11-22
Sidekiq error handler plus process fetch errors need to be configurable


ludwick:
Ruby version: 2.4.1
Sidekiq / Pro / Enterprise version(s): sidekiq 5.0.5

Issue:

During Redis failover, the sidekiq logs are excessively noisy giving us one log record per call stack line when some redis calls (naturally) fail. We use a structured logger so we could easily make that one log record (it goes to a fluentd process ultimately) and we do with other exceptions in the system. Getting one error/warn record per actual underlying error helps with error reporting.

The two places this matters that we weren't already handling are:

* The [default exception handler](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/exception_handler.rb#L8) logs each element of the backtrace on a separate line. This one I can easily override, though it's a bit weird as I need to delete the existing handler by class name and add our own.
* The [other](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/processor.rb#L110) is `handle_fetch_exception` in `Sidekiq::Processor` which I can't override and can't figure out how to monkeypatch.

I'll see if I can send you a PR if this is a change you're willing to accept. It seems pretty straightforward to re-use the existing `Sidekiq.error_handlers` for the fetch exception which is all I really need.

mperham:
Sure, PRs are always welcome to improve error handling and minimize noise.

ludwick:
Thanks! I just wanted to check in case there was a reason why `Sidekiq::Processor` couldn't use that handler.

PR #3675, opened 2017-11-22
Use exception handler for logging for fetch exception

ludwick:
Extends the `ExceptionHandler` to support multiple log levels and
additional context text in order to use it when the `Processor`
encounters an exception when fetching new work. The added options ensure
that the new behavior is fairly close to the old behavior (same log
level, same fetch error message).

Fixes mperham/sidekiq#3673


mperham:
Nice work, I appreciate the effort to make the API backwards compatible.

ludwick:
Awesome thank you!

PR #3677, opened 2017-11-28
Make error handler change fully backwards compatible

ludwick:
The change that Fixes mperham/sidekiq#3673 actually broke any existing
exception handlers due to differences in expected parameter count. This
fixes it explicitly which normally seems like it won't matter but there
are random monitoring gems that install their own sidekiq exception
handlers that I don't want to break.

Sorry I broke things. :(

### DIFF
--- a/lib/sidekiq/exception_handler.rb
+++ b/lib/sidekiq/exception_handler.rb
@@ -24,7 +24,13 @@ module Sidekiq
     def handle_exception(ex, ctxHash={}, options={})
       Sidekiq.error_handlers.each do |handler|
         begin
-          handler.call(ex, ctxHash, options)
+          arity = handler.method(:call).arity
+          # new-style three argument method or fully variable arguments
+          if arity == -3 || arity == -1
+            handler.call(ex, ctxHash, options)
+          else
+            handler.call(ex, ctxHash)
+          end
         rescue => ex
           Sidekiq.logger.error "!!! ERROR HANDLER THREW AN ERROR !!!"
           Sidekiq.logger.error ex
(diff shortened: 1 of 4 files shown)


mperham:
No, i don’t want arity checks. Maybe you should reconsider the original impl.

ludwick:
Customize the logging how? The issue is I wanted to have "single line" (json) format logs and can't get that with the existing handler. The only reason I did it the way I did (bringing in the options) was I was concerned some folks would have alerting / log processing that cues off the log level of the fetch errors.

ludwick:
I guess, to add on, my issue is I want unified / simplified logging and to me these are pretty similar cases that I want to handle similarly. The unit tests are written with a `proc` for the handler which works fine with the extra argument to call but of course not everyone's will be. I could add another handler list on `Sidekiq` but I'm guessing most users would ultimately just have both go to the same places.

mperham:
Why not just change processor.rb to call handle_exception like every place else?  Isn't that the crux of your issue: it's a special snowflake and logging the backtrace directly rather than calling handle_exception?

ludwick:
I did make it call handle exception -- it's just I was trying to maintain pretty strict backwards compatibility about what gets logged in terms of strings and log levels that folks might be alerting on. :/

I did just find another case where I'd want to change so I need to revisit this and think about it some more. I thought I had searched for *all* cases of backtrace logging like this was before but apparently I missed another one.

mperham:
Sounds good, I like the idea of ditching any backtrace logging outside of handle_exception.

ludwick:
Okay. Compromise change idea. Sorry to be such a bother. I've never had to think quite so broadly about intended use of a code change (that is, sidekiq is awesome and very very popular!)

PR #3680, opened 2017-12-02
Symbolize all config.yml keys when loading

### LINKED ISSUE #3672: Config file keys inconsistent types
Ruby version: 2.4.1
Sidekiq / Pro / Enterprise version(s): 5.0.5

It is confusing and error-prone for the sidekiq config file to require a mix of string keys and symbol keys. It currently appears that options need to be symbol keys, but environment groups need to be string keys.
I ran into issues with my workers yesterday when setting up a new project because of `queues:` vs `:queues:` in config/sidekiq.yml.

Rails config files tend to use strings everywhere in the Yaml, and then load as a HashWithIndifferentAccess. I don't necessarily think we want to depend on ActiveSupport for that, but other options might include `Hash#deep_symbolize_keys` or `Hash#deep_stringify_keys` and always doing hash lookups with a consistent type.

dpogue:
This ensures that keys in the Yaml file can be specified as either symbols or strings, but will always be treated as symbols after loading.

It tries to use ActiveSupport's `deep_symbolize_keys!` method if ActiveSupport is loaded, otherwise falling back to an inline version.

Closes #3672.

### DIFF
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -214,6 +214,14 @@ module Sidekiq
       @environment = cli_env || ENV['RAILS_ENV'] || ENV['RACK_ENV'] || 'development'
     end
 
+    def symbolize_keys_deep!(hash)
+      hash.keys.each do |k|
+        symkey = k.respond_to?(:to_sym) ? k.to_sym : k
+        hash[symkey] = hash.delete k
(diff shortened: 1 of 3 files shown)


mperham:
Thank you!

vassilevsky:
Thanks, Darryl! 👍

PR #3688, opened 2017-12-12
Warn users of the Redis persistence requirement

f3ndot:
Very happy to move this around within the README, change phrasing, etc.

I was bit by not realizing my Redis wasn't configured optimally for Sidekiq and had some dataloss when my instances ran out of memory.

It was at this point I realized the errs of my ways, and found this wiki page post-mortem: https://github.com/mperham/sidekiq/wiki/Using-Redis#memory

I figure that other developers would be able to avoid my mistake (and lack of critical thinking) by exposing quite visibly the Redis requirements Sidekiq expects.


mperham:
This is a little too specific to be placed in the readme.  I would be happy if you want to call out that memory section a little more and give specific configuration snippets.

PR #3715, opened 2018-01-10
Fix nested Sidekiq::Client.via

mdehoog:
We're using `Sidekiq::Client.via(pool)` for sharding our jobs to different Redis'. However, nested `via`s are broken, even if passing the same `pool`. This is because the `ensure` clears the `Thread.current`, even if it didn't get changed.

Fixed by not clearing the `Thread.current` if a pool is already defined in the nested `via`.

### DIFF
--- a/lib/sidekiq/client.rb
+++ b/lib/sidekiq/client.rb
@@ -123,7 +123,7 @@ module Sidekiq
       Thread.current[:sidekiq_via_pool] = pool
       yield
     ensure
-      Thread.current[:sidekiq_via_pool] = nil
+      Thread.current[:sidekiq_via_pool] = nil unless current_sidekiq_pool
     end
 
     class << self
--- a/test/test_client.rb
+++ b/test/test_client.rb
@@ -202,7 +202,20 @@ class TestClient < Sidekiq::Test
       conn.verify
     end
 
-    it 'allows #via to point to different Redi' do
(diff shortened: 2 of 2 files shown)


mperham:
via is not reentrant.  this is documented not to work, right?

https://github.com/mdehoog/sidekiq/blob/ff66f7d25d3ddbc0fb8e50b5829d75e58fe3ebab/lib/sidekiq/client.rb#L122

mperham:
Would this be reentrant and work for you?

```ruby
def self.via(pool)
  raise ArgumentError, "No pool given" if pool.nil?
  current_sidekiq_pool = Thread.current[:sidekiq_via_pool]
  Thread.current[:sidekiq_via_pool] = pool
  yield
ensure
  Thread.current[:sidekiq_via_pool] = current_sidekiq_pool
end
```

mdehoog:
Yes, it's not reentrant, but it allows reentry if the `pool` passed is the same as the previous `pool` passed to `via`. This is covered in spec "allows #via to point to same Redi" (notice the nested `via`). However, on the way out of the inner via, it clears the pool, leaving the outer via without the pool passed. This PR fixes that bug.

Yes, making it reentrant would also work fine; this is how we monkey-patched it in our code.

mdehoog:
Thanks for fixing!

PR #3718, opened 2018-01-11
Use active record cache when available

sobrinho:
This will prevent duplicated queries on database at the same job.

### DIFF
--- a/lib/sidekiq/middleware/server/active_record_cache.rb
+++ b/lib/sidekiq/middleware/server/active_record_cache.rb
@@ -1,0 +1,11 @@
+module Sidekiq
+  module Middleware
+    module Server
+      class ActiveRecordCache
+        def call(*args, &block)
+          ::ActiveRecord::Base.cache(&block)
+        end
+      end
+    end
+  end
+end
--- a/lib/sidekiq/rails.rb
+++ b/lib/sidekiq/rails.rb
@@ -9,10 +9,15 @@ module Sidekiq
     # class block. Definitely before config/environments/*.rb and
     # config/initializers/*.rb.
     config.before_configuration do
-      if ::Rails::VERSION::MAJOR < 5 && defined?(::ActiveRecord)
+      if defined?(::ActiveRecord)
         Sidekiq.server_middleware do |chain|
-          require 'sidekiq/middleware/server/active_record'
-          chain.add Sidekiq::Middleware::Server::ActiveRecord
+          if ::Rails::VERSION::MAJOR < 5
+            require 'sidekiq/middleware/server/active_record'
+            chain.add Sidekiq::Middleware::Server::ActiveRecord
+          end
+
+          require 'sidekiq/middleware/server/active_record_cache'
+          chain.add Sidekiq::Middleware::Server::ActiveRecordCache
         end
       end
     end


sobrinho:
Not sure how to test this, the `Sidekiq::Middleware::Server::ActiveRecord` seems to be not tested either.

mperham:
Rails and Sidekiq already have a bad reputation for memory bloat, often due to ActiveRecord memory usage.  Should we be enabling this by default or instead documenting its availability?  Is it a thread-local cache that is cleared at the end of each job?

sobrinho:
Rails use the appropriated machinery for concurrency and since this cache is enabled by default around HTTP requests, we can rely on that it works.

The cache is cleared after the job finishes, see here: https://github.com/rails/rails/blob/d6466beb9fff9f2ba4f73673e65f087dd6bba488/activerecord/lib/active_record/connection_adapters/abstract/query_cache.rb#L53-L60

This cache is supposed to avoid hitting the database unnecessarily in cases of repeated queries during the block execution, like this:

``` ruby
[31] pry(main)> Conceito.find(1)
  Conceito Load (25.0ms)  SELECT "conceitos".* FROM "conceitos" WHERE "conceitos"."id" = $1 LIMIT 1  [["id", 1]]
=> #<Conceito ...>
[32] pry(main)> Conceito.find(1)
  Conceito Load (7.9ms)  SELECT "conceitos".* FROM "conceitos" WHERE "conceitos"."id" = $1 LIMIT 1  [["id", 1]]
=> #<Conceito ...>
[33] pry(main)> ActiveRecord::Base.cache { Conceito.find(1) }
  Conceito Load (0.5ms)  SELECT "conceitos".* FROM "conceitos" WHERE "conceitos"."id" = $1 LIMIT 1  [["id", 1]]
=> #<Conceito ...>
[34] pry(main)> ActiveRecord::Base.cache { Conceito.find(1); Conceito.find(1); }
  Conceito Load (0.6ms)  SELECT "conceitos".* FROM "conceitos" WHERE "conceitos"."id" = $1 LIMIT 1  [["id", 1]]
  CACHE (0.0ms)  SELECT "conceitos".* FROM "conceitos" WHERE "conceitos"."id" = $1 LIMIT 1  [["id", 1]]
=> #<Conceito ...>
```

I have been using this middleware in production for a couple months now and I'm seeing excellent perf improvements.

mperham:
:+1: Let's see what happens, I think it's time to ship 5.1.0.

seuros:
Hmm, I don't think this middleware should be by default.

What if someone use sidekiq to update the cache ?

mperham:
After a quick reading, the cache seems very conservative but it is not thread-local.  It clears the cache after any update/insert/etc but the data is shared process-wide.

You can also do this within your job:

```
ActiveRecord::Base.uncached do
  # some important SELECT that must hit the DB
end
```

mperham:
Get the old behavior by removing the middleware:

```ruby
  config.server_middleware do |chain|
    chain.remove Sidekiq::Middleware::Server::ActiveRecordCache
  end
end
```

bobber205:
Will this have any effect is we don't use ActiveRecord at all?

webandtech:
👍 Nice! we do this ourselves today... excited to delete code!

sobrinho:
@bobber205 it won't have any effect as it is only injected when AR is available.

@mperham I did a quick research on Rails codebase and found it using a thread variable, where did you find that this is shared across the entire proccess?

mperham:
Enabled is thread local. The cache is a simple Hash that appears to be shared process-wide.

sobrinho:
I'm not sure why but I'm not seeing the cache being shared across the entire process:

```ruby
(development) main:0> Thread.new { ActiveRecord::Base.cache { User.select(:id).find(74); sleep; } }
=> #<Thread:0x00007fbd1fd06860@(pry):5 sleep>
(development) main:0>   User Load (1.6ms)  SELECT  "users"."id" FROM "users" WHERE "users"."id" = $1 LIMIT 1  [["id", 74]]
(development) main:0>
(development) main:0> ActiveRecord::Base.cache { User.select(:id).find(74) }
  User Load (0.5ms)  SELECT  "users"."id" FROM "users" WHERE "users"."id" = $1 LIMIT 1  [["id", 74]]
=> #<User id: 74>
```

Do you see any issues with that?

I mean, I expect this cache to be thread-wide because when the block finishes, the cache is cleaned and that would impact concurrent threads by clearing the cache.

_PS: Seems like there is a mutex to avoid concurrent errors but the cache being cleared would be a problem._

mperham:
I just read the code; the only thread-local usage is for the enabled flag.  It's possible the cache is not enabled in the Threads you used.

sobrinho:
I don't think is that because it is working on each thread independently:

```ruby
(development) main:0> Thread.new { ActiveRecord::Base.cache { 3.times { User.select(:id).find(74) }; sleep; } }
=> #<Thread:0x00007fbd1de26cb0@(pry):10 sleep>
(development) main:0>   User Load (12.8ms)  SELECT  "users"."id" FROM "users" WHERE "users"."id" = $1 LIMIT 1  [["id", 74]]
  CACHE (0.1ms)  SELECT  "users"."id" FROM "users" WHERE "users"."id" = $1 LIMIT 1  [["id", 74]]
  CACHE (0.0ms)  SELECT  "users"."id" FROM "users" WHERE "users"."id" = $1 LIMIT 1  [["id", 74]]
(development) main:0>
(development) main:0> ActiveRecord::Base.cache { 3.times { User.select(:id).find(74) }
(development) main:0> }
  User Load (2.7ms)  SELECT  "users"."id" FROM "users" WHERE "users"."id" = $1 LIMIT 1  [["id", 74]]
  CACHE (0.0ms)  SELECT  "users"."id" FROM "users" WHERE "users"."id" = $1 LIMIT 1  [["id", 74]]
  CACHE (0.0ms)  SELECT  "users"."id" FROM "users" WHERE "users"."id" = $1 LIMIT 1  [["id", 74]]
=> 3
```

The first call hits the database and the second and third uses the cache.

_PS: I kept the threads alive during the cache block by using a sleep._

Now in the main thread, I see the same behavior (first: miss, second and third: hit).

Even using other child threads I see the same behavior (to see if it was related somehow to the main thread):

```ruby
(development) main:0> Thread.new { ActiveRecord::Base.cache { 3.times { User.select(:id).find(74) }; sleep; } }
=> #<Thread:0x00007fbd1fb39c58@(pry):13 sleep>
(development) main:0>   User Load (2.5ms)  SELECT  "users"."id" FROM "users" WHERE "users"."id" = $1 LIMIT 1  [["id", 74]]
  CACHE (0.0ms)  SELECT  "users"."id" FROM "users" WHERE "users"."id" = $1 LIMIT 1  [["id", 74]]
  CACHE (0.0ms)  SELECT  "users"."id" FROM "users" WHERE "users"."id" = $1 LIMIT 1  [["id", 74]]
(development) main:0>
(development) main:0> Thread.new { ActiveRecord::Base.cache { 3.times { User.select(:id).find(74) }; sleep; } }
=> #<Thread:0x00007fbd1fafdc30@(pry):14 sleep>
(development) main:0>   User Load (3.0ms)  SELECT  "users"."id" FROM "users" WHERE "users"."id" = $1 LIMIT 1  [["id", 74]]
  CACHE (0.0ms)  SELECT  "users"."id" FROM "users" WHERE "users"."id" = $1 LIMIT 1  [["id", 74]]
  CACHE (0.0ms)  SELECT  "users"."id" FROM "users" WHERE "users"."id" = $1 LIMIT 1  [["id", 74]]
```

I'm leaving these observations for future reference if anyone hits in a bug because of that, I haven't checked the Rails implementation.

Tested on Rails 5.1 and 4.2.

Thanks!

aprescott:
In case it helps anyone else: we were just bitten by trying to remove the middleware in an app which _doesn't_ use Rails, but _does_ use ActiveRecord. The `Sidekiq::Middleware::Server::ActiveRecordCache` is only defined and added if `rails.rb` is required, which is only when Rails is in use. We initially thought it would be available if AR was in use, independent of Rails. As such, we hit a constant missing error.

Not sure if this middleware should be added in a non-Rails app that uses AR, but it might be worth clarifying things in the 5.1.0 changelog notes.

rafbm:
In case some folks get scared by [this comment above](https://github.com/mperham/sidekiq/pull/3718#issuecomment-357306695):

> After a quick reading, the cache seems very conservative but it is not thread-local. It clears the cache after any update/insert/etc but the data is shared process-wide.

The `ActiveRecord::ConnectionAdapters::QueryCache` code is not explicitly thread-local indeed, but that module is included in `ActiveRecord::ConnectionAdapters::AbstractAdapter` instances, _which are_ thread-local.

You can confirm this by monkey-patching `puts 'QueryCache#initialize'` into the initialize method, then doing a few queries in the console like this (some lines removed for clarity)

```
irb(main):001:0> User.first
QueryCache#initialize
  User Load (1.3ms)  SELECT  "users".* FROM "users" ORDER BY "users"."id" ASC LIMIT 1

irb(main):002:0> User.first
  User Load (1.1ms)  SELECT  "users".* FROM "users" ORDER BY "users"."id" ASC LIMIT 1

irb(main):003:0> Thread.new { User.first }.join
QueryCache#initialize
  User Load (1.6ms)  SELECT  "users".* FROM "users" ORDER BY "users"."id" ASC LIMIT 1

irb(main):004:0> Thread.new { User.first; User.first }.join
QueryCache#initialize
  User Load (1.5ms)  SELECT  "users".* FROM "users" ORDER BY "users"."id" ASC LIMIT 1
  User Load (0.5ms)  SELECT  "users".* FROM "users" ORDER BY "users"."id" ASC LIMIT 1
```

ActiveRecord connections are always thread-local, otherwise it would be a nightmare to use in any multi-threaded context. So there’s definitely no risk in leaving the query cache enabled in Sidekiq 5.1+, [as sobrinho said](https://github.com/mperham/sidekiq/pull/3718#issuecomment-357224066):

> Rails use the appropriated machinery for concurrency and since this cache is enabled by default around HTTP requests, we can rely on that it works.

PR #3720, opened 2018-01-12
Fix japanese locale of web

ogom:
Rename `Dead`  from Hani to Kana. (Hani and Kana are [ISO 15924 ](http://unicode.org/iso15924/iso15924-codes.html))
There is no Japanese locale to match the dead.

Also, Add  `BackToApp` and `Latency`.


mperham:
Thank you!

PR #3732, opened 2018-01-25
adds pronunciation instruction to README

jasonfb:


mperham:
Thanks, it is true but I don't think it's README-worthy.

jasonfb:
Your call, all I can tell you is I've met a few devs who mis-pronounce it. also confusing because it ends in a "q" with no following "u" so it breaks normal rules of English. nonetheless thanks for the consideration! 👍

jasonfb:
some pronounce it "sidequeue"

vassilevsky:
[ciderkik] is also a thing

PR #3745, opened 2018-02-02
Make links out of all issue numbers in the changelog

sgrif:
If your changelog says "see issue for details", you should make it easy
for someone to get there.


sgrif:
(This was done with a regex looking for `\#(\d\d\d\d)` so I likely missed super old issues if you were referencing them before 4 digits)

mperham:
Thanks Sean!

PR #3748, opened 2018-02-06
bump redis client gem dependency

camallen:
Ensure compatability with the redis '_client' method introduced in https://github.com/mperham/sidekiq/commit/89fcf0c145de9c49bab037a9e5c8d77d262c2d82

I came across this error `ERR unknown command '_client'` bumping to sidekiq 5.1.0 from 5.0.5. It appears that older versions of redis gem that do not support the new redis api can slip through the install. 

Manually bumping the required redis gem in your Gemfile is a workaround in the meantime for me.

### DIFF
--- a/sidekiq.gemspec
+++ b/sidekiq.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |gem|
   gem.version       = Sidekiq::VERSION
   gem.required_ruby_version = ">= 2.2.2"
 
-  gem.add_dependency                  'redis', '>= 3.3.4', '< 5'
+  gem.add_dependency                  'redis', '>= 4.0.1', '< 5'
   gem.add_dependency                  'connection_pool', '~> 2.2', '>= 2.2.0'
   gem.add_dependency                  'concurrent-ruby', '~> 1.0'
   gem.add_dependency                  'rack-protection', '>= 1.5.0'


mperham:
redis-rb 3.3.4 has the required method.  I'm not clear why you were seeing this error.

mperham:
Sorry if I'm not clear: I don't want to bump the Redis gem major version if I don't have to.

camallen:
Yep agree re the major version bump. I'm happy for this to remain closed.

My test case was using redis 3.3.5 on my rails console:
``` bash
> Sidekiq.redis { |conn| conn._client }
Redis::CommandError: ERR unknown command '_client'
from ../.rvm/gems/ruby-2.4.1@panoptes/gems/redis-3.3.5/lib/redis/client.rb:121:in 'call'
> Sidekiq.redis { |conn| conn }
=> #<Redis client v3.3.5 for Sidekiq-client-PID-7768>

```
To me it appears that `_client` was introduced in https://github.com/redis/redis-rb/pull/653/commits/c239abb43c2dce99468bf94652a3c48b31a1041a and released in [4.0](https://github.com/redis/redis-rb/blob/master/CHANGELOG.md#40). 

As i said, happy for this to remain closed and thanks for the wonderful software :)

mperham:
I don't see where _client was **ever** added to 3.3.x.  For that reason, I'm going to bump to 4.x.

Issue #3752, opened 2018-02-11
Understanding a leak that happens only when running job via sidekiq worker


dbenjamin57:
Sidekiq 5.1 (using Sidekiq Pro 4) with Ruby 2.4.1.
Same issue occurs with Sidekiq 4.2.10 (without pro)

We have an issue where a job of ours when running on Sidekiq leaks memory. 

The use case is as follows: We have a BI system that has millions of records in our DB. Once in a while, our customer wants to export a subset of the data to a CSV (can be 500K records). 
Therefore, we have a CSV worker that serialises the 500K using pagination to a file, uploads it then to private storage and sends the customer the link to the file. 

This job, when running locally on the server (opening rails console and running CSVWorker.new.perform(some_collection)) works fine. It takes a while but finishes. Memory stays approx. static and does not go up during the pagination. 

Then, when we try to run the job via Sidekiq (CSVWorker.new.perform_async(some_collection)) we start seeing graph charts like this (memory utilisation): 

<img width="337" alt="screen shot 2018-02-11 at 18 15 36" src="https://user-images.githubusercontent.com/1243052/36075434-aa45559c-0f57-11e8-94e2-c10bb0e2b010.png">

Eventually, the OOM kills our sidekiq process and our docker dies. To be extra careful with our test we ran our worker on an isolated sidekiq process (own server with no other workers running on it)

Any ideas? I know its not related to our code as we are running the same worker and it works via console. What can be different in this case?

Thanks in advance,
Dan

mikegee:
Other components, like ActiveRecord, might behave differently in a console versus in a Sidekiq worker.

ActiveRecord's caching feature was just enabled in 5.1: https://github.com/mperham/sidekiq/pull/3718, perhaps try turning it off? I dunno.

There's not much to go on, so maybe add some debugging logs to that job to isolate the culprit.

mperham:
Are you testing fine on OSX and leaking on Linux? Huge diff in native extensions there.

dbenjamin57:
Hey @mperham all testing is done on the same linux server. so the only real difference is between running it via a sidekiq worker and running it in console via Worker.new.perform(params)

dbenjamin57:
hey @mikegee what kind of logging would you add? The leak happens both in Sidekiq 5.1 and Sidekiq 4.2.10 so cant be attributed to this feature..

mikegee:
> what kind of logging would you add?

[...]

PR #3756, opened 2018-02-15
Revert "Use active record cache when available"

urbanautomaton:
This reverts commit a27d692656e15636cebe74542371a5e93f3f9194, introduced in #3718.

The reason for reverting is that, as best I can tell, this middleware has no effect. In Rails 5.0+, Sidekiq runs its jobs [using the Rails reloader/executor API](https://github.com/mperham/sidekiq/blob/eca6acc0ce201125a45a5af75fd08c6ce985df5a/lib/sidekiq/rails.rb#L33-L37). However, the ActiveRecord query cache [registers its own hooks with the executor](https://github.com/rails/rails/blob/f8c00c130016b248d1d409f131356632dcc418c6/activerecord/lib/active_record/railtie.rb#L159-L163), and these hooks automatically enable the query cache for the [duration of any work the executor performs](https://github.com/rails/rails/blob/f8c00c130016b248d1d409f131356632dcc418c6/activerecord/lib/active_record/query_cache.rb#L26-L41).

Since the query cache is already on by default for any job run in Rails 5.0+, ~and this middleware is only added if Rails 5.0+ is detected, I don't believe there are any circumstances in which it has an effect.~ Edit: this was incorrect, please [see below](#issuecomment-365970091)

I didn't think it made sense to contribute a test for the removal of a feature, but I do have a [small repro app](https://github.com/urbanautomaton/sidekiq-leak-repro) which disables this middleware and demonstrates that query caching is still enabled.

If there's anything else I can do to instil confidence that the middleware is unnecessary, just let me know and I'll be more than happy to try. 😄 

---

On a related note, I think it would probably forestall a lot of memory leak issues being raised if a note could be added to the [troubleshooting wiki page](https://github.com/mperham/sidekiq/wiki/Problems-and-Troubleshooting) advising users to try explicitly disabling the query cache with `ActiveRecord::Base.uncached { ... }`. I'd be happy to draft something if you think this would be useful.


sobrinho:
This is true in both production and development environments?

What about Rails 4 apps?

sobrinho:
I think there is a misunderstanding here.

The query you are doing in the worker is not the one supposed to be cached by AR.

Change the worker to:

```ruby
class LeakWorker
  include Sidekiq::Worker

  def perform(label, disable_cache=false)
    500.times { User.find(1) }
  end
end
```

Then run the worker watching the log with and without the middleware.

By doing this, you are going to see without the middleware:

```
  User Load (0.1ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = ? LIMIT ?  [["id", 1], ["LIMIT", 1]]
  User Load (0.0ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = ? LIMIT ?  [["id", 1], ["LIMIT", 1]]
  User Load (0.1ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = ? LIMIT ?  [["id", 1], ["LIMIT", 1]]
  User Load (0.1ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = ? LIMIT ?  [["id", 1], ["LIMIT", 1]]
...
```

With the middleware:

```ruby
  User Load (0.3ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = ? LIMIT ?  [["id", 1], ["LIMIT", 1]]
  CACHE (0.0ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = ? LIMIT ?  [["id", 1], ["LIMIT", 1]]
  CACHE (0.0ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = ? LIMIT ?  [["id", 1], ["LIMIT", 1]]
  CACHE (0.0ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = ? LIMIT ?  [["id", 1], ["LIMIT", 1]]
  CACHE (0.0ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = ? LIMIT ?  [["id", 1], ["LIMIT", 1]]
  CACHE (0.0ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = ? LIMIT ?  [["id", 1], ["LIMIT", 1]]
...
```

This means that only the first query is sent to the database and the other ones are retrieved from a local memory cache.

sobrinho:
Remeber that this doesn't impact performance that much on a local database (like your local machine) but you can see the benefits by running the [toxiproxy](https://github.com/Shopify/toxiproxy) to add a small delay like 2~3ms between your ruby process and the database (but probably won't work with SQLite).

urbanautomaton:
> This is true in both production and development environments?

I only tested in the production env. Re-running my tests, it appears that without the middleware, the query cache is *not* enabled by default in development mode. So the middleware would have an effect there after all.

> What about Rails 4 apps?

Ah, I'm a chump - I misread the [conditional here](https://github.com/mperham/sidekiq/blob/eca6acc0ce201125a45a5af75fd08c6ce985df5a/lib/sidekiq/rails.rb#L13-L21) and mistakenly thought the middleware is only loaded in Rails 5.0+.

It's only since Rails 5.0 that Rails has [exposed its executor / reloader API](https://github.com/rails/rails/commit/d3c9d808e3e242155a44fd2a89ef272cfade8fe8), and that sidekiq has [taken advantage of it](https://github.com/mperham/sidekiq/commit/51347c5d97ebdadd9ec1914acc5198765ab8a31a). So in Rails 4 there won't be any query caching by default without this middleware; that's another case where it has an effect. Apologies!

Re: your other comments; I'm aware of what the query cache does, and that the query pattern I used in my reproduction doesn't take advantage of it (although the results are cached; they just don't generate any hits).

That's not my point, though, which is purely whether this middleware makes any difference to whether the cache is enabled or not. And it seems the answer is "sometimes", not never as I mistakenly thought.

    Is query caching enabled?

                        | Rails 4 | Rails 5 dev | Rails 5 prod
    --------------------+---------+-------------+--------------
    without middleware  |   no    |    no       |     yes
       with middleware  |  yes    |   yes       |     yes

It seems beneficial for the Rails 4 behaviour to match the Rails 5 behaviour, but on the other hand it's somewhat confusing that removing the middleware doesn't make any difference in Rails 5 production.

Perhaps the best solution then is to document somewhere that if you want query caching off in your jobs you need to explicitly disable it with `ActiveRecord::Base.uncached { ... }`, rather than remove this middleware?

sobrinho:
Okay, to be on the same page:

1. Rails 5 executor/reloader API that's used by Sidekiq enables query caching by default in production but not on other environments
2. The Sidekiq documentation is not clear about that (I think it's a bug on Rails, TBH)
3. The middleware currently keeps the same behavior between Rails 4 and 5 regardless of the environment

I think the first thing we should do is try to fix that on the Rails side because why on earth it would not cache the queries in the development environment?

If that's fixed on Rails, we could then only attach this middleware to Rails 4 users.

The documentation would need be improved anyway by saying that the query cache must be skipped only by using `uncache { }` and not rely on Sidekiq internals for that (after all the documentation would be different for Rails 4 and Rails 5, which may be confusing and not scalable).

mperham:
From my POV: I'm not concerned about Rails 4 users in this case.  As long as Sidekiq works with Rails 4, there's no need to optimize it.

Query caching should follow the Rails conventions and however it is configured in config/environment/*.rb.  Query caching is not a Sidekiq feature but a Rails feature.

sobrinho:
Well, I'm not familiar with executor/reloader API to investigate this further yet.

Nothing too obvious came up to query cache to not be working in development mode and there is no specific configuration for query cache on `config/environment/*.rb`, even creating the `tmp/caching-dev.txt` that is supposed to enable all caches similar to production had no effect.

My concern here to remove this middleware is that Rails 4 users won't have query cache enabled at all (we may not be supporting this version anymore and that's okay) and Rails 5 users won't have query cache enabled in development mode.

The middleware is providing a common behavior for both Rails versions and all rails environments.

We may just update the Sidekiq documentation to warn that query cache is always enabled and it may be skipped over by using `ActiveRecord::Base.uncached { ... }`.

urbanautomaton:
Thanks for the responses, both.

I've dug further into the Rails 5 dev / production issue and discovered that this is an artifact of load order (I assume due to autoloading).

For the first job I run on a freshly booted sidekiq processor in Rails 5 dev, the query cache isn't enabled - for all subsequent jobs, it's enabled just like in production. So the true story is (and sorry it took so long to arrive here):

    Is query caching enabled?

                        | Rails 4 | Rails 5 dev | Rails 5 prod
    --------------------+---------+-------------+--------------
    without middleware  |   no    |   yes       |     yes
       with middleware  |  yes    |   yes       |     yes


I lean towards the view that we shouldn't be enabling the query cache for Rails 4, both for the reasons @mperham lists, and because the query cache is not guaranteed to be beneficial for background jobs. Indeed for batched reads over large numbers of records (such as you might do when generating CSV reports, see #3752) it's a guaranteed memory leak.

Given the above, I'd probably go back to my original suggestion that the middleware be removed, meaning that the default caching behaviour is no longer Sidekiq's concern. Then on the troubleshooting wiki page we can note that query caching is on by default in Rails 5, and show how to disable it.

Does that sound reasonable? If so I'll rewrite this commit message to give the correct explanation.

sobrinho:
Well, I think so.

If the middleware is only, in fact, acting for Rails 4, we can remove it.

mperham:
@sobrinho I'm ok with not having the query cache activated on Rails 4 if it simplifies the feature/code.

sobrinho:
@mperham I'm ok too, we can merge this then.

urbanautomaton:
Thanks both. I've [updated the wiki](https://github.com/mperham/sidekiq/wiki/Problems-and-Troubleshooting#activerecord-query-cache) with specific guidance re: disabling the query cache for batched reads, hope that's okay.

FWIW (and for anyone seeing this later), there appears to be a patch in Rails master to disable the query cache for precisely the pathological case of batched reads, so presumably at some point this will become less of a problem: https://github.com/rails/rails/pull/28867

krzysiek1507:
Hi @mperham Do you ETA for a new version with this revert?

mperham:
End of march.

PR #3759, opened 2018-02-16
Remove String#freeze calls

mperham:
Explicit freezing was required pre-Ruby 2.3.  With 2.3, `frozen_string_literal: true` is available to do the same thing without uglifying the code.

Caveat upgrader: this change will lead to increased memory usage and slower performance on Ruby 2.2.

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -1,4 +1,3 @@
-# encoding: utf-8
 # frozen_string_literal: true
 require 'sidekiq/version'
 fail "Sidekiq #{Sidekiq::VERSION} does not support Ruby versions below 2.2.2." if RUBY_PLATFORM != 'java' && RUBY_VERSION < '2.2.2'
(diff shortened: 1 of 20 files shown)


lib/sidekiq.rb
```diff
@@ -48,7 +47,7 @@ module Sidekiq
     "connected_clients" => "9999",
     "used_memory_human" => "9P",
     "used_memory_peak_human" => "9P"
-  }.freeze
```
SamSaffron: note this is a behavior change cause Hash is not a string literal so the hash will no longer be frozen

mperham: In this case, it's not a concern.

krzysiek1507:
Didn't you miss it in those files?
```
lib/sidekiq/delay.rb
lib/sidekiq/job_logger.rb
lib/sidekiq/job_retry.rb
lib/sidekiq/middleware/server/active_record.rb
lib/sidekiq/rails.rb
```

mperham:
thanks for the check, fixed!

krzysiek1507:
You're welcome!

PR #3781, opened 2018-03-01
Restrict Ruby 2.3.0 due to frozen string literal adoption

Tensho:
`frozen_string_literal` requires Ruby 2.3. It's worth to specify that requirement in gemspec.
https://www.ruby-lang.org/en/news/2015/12/25/ruby-2-3-0-released/

I understand it's not mandatory, as far as pragma just skipped in earlier versions of Ruby, but 2.2 is EOL.


mperham:
We don't change Ruby version support without a major version bump.

PR #3786, opened 2018-03-13
Fix warning on redis-namespace 1.5+

ixti:


ixti:
As of redis-namespace 1.5 blind passthrough of server commands being deprecated and will be completely removed in 2.0: https://github.com/resque/redis-namespace#blind-passthrough

mperham:
https://github.com/resque/redis-namespace/issues/151

PR #3797, opened 2018-03-21
Remove redundant code & set environment via ::Rails.env

manojmj92:
This PR deals with removing a redundant piece of code, which is already present in Rails, in the implementation of `Rails.env` method. The relevant piece of code is [here](https://github.com/rails/rails/blob/master/railties/lib/rails.rb#L73).

Since Rails is already implementing the same logic, in my humble opinion, I don't feel there is a need for Sidekiq to repeat the same to set the `@environment` variable.

### DIFF
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -218,7 +218,7 @@ module Sidekiq
     end
 
     def set_environment(cli_env)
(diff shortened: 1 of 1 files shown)


manojmj92:
Ah, I just realised this is wrong. User should be able to boot up Sidekiq in whatever environment they like. Closing this.

Issue #1215, opened 2013-09-30
Connecting Sidekiq to Twemproxy(Nutcracker)


felipeclopes:
I'm trying to connect my sidekiq to a twemproxy instead of a direct Redis connection, but I'm getting the following error:

```
2013-09-30T17:16:53Z 23329 TID-3q3o4 INFO: Booting Sidekiq 2.12.1 using redis://localhost:22121 with options {}
2013-09-30T17:16:53Z 23329 TID-3q3o4 INFO: Running in ruby 1.9.3p448 (2013-06-27 revision 41675) [x86_64-linux]
2013-09-30T17:16:53Z 23329 TID-3q3o4 INFO: See LICENSE and the LGPL-3.0 for licensing details.
Connection lost (ECONNRESET)
/mnt/-production/shared/bundle/ruby/1.9.1/gems/redis-3.0.4/lib/redis/client.rb:210:in `rescue in io'
/mnt/-production/shared/bundle/ruby/1.9.1/gems/redis-3.0.4/lib/redis/client.rb:206:in `io'
/mnt/-production/shared/bundle/ruby/1.9.1/gems/redis-3.0.4/lib/redis/client.rb:214:in `read'
/mnt/-production/shared/bundle/ruby/1.9.1/gems/redis-3.0.4/lib/redis/client.rb:84:in `block in call'
/mnt/-production/shared/bundle/ruby/1.9.1/gems/redis-3.0.4/lib/redis/client.rb:187:in `block (2 levels) in process'
/mnt/-production/shared/bundle/ruby/1.9.1/gems/redis-3.0.4/lib/redis/client.rb:295:in `ensure_connected'
/mnt/-production/shared/bundle/ruby/1.9.1/gems/redis-3.0.4/lib/redis/client.rb:177:in `block in process'
/mnt/-production/shared/bundle/ruby/1.9.1/gems/redis-3.0.4/lib/redis/client.rb:256:in `logging'
/mnt/-production/shared/bundle/ruby/1.9.1/gems/redis-3.0.4/lib/redis/client.rb:176:in `process'
/mnt/-production/shared/bundle/ruby/1.9.1/gems/redis-3.0.4/lib/redis/client.rb:84:in `call'
/mnt/-production/shared/bundle/ruby/1.9.1/gems/redis-3.0.4/lib/redis.rb:410:in `block in keys'
[...]

PR #3800, opened 2018-03-22
client: bypass multi/exec when pushing single message

igorwwwwwwwwwwwwwwwwwwww:
TLDR: this patch bypasses multi/exec when pushing a single message in the sidekiq client, which allows the client to be used with [twemproxy](https://github.com/twitter/twemproxy). this is extremely useful when running unicorn.

when running on unicorn, a common issue is managing the number of connections to other systems, e.g. redis. because each unicorn worker runs in its own process, each process gets its own connection, so in-process connection pooling is not possible.

a solution to this conundrum is to use an external pooling proxy such as [twemproxy](https://github.com/twitter/twemproxy).

twemproxy brings with it some limitations, however. some commands [are not supported](https://github.com/twitter/twemproxy/blob/master/notes/redis.md), including multi/exec. so in order to be able to use the sidekiq client with twemproxy, we must not use multi/exec.

side-note: `client setname` is also not supported, so the redis client must be explicitly configured with `{id: nil}` to prevent sidekiq from setting a generic client name.

it turns out that when pushing a single message, multi/exec is not needed, because in that case we will always run only a single redis command.

atomic_push will either perform a `zadd` or an `sadd` and an `lpush`. because the `sadd` is idempotent, it should be safe.

refs https://github.com/mperham/sidekiq/issues/1215

### DIFF
--- a/lib/sidekiq/client.rb
+++ b/lib/sidekiq/client.rb
@@ -71,7 +71,9 @@ module Sidekiq
       payload = process_single(item['class'], normed)
 
       if payload
-        raw_push([payload])
+        @redis_pool.with do |conn|
+          atomic_push(conn, [payload])
+        end
         payload['jid']
       end
     end


mperham:
twemproxy is not supported.  Supporting it would require loads of hacks, like this, throughout the system.  Not going to happen.

You haven't really explained why connection pooling is a problem.  Sidekiq's pool is lazy.  If you have a single threaded client, it will only create one connection per process.

Issue #3782, opened 2018-03-05
High memory consumption using Sidekiq with ActiveJob


davidpiegza:
Ruby version: 2.4.3
Sidekiq / Pro / Enterprise version(s): 5.1.1

I discovered a high memory usage when using Sidekiq together with ActiveJob. I created a sample app at https://github.com/davidpiegza/sidekiq-memory-test/pull/1. 

In my tests, the results for creating 250_000 jobs are:

With ActiveJob: 520 MB
Without ActiveJob: 60 MB

The tests run on a new created Rails app in production environment with a sidekiq concurrency of 10. This high memory consumption also happens on Heroku with multiple workers running.

mperham:
In designing and building Sidekiq, I've tried to make pragmatic trade-offs between performance and developer friendliness. I can also optimize the happy path and provide special APIs for edge cases. For instance, you can make your benchmark even faster if you use the push_bulk API. This load tester creates 100,000 native jobs in 2 seconds (vs 40 sec for AJ) on my laptop:

https://github.com/mperham/sidekiq/blob/master/bin/sidekiqload#L92-L100

AJ has much more complex Ruby serialization/deserialization logic, which will also chew up memory/CPU. Sidekiq's de/serialization is explicitly "JSON only" (one of those pragmatic decisions mentioned above) and therefore is implemented in pure C within the "json" gem.

And when you actually want to process those jobs? The 100,000 jobs take 22 seconds to process on my laptop. The same AJ jobs take 65 seconds to process, so ~3x the overhead.

elia:
Having heard from many people that is crucial not to use Sidekiq through AJ for performance reasons I was wondering about the absolute numbers of the overhead difference. I suspected that it might be not so relevant for many use-cases in which the time taken for performing the job is orders of magnitude more than the overhead of AJ. 

**Running the numbers provided by the example turns out that `(65.0 / 100_000.0 - 22.0 / 100_000.0)` results in about 0.43ms per job.**

Since many many jobs are extracted to perform slow operation out of the request-response cycle I suggest adding a note to clarify that the 3x difference is relevant only for super-large amounts of jobs that execute very fast. ⚡️

@mperham performance numbers are tricky, let me know if I missed something in this line of reasoning or if you have any other thoughts/experience on this topic 🙏

mperham:
[...]

Issue #2985, opened 2016-05-24
Better integration between Sidekiq thread count and Active Record pool size.


schneems:
Right now if someone wants to use sidekiq they must configure the number of threads they want to run on or use the default (25). If they are using a database (likely) then to not hit a connection timeout error they will need to have at least as many connections in their database pool as sidekiq threads. Currently they must configure these two values in two different places, database pool in `database.yml` for ActiveRecord, and sidekiq connections either via command line or the sidekiq yml file. This is a problem as it is difficult to remember when you are modifying one value that you need to modify both.

In Rails 5 there is a default environment variable to configure the number of threads that will be in the database connection pool

```yml
# config/database.yml
  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
```

This convention is used along side of web server configuration values. For example Rails 5 also ships with puma which is configured to read from this value:

```ruby
# config/puma.rb
threads_count = ENV.fetch("RAILS_MAX_THREADS") { 5 }.to_i
threads threads_count, threads_count
```

It is common that the number of sidekiq threads run in a "worker" process would be greater than the number of web threads used by a server like Puma. For this reason it does not make sense to default the sidekiq connections to `RAILS_MAX_THREADS`. However it would make sense for sidekiq to use this enviornment variable interface and se the environment variable:

```ruby
# sidekiq/lib/<somewhere>.rb
ENV["RAILS_MAX_THREADS"] = Sidekiq.number_of_threads #pseudo code
```

So now when the application boots in a "worker" process via the `sidekiq` command and we know that we are using Rails, the appropriate number of database connections will be configured by default.

## Potential Problems

This behavior may be slightly unexpected (for Sidekiq to change an environment variable at boot time) however I believe it is perfectly warranted. As a mitigation step we could detect when an enviornment variable was already set

```ruby
# sidekiq/lib/<somewhere>.rb
puts "Sidekiq is setting RAILS_MAX_THREADS=#{ Sidekiq.number_of_threads }, was previously #{ ENV['RAILS_MAX_THREADS'] }"
ENV["RAILS_MAX_THREADS"] = Sidekiq.number_of_threads #pseudo code
```

A warning or similar would eliminate any confusion. It would also allow the use of `RAILS_MAX_THREADS` for when the web process is booted, and a different configuration option for use when sidekiq is booted.

People may want to opt out of this behavior. They can do this by modifying their `database.yml` file to use a different environment variable

```ruby
# config/database.yml
[...]

PR #3807, opened 2018-03-27
Use `RAILS_MAX_THREADS` for client pool size

### LINKED ISSUE #3806: `RAILS_MAX_THREADS` parity with Redis pool for client (i.e. Rails server)
Ruby version: 2.5.0
Rails: 5.1.5
Sidekiq / Pro / Enterprise version(s): Sidekiq 5.1.1

## Issue

In Sidekiq 4.2.0 (issue #2985 implemented https://github.com/mperham/sidekiq/commit/52828e42123cc1c54b2a326bda5d00965ec05acb) the ability to use `RAILS_MAX_THREADS` was introduced to allow tuning of concurrency with the `RAILS_MAX_THREADS`. The problem is this only applies for the [Sidekiq server run via CLI](https://github.com/mperham/sidekiq/blob/60a41d3f36db25ff2bebade5bb516d38d4240b7b/lib/sidekiq/cli.rb#L243) (e.g. `worker: RAILS_MAX_THREADS=10 bundle exec sidekiq`), not the Rails server process (e.g. `web: bundle exec puma -C config/puma.rb`).

When the Rails server is running the connection is often made to `Sidekiq::Client` (such as with [`ActiveJob::QueueAdapters::SidekiqAdapter`](https://github.com/rails/rails/blob/v5.1.5/activejob/lib/active_job/queue_adapters/sidekiq_adapter.rb#L20) which does not have a `redis_pool` yet. It then attempts to create one [defaulting to a pool size of 5](https://github.com/mperham/sidekiq/blob/60a41d3f36db25ff2bebade5bb516d38d4240b7b/lib/sidekiq/redis_connection.rb#L18).

<details>
<summary>Sample Rails `ActiveJob` Stack</summary>

```
.gem/ruby/2.5.0/gems/sidekiq-5.1.1/lib/sidekiq.rb:125:in `redis_pool'
.gem/ruby/2.5.0/gems/sidekiq-5.1.1/lib/sidekiq/client.rb:42:in `initialize'
.gem/ruby/2.5.0/gems/sidekiq-5.1.1/lib/sidekiq/client.rb:131:in `new'
.gem/ruby/2.5.0/gems/sidekiq-5.1.1/lib/sidekiq/client.rb:131:in `push'
.gem/ruby/2.5.0/gems/activejob-5.1.5/lib/active_job/queue_adapters/sidekiq_adapter.rb:20:in `enqueue'
.gem/ruby/2.5.0/gems/activejob-5.1.5/lib/active_job/enqueuing.rb:51:in `block in enqueue'
.gem/ruby/2.5.0/gems/activesupport-5.1.5/lib/active_support/callbacks.rb:108:in `block in run_callbacks'
.gem/ruby/2.5.0/gems/activejob-5.1.5/lib/active_job/logging.rb:15:in `block (3 levels) in <module:Logging>'
.gem/ruby/2.5.0/gems/activejob-5.1.5/lib/active_job/logging.rb:44:in `block in tag_logger'
.gem/ruby/2.5.0/gems/activesupport-5.1.5/lib/active_support/tagged_logging.rb:69:in `block in tagged'
.gem/ruby/2.5.0/gems/activesupport-5.1.5/lib/active_support/tagged_logging.rb:26:in `tagged'
.gem/ruby/2.5.0/gems/activesupport-5.1.5/lib/active_support/tagged_logging.rb:69:in `tagged'
.gem/ruby/2.5.0/gems/activejob-5.1.5/lib/active_job/logging.rb:44:in `tag_logger'
.gem/ruby/2.5.0/gems/activejob-5.1.5/lib/active_job/logging.rb:14:in `block (2 levels) in <module:Logging>'
.gem/ruby/2.5.0/gems/activesupport-5.1.5/lib/active_support/callbacks.rb:117:in `instance_exec'
.gem/ruby/2.5.0/gems/activesupport-5.1.5/lib/active_support/callbacks.rb:117:in `block in run_callbacks'
.gem/ruby/2.5.0/gems/activesupport-5.1.5/lib/active_support/callbacks.rb:135:in `run_callbacks'
.gem/ruby/2.5.0/gems/activejob-5.1.5/lib/active_job/enqueuing.rb:47:in `enqueue'
.gem/ruby/2.5.0/gems/activejob-5.1.5/lib/active_job/enqueuing.rb:18:in `perform_later'
```

</details>

<br/>

Depending on network speed, server load, job data size, and the `RAILS_MAX_THREADS` setting it's possible all of the Redis connections are in use resulting in a `Timeout::Error`.

## Workaround

Use an initializer to explicitly set the pool size:

```ruby
# frozen_string_literal: true

# See https://github.com/mperham/sidekiq/wiki/Using-Redis
  # Without an explicit `size` here the pool will be set to 5
  #
  # See:
  #   - https://github.com/mperham/sidekiq/blob/v5.1.1/lib/sidekiq/redis_connection.rb#L18
  #   - https://github.com/mperham/sidekiq/blob/v5.1.1/lib/sidekiq/cli.rb#L244
  config.redis = {
    size: Integer(ENV.fetch("RAILS_MAX_THREADS", 5)),
  }
end
```

## Proposed Solution

I believe to resolve this the Redis pool creation needs similar logic to the CLI. I recognize this is complicated by the fact there is both a server and a client. I think the solution is something like the following, but I don't have enough knowledge of the Sidekiq internals to say for certain:

```ruby
# Sidekiq::RedisConnection.create (lib/sidekiq/redis_connection.rb#L18)
# ...

size = if options[:size]
         options[:size]
       elsif Sidekiq.server?
         Sidekiq.options[:concurrency] + 5
       elsif ENV["RAILS_MAX_THREADS"]
         Integer(ENV["RAILS_MAX_THREADS"])
       else
         5
       end

verify_sizing(size, Sidekiq.options[:concurrency]) if Sidekiq.server?

# ...
```

cupakromer:
This is a follow up to #2985 (52828e4) adding similar support for the client connection pool. For Rails servers, Sidekiq is not loaded from the CLI so the prior change to support setting the concurrency via `RAILS_MAX_THREADS` is not applied to the web server process. This means for Rails servers which do not configure a custom size through an initializer they will run with the default connection pool size of 5.

When the Rails server runs the initial Redis connection may be made through `Sidekiq::Client` (e.g. from [`ActiveJob::QueueAdapters::SidekiqAdapter`](https://github.com/rails/rails/blob/v5.1.5/activejob/lib/active_job/queue_adapters/sidekiq_adapter.rb#L20)). This causes the `redis_pool` to be initialized without any options, setting the pool size to the default of 5.

    .gem/ruby/2.5.0/gems/sidekiq-5.1.1/lib/sidekiq.rb:125:in `redis_pool'
    .gem/ruby/2.5.0/gems/sidekiq-5.1.1/lib/sidekiq/client.rb:42:in `initialize'
    .gem/ruby/2.5.0/gems/sidekiq-5.1.1/lib/sidekiq/client.rb:131:in `new'
    .gem/ruby/2.5.0/gems/sidekiq-5.1.1/lib/sidekiq/client.rb:131:in `push'
    .gem/ruby/2.5.0/gems/activejob-5.1.5/lib/active_job/queue_adapters/sidekiq_adapter.rb:20:in `enqueue'
    .gem/ruby/2.5.0/gems/activejob-5.1.5/lib/active_job/enqueuing.rb:51:in `block in enqueue'
    .gem/ruby/2.5.0/gems/activesupport-5.1.5/lib/active_support/callbacks.rb:108:in `block in run_callbacks'
    .gem/ruby/2.5.0/gems/activejob-5.1.5/lib/active_job/logging.rb:15:in `block (3 levels) in <module:Logging>'
    .gem/ruby/2.5.0/gems/activejob-5.1.5/lib/active_job/logging.rb:44:in `block in tag_logger'
    .gem/ruby/2.5.0/gems/activesupport-5.1.5/lib/active_support/tagged_logging.rb:69:in `block in tagged'
    .gem/ruby/2.5.0/gems/activesupport-5.1.5/lib/active_support/tagged_logging.rb:26:in `tagged'
    .gem/ruby/2.5.0/gems/activesupport-5.1.5/lib/active_support/tagged_logging.rb:69:in `tagged'
    .gem/ruby/2.5.0/gems/activejob-5.1.5/lib/active_job/logging.rb:44:in `tag_logger'
    .gem/ruby/2.5.0/gems/activejob-5.1.5/lib/active_job/logging.rb:14:in `block (2 levels) in <module:Logging>'
    .gem/ruby/2.5.0/gems/activesupport-5.1.5/lib/active_support/callbacks.rb:117:in `instance_exec'
    .gem/ruby/2.5.0/gems/activesupport-5.1.5/lib/active_support/callbacks.rb:117:in `block in run_callbacks'
    .gem/ruby/2.5.0/gems/activesupport-5.1.5/lib/active_support/callbacks.rb:135:in `run_callbacks'
    .gem/ruby/2.5.0/gems/activejob-5.1.5/lib/active_job/enqueuing.rb:47:in `enqueue'
    .gem/ruby/2.5.0/gems/activejob-5.1.5/lib/active_job/enqueuing.rb:18:in `perform_later'

For the majority of cases, a client pool size of 5 is sufficient. However, servers which utilize a high number of threads, with large job payloads, and which may experience some network latency issues can see `Timeout::Error` crashes. This may be further exacerbated by the ~2-20x performance decrease through `ActiveJob` (#3782). Rails addresses this general type of connection issue for the main database by suggesting that the DB pool size match the number of threads running. This change applies that logic to the default client pool size by leveraging the same environment setting; this way there's a connection available per thread.

This may also have the side effect of a slight performance boost, as there is less of a chance that threads will be blocked waiting on connections. The trade-off is that there may be a memory profile increase to handle the additional Redis connections in the pool; note the pool only creates new connections as necessary to handle the requests.

Resolves #3806

### DIFF
--- a/lib/sidekiq/redis_connection.rb
+++ b/lib/sidekiq/redis_connection.rb
@@ -15,7 +15,15 @@ module Sidekiq
         options[:id] = "Sidekiq-#{Sidekiq.server? ? "server" : "client"}-PID-#{$$}" if !options.has_key?(:id)
         options[:url] ||= determine_redis_provider
 
-        size = options[:size] || (Sidekiq.server? ? (Sidekiq.options[:concurrency] + 5) : 5)
+        size = if options[:size]
+                 options[:size]
+               elsif Sidekiq.server?
+                 Sidekiq.options[:concurrency] + 5
+               elsif ENV['RAILS_MAX_THREADS']
+                 Integer(ENV['RAILS_MAX_THREADS'])
+               else
+                 5
+               end
 
         verify_sizing(size, Sidekiq.options[:concurrency]) if Sidekiq.server?
 
--- a/test/test_redis_connection.rb
+++ b/test/test_redis_connection.rb
@@ -20,6 +20,61 @@ class TestRedisConnection < Sidekiq::Test
       assert_equal "Sidekiq-server-PID-#{$$}", pool.checkout.connection.fetch(:id)
     end
 
+    # Readers for these ivars should be available in the next release of
+    # `connection_pool`, until then we need to reach into the internal state to
+    # verify the setting.
+    describe "size" do
+      def client_connection(*args)
+        Sidekiq.stub(:server?, nil) do
+          Sidekiq::RedisConnection.create(*args)
+        end
+      end
+
+      def server_connection(*args)
+        Sidekiq.stub(:server?, "constant") do
+          Sidekiq::RedisConnection.create(*args)
+        end
+      end
+
+      it "uses the specified custom pool size" do
+        pool = client_connection(size: 42)
+        assert_equal 42, pool.instance_eval{ @size }
+        assert_equal 42, pool.instance_eval{ @available.length }
+
+        pool = server_connection(size: 42)
+        assert_equal 42, pool.instance_eval{ @size }
+        assert_equal 42, pool.instance_eval{ @available.length }
+      end
+
+      it "defaults server pool sizes based on concurrency with padding" do
+        _expected_padding = 5
+        Sidekiq.options[:concurrency] = 6
+        pool = server_connection
+
+        assert_equal 11, pool.instance_eval{ @size }
+        assert_equal 11, pool.instance_eval{ @available.length }
+      end
+
+      it "defaults client pool sizes to 5" do
+        pool = client_connection
+
+        assert_equal 5, pool.instance_eval{ @size }
+        assert_equal 5, pool.instance_eval{ @available.length }
+      end
+
+      it "changes client pool sizes with ENV" do
+        begin
+          ENV['RAILS_MAX_THREADS'] = '9'
+          pool = client_connection
+
+          assert_equal 9, pool.instance_eval{ @size }
+          assert_equal 9, pool.instance_eval{ @available.length }
+        ensure
+          ENV.delete('RAILS_MAX_THREADS')
+        end
+      end
+    end
+
     it "disables client setname with nil id" do
       pool = Sidekiq::RedisConnection.create(:id => nil)
       assert_equal Redis, pool.checkout.class


mperham:
Really solid work.  🤘

PR #3810, opened 2018-03-31
Add link to retry job info in web UI

jackkinsella:


mperham:
The retry link is the time, first column in the row.  This is redundant.

jackkinsella:
Indeed, then it is superfluous. 

Somehow it never occurred to anyone on our team that clicking on the time would lead to more details. For whatever reason, we didn't expect it there.

If you don't have a strong preference, would you consider switching to link on the class name instead? IMO that's a bit more intuitive — in my head I'm zooming in on the details of the job class, not the time.

mperham:
It’s like that because the retry time is the unique part of the row. It seems unintuitive that the same class name would link to different places for two jobs of the same type. For instance the queue name links to the same queue page.

PR #3836, opened 2018-05-02
Stop the server complaining when shut down at the same time as Redis

chewi:
This makes a lot of noise if they are stopped simultaneously, which can happen when stopping runit, for example. You generally wouldn't do this in production but the noise is annoying in development.

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -93,6 +93,8 @@ module Sidekiq
       retryable = true
       begin
         yield conn
+      rescue Redis::BaseConnectionError
+        raise unless server? && Sidekiq::CLI.instance.launcher.stopping?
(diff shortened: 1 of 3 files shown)


mperham:
I don't see this change as relevant to 99% of Sidekiq users so I'm not likely to accept it.

If you shut things down in order, you won't see the noise, right?  Does runit have declared dependencies so if you shut down Redis, it will shut down Sidekiq first?

chewi:
Unfortunately runit's dependency handling is very basic. You can start some things before others but stopping them in the right order is a manual affair.

mperham:
Redis is extremely runtime efficient, like 5MB of RSS or similar.  I leave it running 24/7, just make sure it's only listening on localhost:6379 so you don't have any security problems.

Issue #3848, opened 2018-05-24
Redis performance issues due to `smembers('processes')` when using large number of processes


ffiller:
Ruby version: 2.5.1
Sidekiq / Pro / Enterprise version(s): sidekiq (5.1.3), sidekiq-pro (4.0.2), sidekiq-ent (1.7.1)

Hi Mike,

We're running 1k+ sidekiq processes and are seeing the redis `smembers('processes')` calls causing performance issues with our redis instance. These queries are showing between 100-500ms latency in the slowlog.

I'm assuming the guarantees that come with scan aren't sufficient for these calls?

Do you have any recommendations on what to do?

Thanks
Fried @Baremetrics

mperham:
I think it would be fine to move to a scan-based impl; it was implemented before Redis 2.8 was required so scan wasn't available.  Want to send a PR?

ffiller:
See #3849

scottrobertson:
Hey @mperham 

Just a random thought...

We are using Sidekiq Swarm, i don't suppose any of that code (i cannot see it as it aint open source) is using similar methods to keep track of processes that could be rewritten to use this more performant code?

mperham:
Generally, features use the API, so they’d all pick up your change. Swarm uses normal Unix methods only to track child processes.

PR #3849, opened 2018-05-24
decrease redis latency by switching smembers queries to sscan

ffiller:
relates to #3848

@mperham I wasn't quite sure where to put the iterator over the cursor, so for now it's just a module method in the same file. Let me know if you'd like that moved some place else.


mperham:
can you merge master and see if that fixes jruby?

lib/sidekiq/api.rb
```diff
@@ -700,6 +712,7 @@ def self.timeout
   #
   class ProcessSet
     include Enumerable
+    extend RedisIterator
```
mperham: include here also.  refactor cleanup to be an instance method.  Push back if I'm missing something.

lib/sidekiq/redis_iterator.rb
```diff
@@ -0,0 +1,16 @@
+module Sidekiq
+  module RedisIterator
```
mperham: Since it's only used within the API, put in sidekiq/api rather than its own file?  Is `RedisScanner` a better name?

test/test_redis_iterator.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
```
mperham: I would add to test_api rather than its own file.

ffiller:
@mperham Thanks for the review. Have addressed, however jruby is still failing.

mperham:
🤘

PR #3868, opened 2018-06-15
Add SemVer compatibility badge

greysteil:
First up, thanks for Sidekiq - it's brilliant.

Not sure what your opinion of badge is, so I thought I'd create a PR. This adds a SemVer compatibility badge that displays the percentage of CI runs that pass when updating `sidekiq` between SemVer compatible versions. Data comes from Dependabot (which I built it).

Here's what the badge looks like:

[SemVer](https://dependabot.com/compatibility-score.html?dependency-name=sidekiq&package-manager=bundler&version-scheme=semver)

Any feedback very welcome.


mperham:
Thanks Grey.  It's a data point but I don't see this as useful enough for everyone to justify including in the readme.

greysteil:
👍. Thanks for taking look!

PR #3877, opened 2018-07-06
fix charset attribute of Web UI meta element

h-izumi:
I have seen corrupted Web UI on Internet Explorer 11(locale: ja).

According to [MDN document](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta), `charset` must contain a [standard IANA MIME name for character encodings](https://www.iana.org/assignments/character-sets/character-sets.xhtml).

So I changed `charset`, and get correct view.

## before (`utf8`)
before

## after (`utf-8`)
after

Regards.


mperham:
Thank you!

PR #3880, opened 2018-07-09
Update ES translation

diegojromerolopez:
In Spanish "queue" can be translated as "cola" or "fila". The latter translation has a meaning of "aligning some objects" (like "row" in English).


mperham:
Gracias!

PR #3894, opened 2018-07-18
Better scheduling for large clusters, fixes #3889

### LINKED ISSUE #3889: Scheduled set not enqueueing due to large retry set

Sidekiq 4.2.10
Pro 3.5.0

I have noticed a problem where my schedule set doesn't enqueue for 10-20 minutes, usually once per hour. I have whittled it down to the fact that retry/schedule happens in the same poller, and my retry set is quite large (>100k entries). My theory currently is that the retry enqueueing is taking quite a long time to operate, which is causing the scheduled enqueueing to be delayed by that much time.

Is it possible for users of Sidekiq to separate the retry / schedule enqueueing into different actors? In this way, they could execute simultaneously rather than one after the other. I have checked 5.x vs 4.x but I do not see a difference.

I looked at the option of putting the set order to `scheduled, retry` in the Enq class, but I think that this would not help because the retry enqueueing would still take the same amount of time and execute in the same thread as scheduled.

I could also be wrong and this isn't related to the size of the retry set at all. The redis cluster isn't reacting negatively during it.

mperham:
Today we add 50% to the sleep time so that processes cluster around the target time:

50% --> target <-- 150%

This works well for small clusters, e.g. less than 10 processes.

The problem is that, for large clusters, the processes will never sleep less than 50% of (process_count * poll average) which breaks the average and delays job scheduling for several minutes.

Instead, beyond 10 processes, don't add that 50% buffer.  Allow the processes to sleep anywhere within the timespan:

0% --> target <-- 200%

With many processes, the average sleep within a 200% time period should work out close enough to 100%.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -5,6 +5,7 @@
 HEAD
 -----------
 
+- Smoother scheduling for large Sidekiq clusters [#3889]
 - Switch Sidekiq::Testing impl from alias\_method to Module#prepend, for resiliency [#3852]
 - Update Sidekiq APIs to use SCAN for scalability [#3848, ffiller]
 - Remove concurrent-ruby gem dependency [#3830]
--- a/lib/sidekiq/scheduled.rb
+++ b/lib/sidekiq/scheduled.rb
@@ -97,9 +97,34 @@ module Sidekiq
         sleep 5
       end
 
-      # Calculates a random interval that is ±50% the desired average.
       def random_poll_interval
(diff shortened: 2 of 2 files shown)


sb8244:
This makes sense to me, Mike. It doesn't necessarily guarantee the timing, but it should statistically work out. Interestingly, it might work out even better with a larger number of processes because the chance of getting a small value (cluster wide) increases with the number of servers in the cluster.

sb8244:
Thank you Mike!

PR #3896, opened 2018-07-24
Add queue size to web show page

spk:

### DIFF
--- a/web/views/queue.erb
+++ b/web/views/queue.erb
@@ -5,6 +5,7 @@
       <% if @queue.paused? %>
         <span class="label label-danger"><%= t('Paused') %></span>
       <% end %>
+      <span>(<%= number_with_delimiter(@queue.size) %>)</span>
     </h3>
   </div>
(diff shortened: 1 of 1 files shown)


mperham:
I don't like the look of it in the page title.  Maybe there's a better design but I'm not a designer so I can't offer an alternative.

thegorgon:
Sorry to pile on, but what if it was like a badge? e.g [bootstrap](https://getbootstrap.com/docs/4.1/components/badge/) or [materialize](https://materializecss.com/badges.html)

spk:
i've updated the PR and screenshot based on @thegorgon suggestion

mperham:
Nice, much better.  `@queue.size` will make a network roundtrip to Redis and be out of sync with the queue pages.  Use `@total_size` instead which suffers from neither problem.

bobber205:
I love this change!

spk:
Thanks @mperham for the fix, cheers

PR #3899, opened 2018-07-27
preserve the jid even if it contains -

ananthakumaran:
https://github.com/akira/exq uses UUID as jid which contains
`-`. The current behaviour parses the jid incorrectly, which prevents
some of functionalities in web UI like Retry Now, Delete etc from
working correctly.


lib/sidekiq/web/helpers.rb
```diff
@@ -184,7 +184,7 @@ def job_params(job, score)
     end
 
     def parse_params(params)
-      score, jid = params.split("-")
+      score, jid = params.split("-", 2).reject(&:empty?)
```
davydovanton: could you add some specs for this case?

ananthakumaran: added

mperham: Thanks for this fix, I didn't know about exq's JIDs.

What is the purpose of the `reject` call?  What case does it fix?

ananthakumaran:
```
irb(main):002:0> "123-".split("-", 2)
=> ["123", ""]
irb(main):003:0> "123-".split("-")
=> ["123"]
```
There is a test case in https://github.com/ananthakumaran/sidekiq/blob/6da9c5b3365918913242f1dcecefb47bb0bca980/test/test_web.rb#L454 which depends on this behaviour

mperham: Don't handle that case.  That's malformed input and so undefined behavior.

ananthakumaran: should I change the assertion in the failing spec then?

mperham: Did you add that spec or did I?  I think you should remove the failing spec.  post'ing "123-" is not valid.  All jobs must have a JID; the system should break otherwise.

PR #3908, opened 2018-08-01
Allow passing int value or duration to Worker#sidekiq_retry_in

DmitryTsepelev:
Now `sidekiq_retry_in` accepts a block, which can determine a delay before the next retry based on the exception and retry count. This PR contains a change allowing to pass an integer value or duration - it can be helpful for cases when we don't need dynamic value detection. For instance, instead of `sidekiq_retry_in { 5.minutes }` we can write `sidekiq_retry_in(5.minutes)`


mperham:
Retries without backoff are a bad idea.  I don't want the API to make this any easier.

DmitryTsepelev:
Got it, thanks for the review

PR #3913, opened 2018-08-03
Add CSP header to web admin to ensure it matches needs

thegorgon:
Given that the Sidekiq admin does not define it's own Content-Security-Policy, if a parent application is using a CSP, e.g. in Rails 5.2, the parent policy will be used for the admin pages. This may cause problems with the admin pages, depending on the application policy.

### DIFF
--- a/lib/sidekiq/web/application.rb
+++ b/lib/sidekiq/web/application.rb
@@ -7,6 +7,21 @@ module Sidekiq
     CONTENT_LENGTH = "Content-Length"
     CONTENT_TYPE = "Content-Type"
     REDIS_KEYS = %w(redis_version uptime_in_days connected_clients used_memory_human used_memory_peak_human)
+    CSP_HEADER = [
+      "default-src 'self' https: http:",
+      "child-src 'self'",
+      "connect-src 'self' https: http: wss: ws:",
+      "font-src 'self' https: http:",
+      "frame-src 'self'",
+      "img-src 'self' https: http: data:",
+      "manifest-src 'self'",
+      "media-src 'self'",
+      "object-src 'none'",
+      "script-src 'self' https: http:",
+      "style-src 'self' https: http: 'unsafe-inline'",
+      "worker-src 'self'",
+      "base-uri 'self'"
+    ].join('; ').freeze
 
     def initialize(klass)
       @klass = klass
@@ -279,6 +294,7 @@ module Sidekiq
           "Content-Type" => "text/html",
           "Cache-Control" => "no-cache",
           "Content-Language" => action.locale,
+          "Content-Security-Policy" => CSP_HEADER
         }
 
         [200, headers, [resp]]
--- a/test/test_web.rb
+++ b/test/test_web.rb
@@ -54,6 +54,15 @@ class TestWeb < Sidekiq::Test
       assert_match(/Oversikt/, last_response.body)
     end
 
+    it 'can provide a default, appropriate CSP for its content' do
+      get '/', {}
+      policies = last_response.headers["Content-Security-Policy"].split('; ')
+      assert_includes(policies, "connect-src 'self' https: http: wss: ws:")
+      assert_includes(policies, "style-src 'self' https: http: 'unsafe-inline'")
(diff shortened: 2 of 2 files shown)


lib/sidekiq/web/application.rb
```diff
@@ -291,6 +292,26 @@ def call(env)
       resp
     end
 
+    def get_csp_header(scheme)
```
mperham: This method has two possible inputs, http and https, but nothing else dynamic so it only has two possible outputs. What do you think about memoizing the two constructed policies to minimize the runtime overhead?

What should it do in the case of a non-http scheme?  Is that even possible?

thegorgon: memoizin: can do.

in case of non-http: it should probably not include the header. Will update for that case.

thegorgon:
Updated to memoize and ignore the header on non https? requests. 

One concern - if the assets for the admin page are hosted outside the main app (e.g. on S3) this CSP will be problematic. Is that a common use case? If so, there are a few ways to update the policy to account for that case.

mperham:
There are 3rd party Sidekiq web extensions that might use external resources, eg sidekiq-cron.

lib/sidekiq/web/application.rb
```diff
@@ -291,6 +294,30 @@ def call(env)
...
+      @csp_headers ||= {}
+      @csp_headers[scheme] ||= begin
+        return nil unless %w[https http].include?(scheme)
```
mperham: A nil value will not memoize.  I'd suggest you use "" and change your guard to `if csp_header != ""`

lib/sidekiq/web/application.rb
```diff
@@ -291,6 +294,30 @@ def call(env)
       resp
     end
 
+    def get_csp_header(scheme)
+      @csp_headers ||= {}
```
mperham: I'd make this hash a constant; I'm not sure of the lifetime of instance variables here.

mperham:
So... I did a search on GitHub for all Sidekiq extensions that use `add_to_head` to add asset references and it turns out that the only extension that does so is... Sidekiq Enterprise.  Me. 😎

It adds this:

```
<link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css">
```

How does the CSP affect this?

mperham:
It does so to add this question mark icon... which is objectively ridiculous.  Adding an entire asset for one character on one page.  I will find a better solution that does not require an external asset.

concurrent

thegorgon:
Ok - rethought a bit and updated:

• The response is always HTML (see header here: https://github.com/mperham/sidekiq/pull/3913/files#diff-150202094cea57cb2b6de0dbddd335bdR297), so including a CSP is always valid. 
• being restrictive to only https/wss for https/wss requests is a nice security addition, but it's not really necessary. Simplify to a single CSP.
• Add the boostrapcdn external source as a valid stylesheet source.

My only remaining concern is that consuming applications compile and deploy the assets for sidekiq admin into a CDN (e.g. using Sprockets). It doesn't look to me from a cursory glance like this is possible, given the asset paths are all derived from request env parameters, but I may be missing something.

In addition to a few other things (blocking eval, blocking flash, etc.) This policy only allows styles and scripts to be loaded from the same domain as the page is loaded from (and the boostrap cdn for the enterprise usecase). Any other scripts/styles will be blocked from use. If this could be problematic, I can envision 2 options for moving forward.

1. We provide a config to change the CSP header for Sidekiq admin pages. 
2. We don't provide any CSP in the admin pages and instead leave it to application owners to setup their own CSP (e.g. with middleware).

mperham:
It looks like this provides a default CSP, and the user could install a Rack middleware that overrides the CSP.  Does that sound right?  I’ve updated Sidekiq enterprise to not use that external resource anymore so you do not need to special case it.

thegorgon:
Fair point!

Removed the cdn exception, will merge on a green build.

mperham:
Thank you, nice followup.

thegorgon:
Thank you for Sidekiq!

Issue #3905, opened 2018-07-30
No TTL on stats keys in Redis


odarriba:
Ruby version: 2.4.0
Sidekiq / Pro / Enterprise version(s): Sidekiq 4.2.10

We are seeing lots of stats keys in our Redis isntance, some of them from more than six months ago. I have been checking and I cannot see any code in which stats are removed (they had TTL before, in older versions, but not anymore).

Is there any way to set TTL safely for this? We thought on putting a default expiry on Redis connection settings but we are not sure about if this is safe to use with Sidekiq :thinking: 

Thanks!

mperham:
See changelog, v4.1.2 and 4.1.3.

odarriba:
Hi!

That changelogs are related to memory leaks related to worker keys, the one we are seeing are like this:

```
...
236) "my_app:stat:processed:2017-09-20"
237) "my_app:stat:processed:2018-06-24"
238) "my_app:stat:failed:2017-11-10"
239) "my_app:stat:failed:2018-02-15"
240) "my_app:stat:processed:2017-11-11"
241) "my_app:stat:processed:2018-03-24"
242) "my_app:stat:processed:2018-06-25"
243) "my_app:stat:failed:2018-01-23"
244) "my_app:stat:failed:2018-04-23"
245) "my_app:stat:processed:2017-12-04"
246) "my_app:stat:failed:2018-03-12"
247) "my_app:stat:failed:2018-05-26"
248) "my_app:stat:failed:2017-10-29"
249) "my_app:stat:processed:2017-12-19"
250) "my_app:stat:processed:2018-05-03"
251) "my_app:stat:failed:2018-07-08"
....
```

And if we check the TTL:

```
127.0.0.1:6379> TTL "my_app:stat:failed:2017-12-07"
(integer) -1
```

mperham:
You're right, I was mistaken.  Looks like Sidekiq does not expire them.

Under normal circumstances, Sidekiq would put a 5 year TTL on them -- they are used to draw the history graph.  That's long enough that effectively they wouldn't have a TTL.  It currently "leaks" 2 keys and < 100 bytes per day, that's 730 keys and ~35KB per year.

odarriba:
Yep, not a big issue, but just wanted to let you know because I know there was a previous change to set it to 5 years and it was lost in a later change. Thanks!

mperham:
[...]

PR #3916, opened 2018-08-06
Add TTL for stats keys

odarriba:
This change adds a 5 years TTL on redis keys containing daily stats. See #3905 for more details


mperham:
Thank you!

PR #3928, opened 2018-08-16
Fix bug where latency calculation would fail converting a string to a float.

undergroundwebdesigns:
In our production instance of Sidekiq we periodically run into an error where the `queues` page of the sidekiq UI failed to load due to an error casting a string to a float. The stack trace points to line 263 of api.rb.

This PR simply makes sure that the `enqueued_at` time pulled from sidekiq is a float before trying to subtract it from `time` to avoid this error.

This may be better fixed by figuring out how `enqueued_at` is being set to a string in the first place and fixing that.

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -259,7 +259,7 @@ module Sidekiq
       return 0 unless entry
       job = Sidekiq.load_json(entry)
       now = Time.now.to_f
-      thence = job['enqueued_at'] || now
+      thence = job['enqueued_at'].to_f || now
       now - thence
     end
 
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -259,7 +259,7 @@ module Sidekiq
       return 0 unless entry
       job = Sidekiq.load_json(entry)
       now = Time.now.to_f
-      thence = job['enqueued_at'].to_f || now
+      thence = job['enqueued_at'] ? job['enqueued_at'].to_f : now
       now - thence
     end
 


mperham:
That breaks the || operator on the line.

mperham:
The issue here is that enqueued_at should always be a Float.  The question is: should Sidekiq fail fast on bad data so the issue can be detected and fixed or should it silently work around the issue?

The job payload is absolutely fundamental to the entire Sidekiq ecosystem.  The data in it must conform for things to work properly so I'm inclined to want this to explode instead of silently fixing the issue. WDYT?

mperham:
Hmm, I wonder if this is a bug in Sidekiq Pro's reliable scheduler which has a pretty gross hack in Lua in this area.  Let me investigate.

mperham:
By any chance do you have the reliable scheduler enabled?

undergroundwebdesigns:
Yes we do have the reliable scheduler enabled.

mperham:
Ok, it looks like if a client library injects enqueued_at as a String into the job payload, the reliable scheduler will ignore the value and not update it.  This will pass thru the bad data -- reliable scheduler doesn't fix the problem but it doesn't make it worse.

Are you pushing jobs to redis via any other mechanism aside from the Sidekiq::Client and Sidekiq::Worker APIs?  If so, can you share the code?

undergroundwebdesigns:
We're using sidekiq pro 4.0.2, and redis 3.2.12.

I'm not aware of anything in our code that pushes sidekiq jobs outside the normal API's, though we are using the [Karfka Sidekiq backend](https://github.com/karafka/sidekiq-backend), version 1.2.0, perhaps that's related?

mperham:
I'm going to pass on this because I want the malformed job to fail fast.  The Job Format wiki page shows enqueued_at as an epoch float and that is the only form accepted today.

https://github.com/mperham/sidekiq/wiki/Job-Format

PR #3936, opened 2018-08-24
Raise error for duplicate queue names in config, fixes #3911

### LINKED ISSUE #3911: Duplicate Queue in Config Switches from Strict to Random Weighted Fetch
Sidekiq / Pro / Enterprise version(s): sidekiq-5.1.3, sidekiq-pro-4.0.3

I noticed that if mistakenly duplicate a queue name in my config, `SuperFetch` will switch from strict ordering to random weighted because the `@queues.length == @queues.uniq.length` check fails.  Would it be possible to raise an `ArgumentError` for this misconfiguration instead?

chadrschroeder:
Fixes #3911.


mperham:
Do you think this is worthy of a change log entry?

mperham:
👍

Issue #3958, opened 2018-09-11
Jobs lingering in busy despite having completed (on jruby)


mattbooks:
Ruby version: jruby 9.1.16.0
Sidekiq / Pro / Enterprise version(s): 5.2.1 / pro 3.4.5

Please include your initializer and any error message with the full backtrace.
- It's hard for me to get that, and there are no exceptions.

Are you using an old version? no
Have you checked the changelogs to see if your issue has been fixed in a later version? Yes. A different concurrency issue with `WORKER_STATE` was fixed recently, however: #3921

**TL;DR: I believe this is a thread safety issue with `WORKER_STATE` on jruby**

Here's why I think that: 

I can see jobs linger in the busy tab in the web UI. I believe they are finished executing because I can see the logs from them completing.

I checked if the data powering the admin page was stale, but it appears as though the heartbeat in `sidekiq/launcher` is refreshing it, as I can see the expiry being refreshed for `hget("<proc>:workers")`.

Since that reads from the global `WORKER_STATE` variable, which is annotated with 

```
# This is mutable global state but because each thread is storing
# its own unique key/value, there's no thread-safety issue AFAIK.
```

I suspected this might be a thread safety issue. I also know that MRI apps at my company are not reporting this issue, so perhaps it is an issue with the jruby implementation, which is not threadsafe AFAIK (I think you need to use `Concurrent::Hash`)

To test this theory, I wrote this script:

```
require 'concurrent/hash'

GLOBAL_HASH = Concurrent::Hash.new
# GLOBAL_HASH = {}

do_modifications = Proc.new do |name|
  Thread.new do
    loop do
      begin
        k = "#{name}:#{rand.to_s}"
        GLOBAL_HASH[k] = name
        GLOBAL_HASH.delete(k)
      rescue => e
        puts "caught error in #{name}"
        puts e
        raise
      end
    end
  end
end

do_modifications.call('foo')
do_modifications.call('bar')

puts "Running!"
1000000.times do
  # puts GLOBAL_HASH
  bad_hash = GLOBAL_HASH.dup
  raise "Too many keys: #{bad_hash}" if bad_hash.size > 2
end
```

(run with `while true; do; jruby test.rb; done` since it doesn't fail every time)

with `{}` it raises, with `Concurrent::Hash`, it does not. I have not seen this fail on MRI either.

Indeed, testing this in our staging environment using `Concurrent::Hash` for `WORKER_STATE` seems to resolve the issue.

[...]

PR #3959, opened 2018-09-11
Use a mutex to synchronize access to shared WORKER_STATE hash

mattbooks:
On jruby, the Hash implementation is not threadsafe, and leads to jobs not being
deleted from the busy list.

See #3958

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -6,6 +6,7 @@ HEAD
 ---------
 
 - Raise error for duplicate queue names in config to avoid unexpected fetch algorithm change [#3911]
+- Wrap WORKER_STATE in a mutex since Hash is not threadsafe on jruby [#3958]
 
 5.2.1
 -----------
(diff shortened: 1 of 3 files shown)


mperham:
Need to delete or memory leak!

mperham:
Also, there's a read loop in launcher.rb that uses `dup`, need to impl that I think?

mattbooks:
Yes, and a couple other that I missed (`size` and `clear`)

mattbooks:
> Need to delete or memory leak!

I'm not sure what this is referring to

mperham:
You have a delete method on your SharedWorkerState but don't use it anywhere.  Where are the hash elements removed?

mattbooks:
Oh, that is the same method as the hash was using, so it doesn't appear in the diff. It's still called from the ensure on L248 of `processor.rb`

mattbooks:
Anything else for this? Did you want to try to add the test for the race? Given how long it took my script (in the issue) to repro, I am skeptical it'll work as a test, but I could try to figure something out.

mperham:
Good enough.  The code looks clean and if you say this fixes your problem, that's enough of a test for me.

PR #3970, opened 2018-09-17
Breaking changes

mperham:
Unclear when or how we will roll these out.  Perhaps we target this for Sidekiq 6.0 but they seem like small code changes for a major version bump.  OTOH I want to be respectful and minimize upgrade pain.


seuros:
I don't think these changes requires a major breaking version bump.

mperham:
Here's my suggested rollout plan:

* Add a check and warning message if REDIS_PROVIDER contains a URL.  Release with 5.2.3 in October.
* Change to an explicit error (as in this PR) in 5.3.0, released 1Q2019.

mperham:
Should 5.2.3 print out a warning if the shutdown timeout is 8 seconds?

lib/sidekiq/redis_connection.rb
```diff
@@ -115,27 +115,20 @@ def determine_redis_provider
         # REDIS_PROVIDER=MY_REDIS_URL
         # and Sidekiq will find your custom URL variable with no custom
         # initialization code at all.
+        #
         p = ENV['REDIS_PROVIDER']
         if p && p =~ /\:/
```
Tensho: No need for `p` existence check:
```
nil =~ /\:/ #=> nil
```

test/test_redis_connection.rb
```diff
@@ -223,6 +223,18 @@ def with_env_var(var, uri, skip_provider=false)
     end
 
     describe "with REDIS_PROVIDER set" do
+      it "rejects URLs in REDIS_PROVIDER" do
+        uri = 'redis://sidekiq-redis-provider:6379/0'
+
+        ENV['REDIS_PROVIDER'] = uri
+
+        assert_raises RuntimeError do
+          Sidekiq::RedisConnection.__send__(:determine_redis_provider)
```
Tensho: Why don't you want to use the same cool technique with `private unless $TESTING`?

PR #3971, opened 2018-09-18
Namespace ENV vars for sidekiq

seuros:
- Deprecate REDIS_PROVIDER in favor of SIDEKIQ_REDIS_PROVIDER
- Added SIDEKIQ_REDIS_URL to specify redis url for sidekiq


mperham:
Should we have warning messages if using an old variable name?

seuros:
I don't think so, because REDIS_URL is the default variable for Redis gem and can it be used for other connections (Rails cache for example)
I think we should encourage the usage of SIDEKIQ_REDIS_URL. 

Also what do you think about using SIDEKIQ_REDIS_ENVNAME  instead ? PROVIDER is confusing.

mperham:
After thinking about it, I'm :-1: on the SIDEKIQ_ prefix.  I think it's unnecessary change to put on the user.

seuros:
Maybe we could have a poll wdyt ? 
Here is my use case : In a microservice architecture, it is easier to know where variables are used without having to dig into documentation/code. 

Just to make it clear
I deprecated REDIS_PROVIDER only.  
REDIS_URL will always be a fallback to SIDEKIQ_REDIS_URL.

in Rails , if you set activesupport cache as :redis, and don't configure it. Both Sidekiq and Rails cache will start using REDIS_URL.

mperham:
The REDIS_PROVIDER variable tells a process where to find its Redis URL. The same variable can be used by Sidekiq, Resque, the Redis cache store, user code, etc. There's nothing Sidekiq-specific about it. I don't want to ask the user to add a prefix without clear benefit.

PR #3973, opened 2018-09-19
Fixed mobile navigation.

navied:
Fixed up css, and removed duplicate nav element.

### DIFF
--- a/web/views/_nav.erb
+++ b/web/views/_nav.erb
@@ -32,12 +32,6 @@
           <% end %>
         <% end %>
 
-        <li class="dropdown" data-navbar="dropdown">
-          <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false">
-            <%= t('Plugins') %> <span class="caret"></span>
-          </a>
-        </li>
-
         <% Sidekiq::Web.custom_tabs.each do |title, url| %>
           <li class="<%= current_path.start_with?(url) ? 'active' : '' %>" data-navbar="custom-tab">
             <a href="<%= root_path %><%= url %>"><%= t(title) %></a>
--- a/web/views/_nav.erb
+++ b/web/views/_nav.erb
@@ -37,9 +37,8 @@
             <a href="<%= root_path %><%= url %>"><%= t(title) %></a>
           </li>
         <% end %>
-      </ul>
-      <ul class="nav navbar-nav navbar-right navbar-livereload" data-navbar="static">
-        <li>
+
+        <li class="navbar-livereload">
           <div class="poll-wrapper">
             <%= erb :_poll_link %>
             <% if Sidekiq::Web.app_url %>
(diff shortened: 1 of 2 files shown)


web/views/_nav.erb
```diff
@@ -36,13 +36,6 @@
           <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false">
```
mperham: If you remove the dropdown-menu, why keep the dropdown-toggle?

This dropdown was added so that smaller desktops (like iPads) wouldn't see the tabs overflow the right and break onto a second line of tabs.  This doesn't happen when you have "Sidekiq Pro" and only one extra tab "Batches", but it does happen when you have "Sidekiq Enterprise" and three extra tabs "Batches", "Limits" and "Cron".  Unfortunately it's tough to get a design that works well with all three packages.

navied: Ah true, you are right. I was't 100% why that was there but that makes sense. I went and removed it. I think seeing the mobile nav on iPads should be okay.

mperham:
There's still a bunch of little CSS issues around the various width breakpoints.  I'm not really happy with the current state, this improves it but there's 3-4 other issues to be fixed.  If you shrink and enlarge the page, you'll likely notice a few, e.g. the processing stats when very thin:

```diff
--- a/web/assets/stylesheets/application.css
+++ b/web/assets/stylesheets/application.css
@@ -126,7 +126,7 @@ header.row .pagination {
   text-align: center;
   width: 14%;
 }
-@media (max-width: 767px) and (min-width: 400px) {
+@media (max-width: 767px) and (min-width: 200px) {
   .summary_bar ul li {
     width: 100%;
   }
```

navied:
Added the processing stats change you mentioned, and also hid the live poll on for iPad portrait mode to prevent the Nav from going into two rows. The only other issues I've seen is scheduled jobs tab, and busy tab has a horizontal scroll on mobile but the solution would require a bit more reworking. 

What were the other changes you were referring to?

mperham:
top navbar breaks here into two lines.  Should this stay on one line and overflow the viewport?
screen shot 2018-09-20 at 9 44 39 am

mperham:
That's really the only pressing issue I see.  I don't like the responsive breakpoints, I'd rather the UI smoothly change horizontal width rather than use discrete points but I don't know the state of the art in CSS.  The UI was built 4-5 years ago when that was best practice.

mperham:
Almost perfect.  Looks like the combo of the two nav buttons breaks.

screen shot 2018-09-20 at 12 35 34 pm

navied:
Fixed, I have never seen that button before show up haha.

mperham:
In routes.rb, do this:

```ruby
require 'sidekiq/web'
Sidekiq::Web.app_url = '/'
```

navied:
Ahh, that's a nifty feature

mperham:
Looks good to me in latest Safari, FF and Chrome.  I can't easily test on mobile.

mperham:
@navied Thank you so much for your follow up and thoroughness.

PR #3980, opened 2018-09-26
Ensure death_handlers are called on jobs with retry: false

Benjamin-Dobell:
Prior to this commit `death_handlers` are only called if a job's retries are exhausted. Consequently, there was different behavior between specifying `retry: false` or `retry: 0` in `sidekiq_options`. 

This PR corrects the issue by ensuring the _global_ handler calls the `death_handlers`. If a job has a non-`false` value for `retry`, then the existing behavior is unaffected i.e. the local handler delegates to `attempt_retry` and when it's finished raises `Sidekiq::RetryJob::Skip`, which informs `global` not to handle the error, thus `death_handlers` are only ever called once.

### DIFF
--- a/lib/sidekiq/job_retry.rb
+++ b/lib/sidekiq/job_retry.rb
@@ -80,8 +80,18 @@ module Sidekiq
       # ignore, will be pushed back onto queue during hard_shutdown
       raise Sidekiq::Shutdown if exception_caused_by_shutdown?(e)
 
-      raise e unless msg['retry']
-      attempt_retry(nil, msg, queue, e)
+      if msg['retry']
+        attempt_retry(nil, msg, queue, e)
+      else
+        Sidekiq.death_handlers.each do |handler|
+          begin
+            handler.call(msg, e)
+          rescue => handler_ex
+            handle_exception(handler_ex, { context: "Error calling death handler", job: msg })
+          end
+        end
+      end
+
       raise e
     end
 


mhenrixon:
Ah man this is awesome! 👍 Thanks for taking the time to do this, I've been absolutely swamped :)

mperham:
`retry: false` is defined as "the job is completely ephemeral", it will disappear if it fails.  The job does not go to the Dead set, therefore it does not "die".  I think this breaks that semantic.  Convince me otherwise?

mperham:
If you want a job to not retry but die, you want `retry: 0`.

mhenrixon:
One reason for unique jobs is to be able to clean up locks for jobs that failed for whatever reason. Right now using death handlers gets most of the locks unless of course someone specified `retry: false` when I have no chance what so ever to remove the lock for this job.

Benjamin-Dobell:
> The job does not go to the Dead set, therefore it does not "die".

Being your software, the semantics of what "dead" means in the context of Sidekiq are obviously at your discretion.

However, I think the definition that you're using (i.e. "ran out of retries") is somewhat counter-intuitive, or at least non-obvious. Processing "death" is a relatively well understood metaphor in computing, and it does not mean "ran out of retries". Things that don't even have a concept of "retry" can "die" or be "killed" e.g. a process.

As such, I would say just because we don't want to keep a **record** of the fact the worker died, doesn't mean that the worker never actually died.

However...

> I think this breaks that semantic.

However, this is a very good point, at least in the context of Sidekiq. If it is indeed well established that, that "dead" means "out of retries", then changing that definition now is probably a very bad idea.

So, that leaves us with the point @mhenrixon has raised.

It's often important to know whether or not a job _failed_. So I'm happy to throw away this PR if we can come up with another solution. Perhaps we need `failure_handlers`? Although introducing yet another type of handler may be overkill.

I am aware that there are generic 'exception_handlers', however this exception is called even when jobs _will_ be retried.

mperham:
@mhenrixon AFAIK you can't use locks safely in a distributed system without having a timeout/expiry.  You'll get the same behavior if Sidekiq segfaults.  This is why Sidekiq Enterprise's unique lock doesn't offer the option to remove the lock on failure, only success.

@Benjamin-Dobell I think your logic is reasonable -- sending the job to the death_handlers doesn't mean it will be persisted in the Dead set but more of a notification that this job will never retry.

Benjamin-Dobell:
A little bit off topic, but I'm intrigued...

>  This is why Sidekiq Enterprise's unique lock doesn't offer the option to remove the lock on failure

I'm far from an expert in this area, however would I be correct in saying you _can_ verify failure happened, _can_ verify success happened, but _can't_ verify that success _will not_ happen?

> You'll get the same behavior if Sidekiq segfaults

If Sidekiq were to segfault, as long as the lock is _recorded_ somewhere (e.g. ledger, much like DB transactions), then upon start-up Sidekiq can see the lock is held and _possibly_ verify whoever held that lock died.

In a distributed system verifying "someone" else died is not guaranteed to be possible, in finite time. So this is where the timeouts come into play. Thus if a timeout is reached, you can't safely release the lock.

However, seems as the "common case" is verified success or verified failure, shouldn't Sidekiq Enterprise (or some other implementation) ought to be able to safely release the lock?

Although, I do suppose it matters what the lock signifies.

If you're doing something non-idempotent in your worker, you probably don't want your worker being free to run again. However, if you're just using your lock to prevent _parallel_ jobs; race conditions, wasted computing resource etc. (my use case) then having the lock freed is ideal.

mhenrixon:
@mperham the only way I offer to remove a lock magically is with a death handler. Unique jobs don't remove any lock when Sidekiq segfaults to some users immense frustration. I keep telling people that if they restart their workers and a job exits prematurely, they will have to clean up that lock manually. I get fewer complaints now than I did when the gem allowed duplicates though so we are moving in the right direction.

The whole point of unique jobs is to ensure uniqueness. Most what I do is also Lua based, replacing existing jobs in either retry, schedule or on queue, removing jobs by unique digest, etc.

So what I want to achieve is for people to stop complaining that I protect them from duplicate jobs while not sacrificing reliability too much. Using the death handlers even with `retry: false` would be a very convenient way of doing this.

That said I am working on a worker options validator, to be able to communicate incompatible settings somehow. There are other settings in the gem that are incompatible. Might add validation for `retry: false` and the use of a death handler.

mperham:
Abstract design issues aside, I think this is a reasonable tweak to semantics.  It's reassuring that ANY job which does not finish successfully under normal circumstances will go through these handlers. Thanks for convincing me! I'll put a note in the changelog about the tweak.

PR #4026, opened 2018-11-24
Cleanup CLI

Tensho:
This is the first part of the CLI cleanup, particularly here I try to actualize them. I'm going to add more tests for the CLI, especially for `run` phase (it's almost absent).

There are 3 major sections, which specify corresponding phases:
- `parse` (options)
- `run`
- `signal handling`
Also, I changed fixture config files name to start with `config`:
- `config.yml`
- `config_string.yml`
- `config_empty.yml`
- `config_environment.yml`

### Refactoring

- Move `demonize` and `write_pid` from `parse` to `run` phase
- Move `queues` `'default'` setting from `validate!` to `setup_options` phase

### Improvements
- Add `pry-buybug` gem to `Gemfile`. There is rescued `require` for it, so why not to put it into the bundle?


test/test_cli.rb
```diff
@@ -1,441 +1,357 @@
 # frozen_string_literal: true
+
 require_relative 'helper'
 require 'sidekiq/cli'
+require 'sidekiq/launcher'
```
Tensho: Maybe it's worth to require launcher at the top of `cli.rb`, but right now you get fail due to missing constant if you run `ruby test/test_cli.rb` solely.

test/test_cli.rb
```diff
@@ -1,441 +1,357 @@
...
 require 'tempfile'
 
-class Sidekiq::CLI
```
Tensho: I guess it's ugly to patch the class for testing purposes this way. But I really liked the technique [here](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/launcher.rb#L63):
```
private unless $TESTING
```
^_^

test/test_cli.rb
```diff
@@ -1,441 +1,357 @@
...
-
-  def valid?
-    !@code
-  end
-end
-
-class TestCli < Sidekiq::Test
-  describe 'CLI#parse' do
-
+class TestCLI < Sidekiq::Test
```
Tensho: @mperham Why have you introduced `Sidekiq::Test`? It basically aliases `Minitest::Test`...

test/test_cli.rb
```diff
@@ -1,441 +1,357 @@
...
+  describe '#parse' do
     before do
       @cli = Sidekiq::CLI.new
       @opts = Sidekiq.options.dup
+      @logger = Sidekiq.logger
```
Tensho: I'm still thinking how to reset global state for each test more efficiently.
This is the first time I work with `minitest` and I still have to find hooks similar to `rspec`.

test/test_cli.rb
```diff
@@ -1,441 +1,357 @@
...
-          @cli.__send__ :parse_queue, opts, 'once', nil
-          assert_equal %w[once], opts[:queues]
-          assert opts[:strict]
-        end
-      end
+      refute($LOADED_FEATURES.any? { |x| x =~ /fake_env/ })
+    end
+
+    it 'does not boot rails' do
+      refute defined?(::Rails::Application)
```
Tensho: I've added a blank line between any test phase – `setup`, `action`, `verify`, `teardown`.

test/test_cli.rb
```diff
@@ -1,441 +1,357 @@
...
+
+      @cli.parse(%w[sidekiq -r ./myapp])
+
+      refute defined?(::Rails::Application)
     end
   end
 
-  describe 'misc' do
+  describe '#run' do
```
Tensho: I've left a couple of skipped (TODO) tests to not forget about them.

test/test_cli.rb
```diff
@@ -1,441 +1,357 @@
 # frozen_string_literal: true
```
Tensho: Generally I considered `test_cli.rb` mostly like integration or system test, not unit one. That's why I prefer to not use stubs/mocks here to be confident in the future `cli.rb` refactoring.

test/test_cli.rb
mperham: I believe cli should require launcher as it uses that class directly

test/test_cli.rb
mperham: I don’t recall where that class came from. Agreed, happy to see it removed.

Tensho:
@mperham Thank you for review. I’m on vacation atm. Gonna apply your comments in a week.

Tensho:
@mperham Addressed all discussed concerns I think. Let me know if there's anything else you can think of.

test/test_cli.rb
mperham: were you going to supply these tests or are these TODO for someone else?

test/test_cli.rb
Tensho: I'm definitely going to cover `#run` with tests ^_^ checkout the code precisely at the moment and think how to do it better. I guess it's possible to test some parts for non-daemonized mode with corresponding refactoring and mocking. Do you have the suggestions?

test/test_cli.rb
Tensho: I'm definitely going to cover `#run` with tests ^_^ checkout the code precisely at the moment and think how to do it better. I guess it's possible to test some parts for non-daemonized mode with corresponding refactoring and mocking. Do you have the suggestions?

mperham: No suggestions.  If you aren't going to implement something yourself, please remove the TODOs.

Tensho: TBH, I don't see any drawbacks to specify some parts of the lib require the testing. But I don't mins to remove them in this PR and will add in the next one related to `#run` tests.

Tensho: Done

PR #4028, opened 2018-11-27
Add TID to procname

nikolai-b:
I've run into issues where some jobs were locking up my database using https://wiki.postgresql.org/wiki/Lock_Monitoring it was possible to print out the application name's that were holding the locks.  These were inevitably `sidekiq version app_name [n of m busy]`.  It was possible to trace them down using by looking at the IP and port that was connected but it would be much easier if each process had the TID as part of the name. 

I've put the TID after the version and in `[]` as a start.  Is this something you would consider?  Would you prefer if it was optional, e.g. `sidekiq --add-tid-name`?

Regards

### DIFF
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -18,6 +18,7 @@ module Sidekiq
     PROCTITLES = [
       proc { 'sidekiq' },
       proc { Sidekiq::VERSION },
+      proc { |me, data| "[#{Sidekiq::Logging.tid}]" },
       proc { |me, data| data['tag'] },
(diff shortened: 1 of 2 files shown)


mperham:
This makes no sense.  TID is Thread ID, a Sidekiq process has many TIDs.  You cannot treat it like a global.

mperham:
Perhaps you want $$ (the PID) in the name?

mperham:
I’m coming down against it. I don’t know of any other process that adds its own PID in the process name. I think instead you want to set a postgres connection string explicitly with whatever debug data you might need.

Issue #4030, opened 2018-11-28
Web UI not showing rate limiting minute bucket history?


geoffharcourt:
Ruby version: 2.5.3
Sidekiq 5.2.3 / Pro 4.0.4 / Ent 1.7.1

https://github.com/mperham/sidekiq/wiki/Ent-Rate-Limiting#web-ui

We're using a number of Bucket limiters (minute-sized buckets) and they appear to be working properly, but we never see history in the Web UI. Is the history limited to `:second` types in the same way that the retries are for buckets?

![screenshot 2018-11-28 11 54 41](https://user-images.githubusercontent.com/319471/49167934-75687980-f304-11e8-87bc-86d506060235.png)

mperham:
You should have 60 data points.  Is it possible they are all 0 because the limiter hasn't been used for the last hour?  Is it possible your browser is blocking the JS for some reason?

```
      # Note that we store all bucket data for one time
      # period larger than the interval.  If this is a
      # :second bucket, we'll store 60 buckets of data for
      # a minute of data.  If this is an :hour bucket, we'll
      # store 24 buckets for a day of data.
```

You can use `Sidekiq::Limiter::Status.new(key).history` in IRB to get the historic data for the limiter, where key is the limiter's name (`lmtr-...`) in the URL bar.

geoffharcourt:
It looks like this was CSP blocking some JS. Thanks for the explanation and prompt response!

PR #4031, opened 2018-11-28
Add unsafe-inline CSP to allow bucket graph to display

geoffharcourt:
The current CSP policy for Sidekiq Web appears to block the inline
script that renders the history charts for bucket limits.

This change updates the script CSP rules to match the style CSP rules to
allow the chart to successfully render. In our application this change
made the bucket charts load successfully.

cc: #3913, #4030 

CSP warnings before fix:
screenshot 2018-11-28 14 27 38


geoffharcourt:
I realize this might be bad for security, so if doing this via nonce is preferable I'm happy to try tackling it that way.

PR #4038, opened 2018-12-01
Code style changes. Converted hashes to ruby 1.9 style.

AKovtunov:
Hey guys, ruby 1.9 gave us new hash styles so I thought it would be nice to contribute to this project by updating the code style.
I already updated some of the wiki articles, now updated error messages, examples, tests and codebase.
If this PR is not representing your interests (maybe will brake ruby < 1.9 support) - feel free just to close it :smiley:

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,4 +5,4 @@ Rake::TestTask.new(:test) do |test|
   test.pattern = 'test/**/test_*.rb'
 end
 
-task :default => :test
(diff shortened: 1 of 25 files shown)


mperham:
Please do not make mass changes like this without asking first. I use the old style for a reason: symbols are not a legal JSON type so they should be avoided for anything that goes in the job payload.

PR #4039, opened 2018-12-05
Add CLI Run Test - Write Pid, System Boot, Print Banner

Tensho:
I tried to cover as much as I can in `#run` method with these changes. I'm still thinking how to test daemonized version...

### Refactgoring

- Extract `Sidekiq::CLI#launch` method to detach actual launcher start from the system preparations (write pid, boot system, print banner) to test in isolation
- Cleanup test `helper`

### Improvements

- Add ultra-simple Rails application to `spec/dummy` for the `Sidekiq::Rails` engine integration test
- Add [`appraisal`](https://github.com/thoughtbot/appraisal) to test `sidekiq` integration with different `rails` versions (ATM 4 and 5). Appraisals may be used to test either different versions of the `sidekiq` dependencies in the future.
-  Add [`minitest-focus`](https://github.com/seattlerb/minitest-focus)
- Add [`minitest-reporters`](https://github.com/kern/minitest-reporters) for convenient work with the tests in RubyMine
- Group gems in bundle

### Related PRs

- #4026


mperham:
Don't got overboard on test coverage.  The code doesn't change often and it's far easier to test manually once than to construct and maintain some really complicated test harness.

Tensho:
I just try to understand the code that way, but generally I agree ^_^

Tensho:
@mperham Are you comfortable with ["appraisals"](https://github.com/thoughtbot/appraisal) presence?

Tensho:
@mperham Please let me know if you are fine with the current changes and commits history 🙇

lib/sidekiq/cli.rb
```diff
@@ -251,8 +249,6 @@ def options
     def boot_system
       ENV['RACK_ENV'] = ENV['RAILS_ENV'] = environment
 
-      raise ArgumentError, "#{options[:require]} does not exist" unless File.exist?(options[:require])
```
Tensho: I removed this check because we already have it in a proper place – validation. The only possible way to get `ArgumentError` exception – call `run` without `parse`, which I find irrelevant. Also, we always available to move validation from options parsing phase to the run phase.
https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/cli.rb#L294

lib/sidekiq/cli.rb
```diff
@@ -272,10 +268,7 @@ def boot_system
         end
         options[:tag] ||= default_tag
       else
-        not_required_message = "#{options[:require]} was not required, you should use an explicit path: " +
```
Tensho: Ditto

test/helper.rb
```diff
@@ -15,36 +20,8 @@
   end
 end
 
-ENV['RACK_ENV'] = ENV['RAILS_ENV'] = 'test'
-
-trap 'TSTP' do
```
Tensho: I haven't found any reason to trap `TSTP` signal in tests

mperham: It's useful for debugging deadlocked tests, I had to use it years ago so it's ok to remove.

test/helper.rb
```diff
@@ -15,36 +20,8 @@
...
-  end
-
-  puts "=" * 80
-end
-
-require 'minitest/autorun'
-
 Sidekiq.logger.level = Logger::ERROR
 
-REDIS_URL = ENV['REDIS_URL'] || 'redis://localhost/15'
```
mperham: I don't want the test suite to use the default Redis `localhost:6379/0`

Tensho: OK, I may return back `ENV['REDIS_URL'] ||= 'redis://localhost/15'`, that should be enough I guess.
Why `15`? Is it kind of virtual host in RabbitMQ?

Tensho: Don't want to be a dick, but we already have tests which rely on default Redis host ^_^
https://github.com/mperham/sidekiq/blob/master/test/test_redis_connection.rb#L84
https://github.com/mperham/sidekiq/blob/master/test/test_redis_connection.rb#L126

Tensho:
@mperham Addressed all your concerns I think. Let me know if there's anything else you can think of.

test/helper.rb
mperham: Oops, yeah, I want people's default redis database to be untouched by Sidekiq's test suite.  We'll need to fix those tests (or make them read only)

Tensho: Already done this, Mike ^_^

mperham:
Thank you for this! I will add a changelog entry to give you credit.

Tensho:
@mperham Thank you 🙇  It's worth to note about `test/dummy` ultra-simple test Rails app somewhere (`README`, `CONTRIBUTING`), so other people don't mess with `./myapp` one. It took me a while to understand `myapp` purpose is for manual testing (it has it's own dependencies declaration in `Gemfile`), not auto tests.

Issue #3735, opened 2018-01-29
CLI does not fail if specified config file specified doesn't exist


thbar:
I've been surprised by the behaviour so I just wanted to ask if this is by design. Thanks!

Ruby version: 2.5.0
Sidekiq / Pro / Enterprise version(s): Sidekiq 5.0.0

While setting up a sidekiq install, I noticed that if one passes a config file like this:

```
bundle exec sidekiq -C config/foobar.yml -r ./setup/boot.rb
```

with `foobar.yml` an non-existent file, Sidekiq will start anyway & will default to the default configuration.

Note that creating the file then running with `--verbose` I can see the options (e.g. concurrency) taking effect.

mperham:
It is by design so things like capistrano-sidekiq can point to a file, even if it doesn't exist.  You can make a case that there should be a warning.

thbar:
I wondered about that - indeed it's convenient when a systemd config refers to it but it's not there yet, I was pretty sure that was the reasoning behind that.

Thanks for the clarification - just wanted to make sure I hadn't missed anything.

I'm going to go ahead and close this. Thanks!

alxgsv:
I've just shot myself in a leg because of this. Created config file for separate queue with typo in it's name several days ago and launched sidekiq. Today I found out that sidekiq has been running with default config all these days :-(

mperham:
I think we can add logic to raise if the user supplies a -C param and it doesn’t exist.

mperham:
PRs welcome, this should be relatively easy to implement.

PR #4047, opened 2018-12-09
Raise if the user supplies a -C param and it doesn’t exist

alxgsv:
Added config file existence check as discussed in comments of #3735.

At first I tried to move config file fallback logic from `config_parse`, because it's not really parsing of a config. But then I thought that it would be more straightforward to just check file existence when optparse catches `-C`: easier to understand the error when it happens, easier to preserve config file fallback, and easier to understand this logic when/if modifying it later.

### DIFF
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -343,6 +343,9 @@ module Sidekiq
         end
 
         o.on '-C', '--config PATH', "path to YAML config file" do |arg|
+          if !File.exist?(arg)
+            raise ArgumentError, "can't find config file #{arg}"
+          end
           opts[:config_file] = arg
         end
 


Tensho:
@alxgsv Hi, thanks for your effort 🙇  I guess you have to incorporate to [this place](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/cli.rb#L372):
```
%w[config/sidekiq.yml config/sidekiq.yml.erb].each do |filename|
  opts[:config_file] ||= filename if File.exist?(filename)
end
```
and it would be more appropriate to move this check to [`validate!`](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/cli.rb#L285) method.

alxgsv:
@Tensho thank you! I completely agree that it's a good practice to leave options parser with only typecasting. That's why I decided not to touch code you've mentioned. It's in the `parse_options` method but has nothing to do with option parsing. And it leaves no chance to observe original command line state at the validation stage (other then creating some additional flags, checks, etc). 

So I decided to touch as less code as possible to fix the bug, but as you said it's not ideal solution too. It's better to close this pull request, because I don't think I can fix this bug in other way without refactoring of `parse_options`, but I have no time for that.

Please ping me if I should open issue for that!

Tensho:
@alxgsv Fixed in #4054, does it make sense to you?

Issue #4046, opened 2018-12-07
Feature Request: structured logging format


renchap:
This is a best practice on production platforms to send logs to a central collector, parse them and store them as a structured document. This allows to create dashboards and alerts based on logs.

Current Sidekiq logs are plain text and need a custom parser to be processed. It would be awesome to be able to (optionally) enable a JSON (or similar, ideally modular) format. Log messages could include a type (start job / stop job), Job name, Job ID, duration (for a stop event) or even job parameters.

For Rails, the `lograge` gem provides this and is used in most production deployments: https://github.com/roidrage/lograge

Current implementations:
- https://github.com/Springest/Sidekiq-Logging-JSON
- Gitlab: https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/18121

olivierlacan:
Although it's a bit more convoluted, the SemanticLogger gem can also convert default Ruby/Rails logs into JSON output which is what we use at Pluralsight to stream JSON to Kibana: https://github.com/rocketjob/semantic_logger

I think the logging could be improved by providing essentials like: 
- perform duration
- completed or failed (exception)

In the `sidekiq.rb` initializer this is how I currently override Sidekiq's JobLogger:

```ruby
if Rails.logger.is_a?(SemanticLogger::Logger)
  module Sidekiq
    class JobLogger
      def call(_item, _queue)
        start = Time.current

        logger.info('start'.freeze)

        yield

        duration = elapsed(start)

        logger.info(message: "done: #{duration} sec", payload: {
          duration: duration * 1000
        })
      rescue StandardError
        duration = elapsed(start)

        logger.info(message: "fail: #{duration} sec", payload: {
          duration: duration * 1000
        })

        raise
      end
    end
  end
end
```

Tensho:
I'd definitely like to work on this for sure, guys. But I'd like to understand the problem we would like to solve. There are 2 aspects we have to consider:
- `Sidekiq.logger` (`Sidekiq::Logging.logger`) is a **general purpose logger**, exists to suport  multi-threaded system logging
- `Sidekiq::JobLogger` is a specific logger, repsonsible only for the job lifecycle logging (elapsed time), prime consumer of `Sidekiq.logger`.

As I understand from the issue request it would be nice to **get structured logging formatter for `Sidekiq::JobLogger` out of the box (without any extra dependencies)**. We don't care much about banner and other configuration info in JSON form here.

[...]

PR #4050, opened 2018-12-10
JSON Logger Formatter

Tensho:
In **"Feature Request: structured logging format"** #4046 we agreed to provide generic JSON logger formatter and pass any job relevant info with the context.

- Add `JSONFormatter` to logger `Sidekiq::Logging`
- Add elapsed time as context to `Sidekiq::JobLogger`
- Add `Sidekiq.logger_formatter` accessor

Example

```
> Sidekiq.logger.formatter
=> #<Sidekiq::Logging::Pretty:0x00007f8c63dec0a0 @datetime_format=nil>

> Sidekiq.logger.info("abc")
2018-12-12T15:06:52.807Z 3301 TID-ousw3yl55 INFO: abc
=> true

> Sidekiq::Logging.with_context("xyz") { Sidekiq.logger.info("abc") }
2018-12-12T15:07:19.009Z 3301 TID-ousw3yl55 xyz INFO: abc
=> true

> Sidekiq.logger.formatter = Sidekiq::Logger::JSON.new
=> #<Sidekiq::Logger::JSON:0x00007f8c62207330 @datetime_format=nil>

> Sidekiq.logger.info("abc")
{"timestamp":"2018-12-12T15:08:15.972Z","pid":3301,"tid":"ousw3yl55","context":[],"severity":"INFO","message":"abc"}=> true

> Sidekiq::Logging.with_context("xyz") { Sidekiq.logger.info("abc") }
{"timestamp":"2018-12-12T15:08:22.072Z","pid":3301,"tid":"ousw3yl55","context":["xyz"],"severity":"INFO","message":"abc"}=> true

> Sidekiq.logger_formatter = :json
=> :json

> Sidekiq.logger.formatter.class
=> Sidekiq::Logger::JSONFormatter

> Sidekiq.logger = nil
=> nil

> Sidekiq.logger_formatter = nil
=> nil

> Sidekiq.logger.formatter.class
=> Sidekiq::Logging::PrettyFormatter
```

### Improvements

- Add [`rspec-around`](https://github.com/splattael/minitest-around)

### Interesting Notes

- Found that JRuby returns the different `Thread.current` value outside and inside the any fiber:
```
> RUBY_PLATFORM
=> "java"
> p Thread.current
#<Thread:0x4d33940d run>
=> #<Thread:0x4d33940d run>
> Fiber.new { Thread.current }.resume
=> #<Thread:0x28cbf866 run>
```
As far as `minitest-around` is implemented with the fibers, tests with stubbed `Thread.current` fail in JRuby.


Tensho:
Breaking changes could be introduced with backward compatibility like [this](https://github.com/mperham/sidekiq/issues/4046#issuecomment-446112552).

lib/sidekiq/logger.rb
```diff
@@ -1,70 +1,34 @@
...
 
-    class Pretty < Logger::Formatter
-      SPACE = " "
+    def initialize(*args)
+      super
 
-      # Provide a call() method that returns the formatted message.
-      def call(severity, time, program_name, message)
-        "#{time.utc.iso8601(3)} #{::Process.pid} TID-#{Sidekiq::Logging.tid}#{context} #{severity}: #{message}\n"
+      formatter_class = case Sidekiq.logger_formatter
```
Tensho: I guess it's a proper place to analyze a logger formatter config here.

lib/sidekiq/logger.rb
```diff
@@ -115,8 +79,35 @@ def self.reopen_logs
...
+      def call(severity, time, program_name, message)
+        "#{::Process.pid} TID-#{Sidekiq.tid}#{format_context(Sidekiq.logger.context)} #{severity}: #{message}\n"
+      end
+    end
+
+    class JSONFormatter < PrettyFormatter
+      def call(severity, time, program_name, message)
+        Sidekiq.dump_json(
+          timestamp: time.utc.iso8601(3),
+          pid: ::Process.pid,
```
Tensho: I thought about `Sidekiq.pid` as an alias to `Process.pid`, but maybe it could confuse other contributors 🤔

lib/sidekiq/processor.rb
```diff
@@ -37,7 +36,7 @@ def initialize(mgr)
       @thread = nil
       @strategy = (mgr.options[:fetch] || Sidekiq::BasicFetch).new(mgr.options)
       @reloader = Sidekiq.options[:reloader]
-      @logging = (mgr.options[:job_logger] || Sidekiq::JobLogger).new
+      @job_logger = (mgr.options[:job_logger] || Sidekiq::JobLogger).new
```
Tensho: Would like to stress here we gonna use the specific logger (actually wrapper around the generic logger).

Tensho:
@mperham I've left a couple of comments to review changes easier. Please let me know what do you think about the code 🙇

lib/sidekiq.rb
```diff
@@ -184,11 +185,20 @@ def self.dump_json(object)
...
+    @logger ||= Sidekiq::Logger.new(STDOUT, level: Logger::INFO)
   end
-  def self.logger=(log)
-    Sidekiq::Logging.logger = log
+
+  def self.logger=(logger)
+    @logger = logger
+  end
+
+  def self.tid
```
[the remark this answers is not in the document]
mperham: Technically it's thread-specific.  I'm not sure it belongs in the Sidekiq module or perhaps Sidekiq::Util would be more appropriate since it's an internal impl, not an API designed to be used by app code.

lib/sidekiq/job_logger.rb
```diff
@@ -1,25 +1,49 @@
 # frozen_string_literal: true
+
 module Sidekiq
   class JobLogger
 
-    def call(item, queue)
```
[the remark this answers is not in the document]
mperham: It's there so others can plug in their own logger class and use those arguments if they want.

mperham:
I'm ok with spec-style tests but don't want to use rspec and would prefer to stick with Ruby's stdlib as much as possible.

mperham:
You can see the major release notes in the root of the repo, where breaking changes are called out, e.g.

https://github.com/mperham/sidekiq/blob/master/4.0-Upgrade.md

lib/sidekiq/processor.rb
```diff
@@ -236,7 +235,7 @@ def clear
     WORKER_STATE = SharedWorkerState.new
 
     def stats(job_hash, queue)
-      tid = Sidekiq::Logging.tid
+      tid = Sidekiq.tid
```
Tensho: Inline local variable

test/test_logger.rb
```diff
@@ -0,0 +1,169 @@
...
+
+class TesLogger < Minitest::Test
+  describe Sidekiq::Logger do
+    after do
+      Thread.current[:sidekiq_context] = nil
+    end
+
+    describe 'initialization' do
+      describe 'formatter' do
+        subject { Sidekiq::Logger.new(STDERR).formatter }
```
Tensho: Prefer to log to `StringIO` instead of default `STDOUT`/`STDERR`. It's easier to assert content match if needed.

test/test_logger.rb
```diff
@@ -0,0 +1,169 @@
...
+
+        describe 'default formatter' do
+          it 'sets pretty formatter' do
+            assert_kind_of Sidekiq::Logger::PrettyFormatter, subject
+          end
+        end
+
+        describe 'when DYNO env var is present' do
+          before do
+            ENV['DYNO'] = 'dyno identifier'
```
Tensho: Looks ugly, but have no idea why the next doesn't work:
```
allow(ENV).to receive(:[]).with('DINO').and_return('dyno identifier')
```
Maybe because of the different string `'DINO'` objects?

test/test_logger.rb
```diff
@@ -0,0 +1,169 @@
...
+          end
+
+          it 'sets without timestamp formatter' do
+            assert_kind_of Sidekiq::Logger::WithoutTimestampFormatter, subject
+          end
+        end
+
+        describe 'when logger formatter :json' do
+          before do
+            Sidekiq.logger_formatter = :json
```
Tensho: Replace with mock:
```
before do
  allow(Sidekiq).to receive(:logger_formatter).and_return(:json)
end
```

Tensho:
Agreements

- Split the changes to 2 separate branches – one for the JSON structured logging feature, another for the `Sidekiq::Logging` refactoring
- Keep `tid` method duplicated in Sidekiq::Logging` and `Sidekiq::Processor`
- Add deprecation warning to `Sidekiq::Logging`
- Remove `rspec-mocks` in favor of minitest mocks

Tensho:
@mperham I've left only JSON formatter here with a minimal refactoring of the logging context. Added warning regarding breaking changes in Sidekiq 6. Please let me know is it sufficient to be merged?

lib/sidekiq/logging.rb
```diff
@@ -5,31 +5,56 @@
...
       def call(severity, time, program_name, message)
-        "#{::Process.pid} TID-#{Sidekiq::Logging.tid}#{context} #{severity}: #{message}\n"
+        "#{::Process.pid} TID-#{Sidekiq::Logging.tid}#{format_context(Sidekiq::Logging.context)} #{severity}: #{message}\n"
+      end
+    end
+
+    class JSON < Pretty
+      def call(severity, time, program_name, message)
+        Sidekiq.dump_json(
+          timestamp: time.utc.iso8601(3),
```
mperham: let's reduce these key sizes since they add logger payload size:

ts, pid, tid, ctx, sev, msg

test/test_job_logger.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+require_relative 'helper'
+require 'sidekiq/job_logger'
+
+class TesJobLogger < Minitest::Test
```
mperham: TestJobLogger

lib/sidekiq/logging.rb
```diff
@@ -5,31 +5,56 @@
...
+        "#{time.utc.iso8601(3)} #{::Process.pid} TID-#{Sidekiq::Logging.tid}#{format_context(Sidekiq::Logging.context)} #{severity}: #{message}\n"
       end
 
-      def context
-        c = Thread.current[:sidekiq_context]
-        " #{c.join(SPACE)}" if c && c.any?
+      private
+
+      def format_context(context)
+        ' ' + context.join(' ') if context.any?
```
mperham: I think the context should format as "class=FooWorker tid=12345abc" in JSON.  In other words, adding context is just adding more key/value pairs. The JSON output may include those keys while the plain logger may omit certain keys (especially if we want to keep the same plain log output format).

mperham:
Close. There’s no need for the ctx{} subhash. All pairs at the top level.

lib/sidekiq/logging.rb
Tensho: Or do we leave the context as an array of strings internally and concatenated with spaces string externally? I mean we can format context per formatter, but what should be considered as internal data structure – array of strings or hash with string keys/values?

Tensho: The solid context string is not very structured, but as far as we don't fully control the context I think it's a wise decision. That means we leave user to provide `key=value` strings to the context. Correct me if I'm wrong?

mperham:
Internal format should be hash so keys can be overridden.

Tensho:
Sorry, I'm a little bit lost. We can't switch to unified (for every formatter) **hash** context, because that will break backward compatibility. For example, `#job_hash_context` is intentionally exposed, method accepts **hash** and returns **string**. The only way I see in this case – accept both strings and hashes to the context array and then concatenate with something like this:
```
def format_context(context)
  return if context.empty?

  context.reduce('') do |acc, part|
    acc << case part
    when String
      part
    when Hash
      part.map { |k, v| "#{k}=#{v}" }.join(' ')
    end
  end
end
```
The context internal format is **array of string and hashes**. Is it what you meant?

test/test_logger.rb
mperham: Great idea for testing.

test/test_logger.rb
mperham: You've misspelled DYNO, that's why?

mperham:
At the moment I believe we should break compatibility in this case.  This is a major logger update so it's understood that some things will change.

Tensho:
Mike, I intentionally split initial branch to `json-logger-fomratter` and `sidekiq-logger` branches to eliminate any breaking change ^_^ 

OK, I'll shift the logger context to **Hash** internally and each kind of formatter will format with own rules. That should look like:
```
job_hash = {
  class: 'BombasticWorker', 
  jid: 'abc5b065c5c4b27fc1102833'
}
elapsed_hash = {
  elapsed: '0.02 sec'
}
Sidekiq::Logging.with_context(job_hash) do
  Sidekiq::Logging.with_context(elapsed_hash) do
    ...
  end
end

# Pretty Formatter (strips class key)

2020-01-01T00:00:00.000Z 4710 TID-ouy7z76mx BombasticWorker JID-abc5b065c5c4b27fc1102833 elapsed: 0.02 sec INFO: This is Fantastic\n

# JSON Formatter

{
  ts: '2020-01-01T00:00:00.000Z',
  pid: 4710,
  tid: 'ouy7z76mx',
  ctx: 'class=BombasticWorker jid=abc5b065c5c4b27fc1102833 elapsed=0.02 sec',
  sev: 'INFO',
  msg: 'This is Fantastic\n'
}.to_json
```
Please confirm my assumptions and that we gonna introduce JSON formatter in Sidekiq **6**, not **5** 🙇 

P.S. I'm OK to iterate on this more and more until I fully understand the goal ^_^

test/test_logger.rb
Tensho: Ah, shame to me, thank you ^_^ BTW, how it should look like in `minitest`? Like this:
```
ENV['DYNO'].stub do
  ...
end
```
?

Tensho:
@mperham A probe to [convert the context to hash internally](https://github.com/mperham/sidekiq/pull/4050#issuecomment-448139611). I feel more comfortable with hash instead of array of strings. Pay attention, context in `JSON` formatter is left to internal (Hash) as it conforms the whole structure of the log entry. Also important it's a breaking change for those, who customized `#job_hash_context` method.
https://github.com/Tensho/sidekiq/compare/json-logger-formatter...Tensho:hash-context

Please let me know your opinion.

P.S. What do you think we rename `Pretty` –> `PrettyFormatter`, `JSON` –> `JSONFormatter`?

Tensho:
Updated [Logging](https://github.com/mperham/sidekiq/wiki/Logging) Wiki Page with more concise example of the custom job logger. Please take a look to the [diff](https://github.com/mperham/sidekiq/wiki/Logging/_compare/d1e05c88ee0ce0dfc52c98e31594c91feab12f74...8c0a778490e22b94069c728d8c589a86fef1e1ee) 🙇

mperham:
I'm unclear why the hash to log doesn't look like this:

```ruby
{
  ts: '2020-01-01T00:00:00.000Z',
  pid: 4710,
  tid: 'ouy7z76mx',
  class: 'BombasticWorker',
  jid: 'abc5b065c5c4b27fc1102833',
  elapsed: 0.02,
  sev: 'INFO',
  msg: 'This is Fantastic\n'
}
```

Why keep a complex substructure within ctx?

Tensho:
> Why keep a complex substructure within ctx?

I'm don't mind to make the entire hash flatten, but I find segregation of **mandatory** (ts, pid, tid, sev, msg) part from the **arbitrary** (cxt) one convenient.

Please answer the next questions and confirm/reject my assumption 🙇 

> What do you think if we rename `Pretty` –> `PrettyFormatter`, `JSON` –> `JSONFormatter`?

> Mike, I intentionally split initial branch to `json-logger-fomratter` (current) and `sidekiq-logger` branches to eliminate any breaking change in the first one ^_^
> **Please confirm my assumptions and that we gonna introduce JSON formatter in Sidekiq 6, not 5 🙇**

mperham:
The issue is that keeping a substructure makes working with log query tools like `jq` more complex because you've nested a few keys.  From the user's perspective, there's no reason for this.  I would be ok if you renamed `ctx` to `job`, since now the keys read much more naturally: job.jid, job.class, etc.

Yes, JSON will go into 6, not 5.  Those formatter names are fine, Json or JSON, whatever is easier.

Tensho:
I did have experience with `jq`, but found in **1** minute how to [parse the nested object](https://jqplay.org/s/a_oJ6fKd1U). I'm against to replace `ctx` with `job` key, because it semantically restricts what can be put to the context. Chances are someone would like to add application-specific info, not worker- or job-related. But if you think this scenario is contrived, I don't mind to do the replacement.

Tensho:
> I didn't have an experience with `jq`, but found in **1** minute how to [parse the nested object](https://jqplay.org/s/a_oJ6fKd1U). I'm against to replace `ctx` with `job` key, because it semantically restricts what can be put to the context. Chances are someone would like to add application-specific info, not worker- or job-related. But if you think this scenario is contrived, I don't mind to do the replacement.

@mperham Please let me know your decision 🙇  I guess this is the final step in this PR.

mperham:
Since JSON will go into Sidekiq 6.0, can you update the PR to target the "breaking_changes" branch, which has become 6.0?  I will find some time to look at this PR over the holiday week.

Tensho:
@mperham Changed base to `breaking_changes`. As far as the logging subsystem gained breaking changed and they are confirmed, it's reasonable to assume `Sidekiq::Logging` –> `Sidekiq::Logger` shift. However, I'd prefer to do this in a separate PR on top of the merged `breaking_changes`.

Tensho:
@mperham Merry Christmas 🎅  Did you have a chance to look at the branch?

Tensho:
https://github.com/Tensho/sidekiq/pull/1

Tensho:
Removed `rails 4` appraisal. Rebased the branch. Please let me know if something else is required.

Tensho:
As far as you dropped ruby versions until `2.4.4` is it reasonable to update the gemspec either?
```
gem.required_ruby_version = ">= 2.4.4"
```

mperham:
Really nice work and thank you for patiently working with my feedback!  I'll make a few changes here and there but you've done some great work here. Would you like a commit bit so you can make any future work in branches directly in mperham/sidekiq?

Tensho:
Thank you, yes, it’d be great. Happy New Year ^_^ 

Andrew Babichev

PR #4054, opened 2018-12-11
Check Config File Existence

### LINKED ISSUE #3735: CLI does not fail if specified config file specified doesn't exist

Tensho:
Ispired by #4047, Fixes #3735.

Parse `sidekiq.yml[.erb]` only in case:
- `:config_file` is not present as CLI argument
- `:require` is a directory (Rails application)
-  Expanded path to `config/sidekiq.yml[.erb]` against `:require` exists

Example when default config file is considered:
```
$ tree haiku
haiku
└─ config
    └─ sidekiq.yml
$ sidekiq -v -r ./haiku
...
:require => "./haiku"
:config_file => "/Users/tensho/Projects/sidekiq/haiku/config/sidekiq.yml"
...
```

### DIFF
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -238,8 +238,11 @@ module Sidekiq
           raise ArgumentError, "No such file #{opts[:config_file]}"
         end
       else
-        %w[config/sidekiq.yml config/sidekiq.yml.erb].each do |filename|
-          opts[:config_file] ||= filename if File.exist?(filename)
+        if File.directory?(opts[:require])
+          %w[config/sidekiq.yml config/sidekiq.yml.erb].each do |filename|
+            path = File.expand_path(filename, opts[:require])
+            opts[:config_file] ||= path if File.exist?(path)
+          end
         end
       end
 
@@ -401,8 +404,8 @@ module Sidekiq
       end
     end
 
-    def parse_config(cfile)
-      opts = YAML.load(ERB.new(IO.read(cfile)).result) || {}
+    def parse_config(path)
+      opts = YAML.load(ERB.new(File.read(path)).result) || {}
 
       if opts.respond_to? :deep_symbolize_keys!
         opts.deep_symbolize_keys!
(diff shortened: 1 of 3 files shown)


test/test_cli.rb
```diff
@@ -244,6 +244,18 @@ class TestCLI < Minitest::Test
             assert_equal 3, Sidekiq.options[:queues].count { |q| q == 'seldom' }
           end
         end
+
+        describe 'default config file' do
+          describe 'when required path is a directory' do
+            focus
```
mperham: Forgot to remove this?

Tensho: Yep, sorry, will remove

mperham:
Thanks!

navied:
This is a bit of breaking change, the previous behavior is to check the default config directories on startup now a config flag is required. Might want to make a note of this in the change log. @mperham

mperham:
@navied Could you be more specific?  Sidekiq should always pick up the default config file if none is specified.  If it does not, that's a bug.

navied:
Prior to 5.2.4, in a non-rails sidekiq setup the default config folder of config/sidekiq.yml was picked up. But the default config file is still picked up with a rails w/ sidekiq setup.

mperham:
Thankls, @tensho do you have time to look into this?

lib/sidekiq/cli.rb
```diff
@@ -367,11 +383,8 @@ def parse_options(argv)
         logger.info @parser
         die 1
       end
-      @parser.parse!(argv)
 
-      %w[config/sidekiq.yml config/sidekiq.yml.erb].each do |filename|
```
navied: Missing code that is potentially the issue.

Tensho:
I wonder `sidekiq` supports default config file in `config` folder aside of the Rails applications. However, the documentation provides the pretty clear statement, without Rails connotation:
> The Sidekiq configuration file is a YAML file that Sidekiq server uses to configure itself, by default located at config/sidekiq.yml.
Sorry for that oversight 🙇  I'll make an appropriate change.

Tensho:
Adjusted in #4077

PR #4055, opened 2018-12-14
Match Redis connection pool size verification with defaults

agrobbin:
When debugging some issues we were having with Sidekiq processes hanging, I noticed that the verification of the Redis connection pool size is different than the code that adds the 5 additional connections to the pool by default.

This synchronizes them, while also adjusting the math check to take into account the additional connections needed for Sidekiq's built-in features.

### DIFF
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -10,6 +10,7 @@ require 'fileutils'
 require 'sidekiq'
 require 'sidekiq/util'
 require 'sidekiq/launcher'
+require 'sidekiq/redis_connection'
 
 module Sidekiq
   class CLI
@@ -76,7 +77,7 @@ module Sidekiq
       # Since the user can pass us a connection pool explicitly in the initializer, we
       # need to verify the size is large enough or else Sidekiq's performance is dramatically slowed.
       cursize = Sidekiq.redis_pool.size
-      needed = Sidekiq.options[:concurrency] + 2
+      needed = Sidekiq.options[:concurrency] + RedisConnection::AUXILIARY_CONNECTIONS
       raise "Your pool of #{cursize} Redis connections is too small, please increase the size to at least #{needed}" if cursize < needed
 
       # cache process identity
--- a/lib/sidekiq/redis_connection.rb
+++ b/lib/sidekiq/redis_connection.rb
@@ -5,6 +5,8 @@ require 'uri'
 
 module Sidekiq
   class RedisConnection
+    AUXILIARY_CONNECTIONS = 5
+
     class << self
 
       def create(options={})
@@ -18,7 +20,7 @@ module Sidekiq
         size = if options[:size]
                  options[:size]
                elsif Sidekiq.server?
-                 Sidekiq.options[:concurrency] + 5
+                 Sidekiq.options[:concurrency] + AUXILIARY_CONNECTIONS
                elsif ENV['RAILS_MAX_THREADS']
                  Integer(ENV['RAILS_MAX_THREADS'])
                else
@@ -46,7 +48,9 @@ module Sidekiq
       #   - enterprise's leader election
       #   - enterprise's cron support
       def verify_sizing(size, concurrency)
-        raise ArgumentError, "Your Redis connection pool is too small for Sidekiq to work. Your pool has #{size} connections but must have at least #{concurrency + 2}" if size <= concurrency
+        if size < (concurrency + AUXILIARY_CONNECTIONS)
+          raise ArgumentError, "Your Redis connection pool is too small for Sidekiq to work. Your pool has #{size} connections but must have at least #{concurrency + AUXILIARY_CONNECTIONS}"
+        end
       end
 
       def build_client(options)
--- a/test/test_redis_connection.rb
+++ b/test/test_redis_connection.rb
@@ -83,6 +83,12 @@ class TestRedisConnection < Minitest::Test
           ENV.delete('RAILS_MAX_THREADS')
         end
       end
+
+      it "raises if the server pool size is too small to support other features" do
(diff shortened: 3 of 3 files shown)


lib/sidekiq/redis_connection.rb
```diff
@@ -5,6 +5,8 @@
 
 module Sidekiq
   class RedisConnection
+    ADDITIONAL_CONNECTIONS_REQUIRED = 5
```
Tensho: `AXILIARY_CONNECTION`?

Tensho:
It would be really nice to cover the different pool size options with tests 🙇

agrobbin:
> I wonder, does CLI contains exactly the [same pool verification](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/cli.rb#L78)?

Looks like it does! I can match up that implementation with this one, or look at maybe removing the duplication entirely. If I'm reading the code correctly, maybe this `CLI` code is unnecessary, unless we are checking for the `!Sidekiq.server?` case. Since we obtain the pool size from the `redis_pool`, which is generated via `RedisConnection.create`, it should `raise` before we even hit this `CLI` check (except when we're initializing the client rather than server).

agrobbin:
Ah I see, it's taking into account when `Sidekiq.redis = ConnectionPool.new(...)`. Will update that check then too!

agrobbin:
Updated @Tensho!

lib/sidekiq/cli.rb
```diff
@@ -76,7 +76,7 @@ def run
       # Since the user can pass us a connection pool explicitly in the initializer, we
       # need to verify the size is large enough or else Sidekiq's performance is dramatically slowed.
       cursize = Sidekiq.redis_pool.size
-      needed = Sidekiq.options[:concurrency] + 2
+      needed = Sidekiq.options[:concurrency] + Sidekiq::RedisConnection::AUXILIARY_CONNECTIONS
```
Tensho: Here we are in the lexical scope of the `Sidekiq` module, so constant lookup will find `RedisConnection` as far as `sidekiq.rb` file already required `redis_connection.rb` globally. IMO, this is more appropriate:
```
require 'sidekiq/redis_connection'
...
needed = Sidekiq.options[:concurrency] + RedisConnection::AUXILIARY_CONNECTIONS
```

test/test_redis_connection.rb
```diff
@@ -83,6 +83,12 @@ def server_connection(*args)
           ENV.delete('RAILS_MAX_THREADS')
         end
       end
+
+      it "raises if the server pool size is too small to support other features" do
+        Sidekiq.options[:concurrency] = 6
```
Tensho: `Sidekiq.options` represents a global state (configuration), please return it back to old value on test teardown 🙇

Tensho:
Mike wrote an [exellent article about extra connections](http://www.mikeperham.com/2015/10/14/optimizing-sidekiq/#drawbacks), maybe it's worth to update the text if the changes applied.

test/test_redis_connection.rb
agrobbin: This is already reset in the top-level `before` block of this test file.

lib/sidekiq/cli.rb
agrobbin: 👍 updated!

mperham:
I'm :-1: on this.  This PR will break all apps that just followed the directions and configured their pool to be conn+2.  We can't do that.

agrobbin:
@mperham it's definitely going to result in some churn on Sidekiq users configuring their own connection pool sizes, but right now aren't those apps already partially broken because their connection pool is too small for Sidekiq to work effectively?

mperham:
Right now we break on size <= conn and recommend size = conn+2.

I'm ok with upping that check to break when size < conn+2 and recommend size = conn+5.

agrobbin:
I thought the "break on size <= conn" while "recommend size = conn+2" was a bug in the `verify_sizing` method. Could you quickly explain why they aren't the same?

mperham:
To be conservative on app breakage, Sidekiq will be totally broken for size <= conn.  It will work with conn+1.

Remember if we break someone's app deploy because of a random change like this in a patch release, we've totally destroyed their trust in this project and its maintainers.  I take this very seriously.

mperham:
Since we've been recommending conn+2 for years now, I'm ok with bumping that check but would prefer to see it land in Sidekiq 6 and not a patch release.

agrobbin:
Totally get the concern about destroying users' trust in Sidekiq @mperham! I definitely don't want that.

I think what tripped me up here was the wording of the `ArgumentError` that Sidekiq is raising here:

> Your Redis connection pool is too small for Sidekiq to work. Your pool has #{size} connections but must have at least #{concurrency + 2}

The "must have at least" made me think that Sidekiq would be legitimately broken when `size < ( conn + 2)`.

If Sidekiq only *requires* 1 extra connection to work appropriately, would you be open to a slightly different approach that does something like this?

* If `size <= conn`, raise (as it already does)
* If `size.between?(conn + 1, conn + 4)`, log a warning that Sidekiq performs best when there are at least 5 additional Redis connections available

mperham:
The required connection count is not fixed.

```
Sidekiq requires n+2: processors + scheduler + heartbeat.
Sidekiq Pro requires n+3 ideally: sidekiq + pubsub.
Sidekiq Enterprise requires n+3 ideally: (no additional conns required)
```

Enabling Pro's super_fetch dramatically lessens the number of Redis connections required. Net result: there's no easy way to determine the right minimum so n+2 is a reasonable minimum for all.

agrobbin:
Alright I'm going to close this. It sounds like the current implementation is the ideal based on the varying requirements of Sidekiq, Pro, and Enterprise. Sorry for the confusion here @mperham @Tensho!

PR #4064, opened 2018-12-29
Move Proctitle from CLI to Launcher

Tensho:
Here are not actually breaking changes, just a refactoring we agreed before. Proctitles should be inside the `Sidekiq::Launcher` class as it starts/updates heartbeat, which is responsible for the process title updates. I've refactored tests a little bit either, gonna add some more in the nearest future.


mperham:
Sorry about that, I forgot about your open PR!

PR #4070, opened 2019-01-06
Run tests against 2.6.0

ohbarye:
This PR just tweaked `.travis.yml` so that CI runs tests against 2.6.0.
I assume (also hope!) `sidekiq` will support Ruby 2.6.0+ as well.

ref https://www.ruby-lang.org/en/news/2018/12/25/ruby-2-6-0-released/


ohbarye:
thank you too!

PR #4071, opened 2019-01-07
Detect Args Hash for ActiveJob::QueueAdapters Display Methods

mstruve:
Recently my team added arguments to Sidekiq jobs to help us identify types of jobs when they are displayed in the Sidekiq admin view. This works great for all jobs except our mailer jobs. Since we insert our additional args at the beginning of the arguments array the assumption that the args_hash we want to display is first in the array is no longer valid. I thought a better way might be to detect the args_hash we want rather than assuming it is the first argument. Let me know your thoughts!

Thanks!

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -348,7 +348,7 @@ module Sidekiq
                    job_class = @item['wrapped'] || args[0]
                    if 'ActionMailer::DeliveryJob' == job_class
                      # MailerClass#mailer_method
-                     args[0]['arguments'][0..1].join('#')
+                     job_class_args[0..1].join('#')
                    else
                     job_class
                    end
@@ -365,7 +365,7 @@ module Sidekiq
                     arg
                   end
                 when "ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper"
-                  job_args = self['wrapped'] ? args[0]["arguments"] : []
+                  job_args = self['wrapped'] ? job_class_args : []
                   if 'ActionMailer::DeliveryJob' == (self['wrapped'] || args[0])
                     # remove MailerClass, mailer_method and 'deliver_now'
                     job_args.drop(3)
@@ -383,6 +383,11 @@ module Sidekiq
 
     def args
       @args || @item['args']
+    end
+
+    def job_class_args
+      h = args.detect { |args| args.is_a?(Hash) && args.has_key?('job_class') } || {}
+      h['arguments']
     end
 
     def jid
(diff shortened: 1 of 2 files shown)


mstruve:
Both build failures look to be due to gem bundler issues
https://travis-ci.org/mperham/sidekiq/jobs/476391376
https://travis-ci.org/mperham/sidekiq/jobs/476391384

mperham:
Sidekick already has support for showing the underlying wrapped class of an active job. Can you be more explicit why that isn’t working for you?

mstruve:
We recently added this piece of middleware that will add a hash to our args like this
```ruby
module Sidekiq
  class AddTeamTagToArgs
    def call(worker_class, job, queue, redis_pool)
      job['args'] ||= []
      unless job['args'].include?({ 'team' => worker_class.to_s.constantize::TEAM_TAG })
        job['args'].unshift({ 'team' => worker_class.to_s.constantize::TEAM_TAG })
      end
    end
  end
end
```
This makes it easy to see in Sidekiq admin what team a job belongs to
team_tag_sidekiq

This tag also gets applied to the ActionMailer Jobs, however due to the assumption in the code that the first argument is the job class hash we end up breaking the Sidekiq dashboard views when an ActionMailer job is enqueued, busy, etc. 

Let me know if that makes sense!

mperham:
To me that looks like a hack.  Can you add the team to the end?  Can you re-namespace so the team is implicit in the class name:

```
AcmeCorp::Inventory::DoSomething
AcmeCorp::Shipping::SomethingWorker
```

One long-term solution is to support optional, ad-hoc job tags and their display/usage within the UI, API, etc. I would be happy to see this in Sidekiq 6.0.  Something like:

```
job = {
  'class' => 'FooWorker',
  'args' => [],
  'tags' => { 'team' => 'a' },
}
```

mperham:
You might also consider having a different Redis / Sidekiq cluster for each team so you can scale indepedently.  In a monolith, you'd use [sharding](https://github.com/mperham/sidekiq/wiki/Sharding) to route a job to a team cluster.

mstruve:
Haha, yeah its definitely a hack. I went looking to some sort of tag support to see if anyone had made a plugin but found nothing which is why we came up with this solution. We can't put the tags on the end of the args list bc the args list for some jobs is longer than what is displayed. I like the namespacing idea, it requires a bit of job reorganization but it would be more doable than sharding for us at this point. Thanks for the suggestion!

Would LOVE to see something like this in Sidekiq 6.0!

PR #4076, opened 2019-01-08
Look for config file at ./config/sidekiq.yml by default (if no -r flag provided)

### LINKED ISSUE #4075: Possible regressions in Sidekiq 5.2.4
Ruby version: 2.5.3
5.2.4 / 4.04 / 1.7.2

In 5.2.4 we saw two issues that I believe may be caused by the refactoring in this recent commit: (EDIT: it was a different recent commit, updated here)
https://github.com/mperham/sidekiq/commit/1d835551f0174a31697a5203e27148d61372cb25

The issues are: 
1. workers with no declared queues via the CLI only process `default`, ignoring queues declared in the config YAML. Workers with queues declared as CLI arguments pick up those queues.
2. The `RAILS_MAX_THREADS` environment variable appears to be suddenly taking precedence over the `concurrency` key in our configuration file. We have our app set to `3` and our `SIDEKIQ_CONCURRENCY=15`.

When we revert to 5.2.3, both of these issues resolve themselves. If we override `RAILS_MAX_THREADS` for our worker process at the CLI with `RAILS_MAX_THREADS=15`, we see the threads change to 15.

I'm still working up a failing test, but wanted to post this information as I'm working this out.

Please include your initializer and any error message with the full backtrace.
Our `config/sidekiq.yml`:
```
---
verbose: true
concurrency: <%= ENV.fetch("SIDEKIQ_CONCURRENCY", 5) %>
timeout: 25
queues:
  - [interims, 15]
  - [gc_assignments, 12]
  - [high_priority, 10]
  - [mailers, 10]
  - [toggle_notifications, 7]
  - [searchkick, 6]
  - [default, 5]
  - [low_priority, 2]
```

Environment settings:

```
RAILS_MAX_THREADS=3
SIDEKIQ_RAILS_MAX_THREADS=16
SIDEKIQ_CONCURRENCY=15
```

Procfile line for Heroku:
```
worker: NEW_RELIC_AGENT_ENABLED=true MALLOC_ARENA_MAX=2 PG_DB_CONNECTION_POOL=${SIDEKIQ_RAILS_MAX_THREADS} bundle exec sidekiq
```

(We have a second worker process which only processes `mailers` and `high_priority` that didn't only process `default` after the version bump.)

greysteil:
Fixes #4075.

v5.2.4 changed Sidekiq's behaviour when no `-r` or `-C` flag was provided. Previously, it would look in the default locations for a config file, and this PR restores that behaviour.

### DIFF
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -238,12 +238,14 @@ module Sidekiq
         if opts[:config_file] && !File.exist?(opts[:config_file])
           raise ArgumentError, "No such file #{opts[:config_file]}"
         end
+      elsif opts[:require] && File.directory?(opts[:require])
+        %w[config/sidekiq.yml config/sidekiq.yml.erb].each do |filename|
+          path = File.expand_path(filename, opts[:require])
+          opts[:config_file] ||= path if File.exist?(path)
+        end
       else
-        if opts[:require] && File.directory?(opts[:require])
-          %w[config/sidekiq.yml config/sidekiq.yml.erb].each do |filename|
-            path = File.expand_path(filename, opts[:require])
-            opts[:config_file] ||= path if File.exist?(path)
-          end
+        %w[config/sidekiq.yml config/sidekiq.yml.erb].each do |filename|
+          opts[:config_file] ||= filename if File.exist?(filename)
         end
       end
 
--- a/test/test_cli.rb
+++ b/test/test_cli.rb
@@ -254,6 +254,15 @@ class TestCLI < Minitest::Test
               assert_equal 25, Sidekiq.options[:concurrency]
             end
           end
+
+          describe 'when no requirement path is provided' do
+            it 'tries ./config/sidekiq.yml' do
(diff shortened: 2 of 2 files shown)


lib/sidekiq/cli.rb
```diff
@@ -238,12 +238,14 @@ def setup_options(args)
...
+          opts[:config_file] ||= path if File.exist?(path)
+        end
       else
-        if opts[:require] && File.directory?(opts[:require])
-          %w[config/sidekiq.yml config/sidekiq.yml.erb].each do |filename|
-            path = File.expand_path(filename, opts[:require])
-            opts[:config_file] ||= path if File.exist?(path)
-          end
+        %w[config/sidekiq.yml config/sidekiq.yml.erb].each do |filename|
+          opts[:config_file] ||= filename if File.exist?(filename)
```
Tensho: That shouldn't work as far as `#parse_config` expects an absolute path. Does it make sense?

Tensho:
@greysteil Thank you 🙇

PR #4077, opened 2019-01-08
Consider Default Required Path on Config File Search

### LINKED ISSUE #4075: Possible regressions in Sidekiq 5.2.4

Tensho:
Fixes #4075

To find out the config file dir we make the next assumptions with the corresponding precedence:

1. CLI `:require` option is an existing dir – take it for `config/sidekiq.yml`
2. CLI `:require` option is a file – take default Sidekiq`:require` option  – `.` (current directory)
3. CLI `:require` option is ommitted – take default Sidekiq`:require` option  – `.` (current directory)

In tests we stub default Sidekiq `:require` option (`.`) with the dummy app folder (`./test/dummy`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ cache: bundler
 services:
   - redis-server
 before_install:
-  - gem update --system
+  - gem install bundler -v '< 2'
 gemfile:
   - gemfiles/rails_4.gemfile
   - gemfiles/rails_5.gemfile
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -239,17 +239,22 @@ module Sidekiq
           raise ArgumentError, "No such file #{opts[:config_file]}"
         end
       else
-        if opts[:require] && File.directory?(opts[:require])
-          %w[config/sidekiq.yml config/sidekiq.yml.erb].each do |filename|
-            path = File.expand_path(filename, opts[:require])
-            opts[:config_file] ||= path if File.exist?(path)
-          end
+        config_dir = if File.directory?(opts[:require].to_s)
+          File.join(opts[:require], 'config')
+        else
+          File.join(options[:require], 'config')
+        end
+
+        %w[sidekiq.yml sidekiq.yml.erb].each do |config_file|
+          path = File.join(config_dir, config_file)
+          opts[:config_file] ||= path if File.exist?(path)
         end
       end
 
       # parse config file options
       opts = parse_config(opts[:config_file]).merge(opts) if opts[:config_file]
 
+      # set defaults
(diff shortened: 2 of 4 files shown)


Tensho:
Sorry, why do you think it will not work? [Here](https://github.com/mperham/sidekiq/pull/4077/files#diff-b88772f3eb6a16d2b09c44e112553b1aR269) is the spec for bare `sidekiq` CLI command.

greysteil:
Oh, good point, actually that will work - had got confused about setting `require` in the default options, but you're totally right.

mperham:
master should be fixed now, anyone want to confirm?

andrew-newell:
> master should be fixed now, anyone want to confirm?

Master seems to have fixed it for my deployment

PR #4089, opened 2019-01-31
Rescue StandardErrors raised from exception's message

bwsewell:
When developers create custom error classes, there is potential for the `message` method to raise an error. Exception Inception!

```ruby
class BadError < StandardError

  def message
    # This will raise a NameError when trying to read this exception's message
    "Hello World" + foo
  end

end
```

Of course, this is not the fault of Sidekiq, however we do [rescue errors that an error handler may throw when logging](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/exception_handler.rb#L22). Other error handling libraries do not account for these mistakes, but Sidekiq jobs that throw errors with bad messages will end up not retrying ever.

### DIFF
--- a/lib/sidekiq/job_retry.rb
+++ b/lib/sidekiq/job_retry.rb
@@ -245,13 +245,14 @@ module Sidekiq
         exception_caused_by_shutdown?(e.cause, checked_causes)
     end
 
+    # Extract message from exception.
+    # Set a default if the message raises an error
     def exception_message(exception)
       begin
         # App code can stuff all sorts of crazy binary data into the error message
         # that won't convert to JSON.
         exception.message.to_s[0, 10_000]
       rescue StandardError => e
-        # If the message raises a StandardError, set a default
         "!!! ERROR MESSAGE THREW AN ERROR !!!"
       end
     end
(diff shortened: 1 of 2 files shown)


lib/sidekiq/job_retry.rb
```diff
@@ -142,7 +142,7 @@ def attempt_retry(worker, msg, queue, exception)
 
       # App code can stuff all sorts of crazy binary data into the error message
       # that won't convert to JSON.
-      m = exception.message.to_s[0, 10_000]
```
mperham: Hmm, I can see this causing a massive amount of headscratching. Should you provide a default error message rather than nil?

Never, ever, ever use `rescue nil`.  Be explicit and document what and why you are rescuing.

bwsewell: 👍 

Agreed, I don't want the retry error description to be overly confusing.

What about explicitly rescuing `StandardError` in a begin..end block and setting the default message to: `!!! ERROR MESSAGE RAISED AN ERROR !!!`?

mperham: Yep, that's fine.

bwsewell: Done 😄

PR #4111, opened 2019-02-27
Format latency

interlinked:
The latency time for queues when viewing /sidekiq/queues was previously listed in seconds.  This pull request formats this time in seconds, minutes, or hours, using the same code that is used elsewhere (ie. /sidekiq/busy) for formatting time.

### DIFF
--- a/web/views/queues.erb
+++ b/web/views/queues.erb
@@ -17,7 +17,7 @@
           <% end %>
         </td>
         <td><%= number_with_delimiter(queue.size) %> </td>
-        <td><%= relative_time(Time.at(queue.latency.seconds.ago)) %> </td>
+        <td><%= queue.latency == 0 ? '' : relative_time(Time.at(queue.latency.seconds.ago)) %> </td>
         <td class="delete-confirm">
           <form action="<%=root_path %>queues/<%= CGI.escape(queue.name) %>" method="post">
             <%= csrf_tag %>
--- a/web/views/queues.erb
+++ b/web/views/queues.erb
@@ -17,7 +17,7 @@
           <% end %>
         </td>
         <td><%= number_with_delimiter(queue.size) %> </td>
-        <td><% queue_latency = queue.latency %><%= number_with_delimiter(queue_latency.round(2)) %><%= (queue_latency < 60) ? '' : " (#{relative_time(Time.at(queue_latency.seconds.ago))})" %> </td>
+        <td><% queue_latency = queue.latency %><%= number_with_delimiter(queue_latency.round(2)) %><%= (queue_latency < 60) ? '' : " (#{relative_time(Time.at(Time.now.to_f - queue_latency))})" %> </td>
         <td class="delete-confirm">
           <form action="<%=root_path %>queues/<%= CGI.escape(queue.name) %>" method="post">
             <%= csrf_tag %>
--- a/web/views/queues.erb
+++ b/web/views/queues.erb
@@ -17,7 +17,7 @@
           <% end %>
         </td>
         <td><%= number_with_delimiter(queue.size) %> </td>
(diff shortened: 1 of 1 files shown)


mperham:
Personally, when viewing latency, I like to see the precise value, and not “about 1 minute ago”. WDYT?

interlinked:
I guess my use-case is that I am constantly dividing the latency by 3600 to see how many hours old the oldest job is, since I often have large numbers of jobs that are queued to run and then they take hours to complete.  I'm running this in my app right now and it's made a huge difference to me...but that's just me :-)

Also, it still gives exact seconds when it's less than a minute, and then after that it only shows minutes
image

Would this be better as an configurable option?
Or would displaying both be even better?:
`3495s (58 minutes ago)`

mperham:
I'm warming up to this.  Is there a way to make it display an interval, e.g. "37 seconds", not "37 seconds ago"?

interlinked:
Did some digging into your sidekiq code...

The date formatting is coming from your inclusion of https://github.com/hustcc/timeago.js in the application.js file, and specifically the locale dictionary there.  Messing around with that would probably mean that we'd want to change the translation for other languages as well.  It also _might mean_ that it would change how the Busy view (/sidekiq/busy) shows processes started "X hours ~~ago~~".

Not sure I know how to hack around in the js to make that work.  Might be easier to change the title "Latency" to something where "X ago" makes more sense.

mperham:
I like `3495s (58 minutes ago)` but how about if you only show the parens if latency > 60s?

Unfortunately "s" doesn't internationalize so it might be best to stick with no suffix.

interlinked:
We're both thinking the same way...I had made those two changes last night (although I wasn't thinking about internationalization with the "s"...just about keeping it the same as it was for consistency).

Here's an example of how it would look now.
image

I'll update the pull request with the code.

mperham:
You need to memoize queue.latency.  It's a round trip to Redis on each call.

interlinked:
Good point...ewww.

mperham:
I'm pretty sure that `seconds` and `ago` are ActiveSupport methods which we can't use.  Try this instead:

```
relative_time(Time.at(Time.now.to_f - queue_latency))
```

interlinked:
Ok.  Must be because I've got them included in my app.  I'll give that a whirl.

mperham:
Ok, any test we can reuse to exercise that if?  We can push a job with an artificial `enqueued_at`.

interlinked:
The only test I'm finding regarding the /queues view is 'can display queues' in test_web.rb (line 108), but latency isn't being tested yet.  I don't have any experience with minitest, but I'm guessing that to test it adequately we'd need to use something like the timecop gem to "freeze time" so that we could get latency values we're expecting.

How do you properly use `enqueued_at` with
`assert Sidekiq::Client.push('queue' => :foo, 'class' => WebWorker, 'args' => [1, 3])`?

mperham:
```
assert Sidekiq::Client.push('queue' => :foo, 'class' => WebWorker, 'args' => [1, 3], 'enqueued_at' => Time.now.to_f - 65)
```

mperham:
Ahh, man, that won't work because Sidekiq::Client isn't using `||=`.  Drat.

mperham:
I'm getting really weird minitest results when I try to get that one test to actually fail.  I'm not sure the suite is actually running that test for some reason.

interlinked:
I was trying `enqueued_at` the way you were suggesting, but couldn't get it to work...glad to see it wasn't me.

I was able to get it to run on my dev machine, but I was explicitly running the test_web.rb tests.

mperham:
I've spent several hours cleaning up the test suite.  test_sidekiqctl ruined everything; everything should now be working.

mperham:
Part of the test suite cleaning was an assertion to verify your change so we're all good now!

interlinked:
Great!  Hope the change is helpful to sidekiq users (and not something that comes back to bite you  :-)  )

PR #4132, opened 2019-03-24
Fix i18n middleware fallback

gearnode:
ref: https://github.com/contribsys/faktory_worker_ruby/pull/22


mperham:
Did you mean to include a random commit from 4117?

gearnode:
@mperham Yes my bad, fixed 😄

PR #4141, opened 2019-04-05
Explicitly signal that we handled an exception with a retry, fixes #4138

### LINKED ISSUE #4138: Job getting lost in case of failed retry
Ruby 2.6.2 with Sidekiq Pro 4.0.5.

I inspected the code and it turns out that in case of exception in `Sidekiq::JobRetry#attempt_retry` we are still acknowledging the job [here](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/processor.rb#L180) which leads to the job being removed from Redis. If we didn't ack the job it would eventually get restored by the SuperFetch as I understand.

Maybe it would be better not to ack the job in case of all unhandled exceptions in `Processor#process`?

mperham:
Under just the right conditions, we could lose a job:

- Job raises an error
- Retry subsystem catches error and tries to create a retry in Redis but this raises a "Redis down" exception
- Processor catches Redis exception and thinks a retry was created
- Redis comes back online just in time for the job to be acknowledged and lost

That's a very specific and rare set of steps but it can happen.

Instead have the Retry subsystem raise a specific error signaling that it created a retry.  There will be three common cases:

1. Job is successful: job is acknowledged.
2. Job fails, retry is created, Processor rescues specific error: job is acknowledged.
3. Sidekiq::Shutdown is raised: job is not acknowledged

Now there is another case:

4. Job fails, retry fails, Processor rescues Exception: job is NOT acknowledged. Sidekiq Pro's super_fetch will rescue the orphaned job at some point in the future.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -5,6 +5,7 @@
 HEAD
 ---------
 
+- Fix edge case where a job failure during Redis outage could result in a lost job [#4141]
 - Better handling of malformed job arguments in payload [#4095]
 - Restore bootstap's dropdown css component [#4099, urkle]
 - Allow `Sidekiq::Worker#set` to be chained
--- a/lib/sidekiq/job_retry.rb
+++ b/lib/sidekiq/job_retry.rb
@@ -56,7 +56,8 @@ module Sidekiq
   #    end
   #
   class JobRetry
-    class Skip < ::RuntimeError; end
+    class Handled < ::RuntimeError; end
+    class Skip < Handled; end
 
     include Sidekiq::Util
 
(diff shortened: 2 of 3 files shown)


tycooon:
Maybe you could clarify what was the actual problem and how did you fix it?

tycooon:
> Job fails, retry fails, Processor rescues Exception: job is NOT acknowledged. Sidekiq Pro's super_fetch will rescue the orphaned job at some point in the future.
that's the mechanism I would rely the most

mperham:
Nope, I phrased it poorly but that scenario should be extremely rare. I mean “the retry cannot be created in Redis due to unexpected error”

tycooon:
Welp, it happened already, and we only are using sidekiq for like 4 months 
I mean if there is problem in the code we should fix it right?

mperham:
What do you think this is?  A fix.

tycooon:
There is a bunch of code that I still don't understand, But I see that now we are not acking every exception and that's seems cool to me

tycooon:
I will test this version versus the last stable for my case

mperham:
I'll merge this later this week if I don't hear anything.

Issue #2794, opened 2016-01-27
Need new feature of filter logger message


zhen639:
For the moment, when a worker retry exhausted or raise an exception, sidekiq would logger the args and error message, but there maybe some private information in the args so that I don't want it be loged.
Is there anyhow to config the sidekiq logger for filter some parameters? similar to "Rails.application.config.filter_parameters += [:password, :args, :class]" in rails?

mperham:
There's no way to do this today.

mperham:
No plans to implement.  I welcome a PR to improve this, perhaps we can just remove the logging of the job args.

PR #4143, opened 2019-04-12
[WIP] Filter arguments with matched keys in web UI

davidpiegza:
Sidekiq prints the arguments of jobs on the "Retries" page which might contain sensitive data like passwords. There was already a discussion about this issue in ticket #2794. I played around a bit to filter out the values of matched keys in the web ui and wanted to get an opinion on this (if it's something that could be added to Sidekiq).

This code is just a quick implementation which filters out arguments for a matched key in a Hash. I could imagine extending it to also filter out arguments in the console and make the key arguments configurable (similar to Log Filtering in Rails).

@mperham WDYT?

### DIFF
--- a/lib/sidekiq/web/helpers.rb
+++ b/lib/sidekiq/web/helpers.rb
@@ -225,7 +225,7 @@ module Sidekiq
 
     def to_display(arg)
       begin
-        arg.inspect
+        filtered_arguments(arg).inspect
       rescue
         begin
           arg.to_s
@@ -233,6 +233,12 @@ module Sidekiq
           "Cannot display argument: [#{ex.class.name}] #{ex.message}"
         end
       end
+    end
+
+    def filtered_arguments(arg)
+      return arg unless arg.is_a?(Hash)
+
+      arg.merge(arg.select { |key, _| key.to_s =~ /password/i }.transform_values { '[FILTERED]' })
     end
 
     RETRY_JOB_KEYS = Set.new(%w(


gearnode:
@davidpiegza Why not just encrypt your sensitive information before push it on the Sidekiq job? This avoid have sensitive information in Redis, Sidekiq log and web ui. And with this way it's not required to maintain a list of sensitive keys.

Encrypted key is Sidekiq Enterprise feature but it can be simple to have minimal implementation using Sidekiq middleware feature and OpenSSL ruby lib.

mperham:
@gearnode is mostly correct, I would argue end developers should never be rolling their own encryption.

The [encryption feature](https://github.com/mperham/sidekiq/wiki/Ent-Encryption) is designed to hold all your sensitive job args and the Web UI natively supports it.

mperham:
I should note: even if you filter, you are still keeping the cleartext persisted to disk in Redis and transferring it in cleartext over the network (unless your Redis connection is TLS-enabled).  Encryption provides end-to-end safety.

davidpiegza:
@gearnode @mperham Thanks for clarifying, I'll go with encrypting it before pushing it to the Sidekiq job. I'll close this PR.

akofink:
Encryption solves the data sensitivity problem, but when a job fails, Sidekiq logs all the args, including the binary encrypted data :/ It makes for very ugly logs. In my use case, we're zipping a ~10MB file as one of the job args, and it causes headaches trying to look through errored logs to figure out what happened.

mperham:
@akofink please do not reopen old issues, open a new issue. Happy to see a PR if you want to fix your own issue.

akofink:
Notice I didn't reopen this? Just some feedback from a user; thought you might be interested.

mperham:
I meant reopen in the sense of kicking it back into everyone’s inbox. If you open a new issue, I would be happy to help.

PR #4153, opened 2019-04-27
change to easy-to-understand variable name queue

Madogiwa0124:
I modified variable name `q` to be easy to read.


mikegee:
This is probably helpful for non-native English speakers.

By the way, "q" and "queue" are pronounced the same, which makes it a very common abbreviation for developers that read them the same in their head.

PR #4210, opened 2019-07-16
Adding Support for Yajl-Ruby for JSON serialization

ritikesh:


mperham:
Why?

ritikesh:
Sorry, I should have added a description. 😅 

Just to give that extra flexibility to choose what JSON parser you want to use. Native JSON vs Yajl vs Oj. For slightly bigger payloads, Yajl seems to be better in terms of performance. Will update relevant benchmarks FYR.

Tensho:
I guess you are not intended to pass huge JSONs across Redis to Sidekiq server. For example, you may store processing data at S3 and pass only identifier as a job parameter to a target worker.

https://github.com/mperham/sidekiq/wiki/Best-Practices#1-make-your-job-parameters-small-and-simple

ritikesh:
These are not huge JSONs per-se. AR objects as a hash because we need to retain the state of the object when enqueuing the job. Pushing it to S3 will add unnecessary overhead of creating a file, network latencies and then deleting it post processing etc.

ritikesh:
the benchmarks were kinda skewed. Native JSON performs better in these cases at least. Shouldn't have jumped the gun, was just excited I guess. 😅 Closing this for now, thank you for your time.

PR #4211, opened 2019-07-16
Set acknowledge to false by default to avoid dropping jobs

mehwoot:
It appears that in some exceptional circumstances a job can be acknowledged as completed when it has not completed.  The code currently sets `ack = true` and then, in a variety of error cases, sets `ack = false`.  However, if an error were to occur immediately after an error had been rescued, then the block that sets `ack = false` would not run, but the ensure block still would, resulting in the job being acknowledged when it did not complete.  

For example, if an error got past the retry subsystem at exactly the same time that SIGTERM was sent to the process, then Thread.raise could be called after `rescue Exception => ex` but before `ack = false`, and the ensure block will be run with `ack = true`.  Another way it might happen (although I can't reproduce this) is if two SIGTERM signals were sent in quick succession to the process.

We are persistently seeing long running jobs being dropped in our production environment of sidekiq pro (with superfetch enabled) running on Heroku.  I can't reproduce the exact conditions on Heroku that seem to give rise to this so I can't say with certainty what the sequence is, but this change fixed it for us.  Even without a reproducible bug I think this change makes sense as it is a more defensive way of approaching this part of the codebase- only acknowledge the job as done if we're absolutely sure that it was completed. With `ack = false` as default, any crazy edge cases won't result in the job being dropped before completed.

### DIFF
--- a/lib/sidekiq/processor.rb
+++ b/lib/sidekiq/processor.rb
@@ -153,21 +153,22 @@ module Sidekiq
         return work.acknowledge
       end
 
-      ack = true
+      ack = false
       begin
         dispatch(job_hash, queue) do |worker|
           Sidekiq.server_middleware.invoke(worker, job_hash, queue) do
             execute_job(worker, cloned(job_hash["args"]))
           end
         end
+        ack = true
       rescue Sidekiq::Shutdown
         # Had to force kill this job because it didn't finish
         # within the timeout.  Don't acknowledge the work since
         # we didn't properly finish it.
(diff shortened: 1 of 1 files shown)


mperham:
I'm extremely leery to change this code as this is the most critical logic in the entire system. This might take some more time and testing.

mehwoot:
Makes sense.  We've forked so it's not an immediate concern to get this merged right now but we'd like to not have to maintain a separate repo forever.

Issue #2424, opened 2015-07-07
sidekiq_options :queue => :<my_custom_queue> fails to enqueue in it.


bgvo:
Sidekiq pushes job to 'default' queue instead of :<my_custom_queue>

My Job:

```rails
class CollectSidsFromItemsInProductionJob < ActiveJob::Base
	include GlobalID::Identification
	include Sidekiq::Worker
	
	sidekiq_options :queue => :high, :retry => false

	def perform(dst)
		Item.where(dst: dst).each do |item|
			Dataio.save_value(IMDB_SIDS_IN_PRODUCTION, item.si)
		end
	end
end
```

However, setting queue as a different option goes as expected:

```rails
class CollectSidsFromItemsInProductionJob < ActiveJob::Base
	include GlobalID::Identification
	include Sidekiq::Worker
	
	sidekiq_options :retry => false
	queue_as :high

	def perform(dst)
		Item.where(dst: dst).each do |item|
			Dataio.save_value(IMDB_SIDS_IN_PRODUCTION, item.si)
		end
	end
end
```

mperham:
You can't use sidekiq_options with active job.

PR #4213, opened 2019-07-21
Allow Sidekiq retry options in ActiveJob

pirj:
[An exception was introduced](https://github.com/mperham/sidekiq/commit/c685d6928ca6aec1f7569ff69871ef9a8dfb2738) that is raised on an attempt to include `Sidekiq::Worker` to an `ActiveJob::Base` descendant.

That was done in order to prevent options not supported by ActiveJob, specifically `queue` (also `pool` doesn't seem to make any effect). See #2424.

Support for `set` [was added later on](https://github.com/mperham/sidekiq/commit/d6538b0b4f3c419ea969dc6710c6431b76a05d2b), including the setting of `retry` from Active Job.

This change bridges the gap by allowing `ActiveJob::Base` descendants to include `Sidekiq::Options` and use `retry` option of `sidekiq_options`, and also `sidekiq_retry_in`, and `sidekiq_retries_exhausted`.

### DIFF
--- a/lib/sidekiq/worker.rb
+++ b/lib/sidekiq/worker.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "sidekiq/client"
+require "sidekiq/worker/options"
 
 module Sidekiq
   ##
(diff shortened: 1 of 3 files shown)


mperham:
I'd suggest renaming the module to Sidekiq::Worker::Options so AJ::Job can include any sidekiq_option, including backtrace or any 3rd party option.  Why limit options?

pirj:
I thought that the use of `queue` and `pool` options will be confusing (just like in #2424) since AJ has its own way of configuring the queue to be used and will ignore both those option settings when enqueueing.

mperham:
Just because you allow everything doesn't mean you need to document everything.  We don't need to document the pool option and if people use AJ, they are unlikely to use the queue option since AJ has its own well-documented method.

pirj:
Updated. As a result, everything options related is moved to `Sidekiq::Options`.
The only `sidekiq_options` method that's left in `Worker` is to have separate, more complete options documentation.

On a different note, would you accept an identical pull request to `5-x` branch?

mperham:
Could you name it Sidekiq::Worker::Options?  Sidekiq.options is already a thing and I want the naming to be clear..

mperham:
I think this refactoring is big enough to justify avoiding 5.x.  We can call this a new feature for 6.x.

pirj:
Fair point. Moved to `Sidekiq::Worker::Options`.

mperham:
I'm going to refactor this a bit but nice job, it's a nice improvement for those using ActiveJob who want to control various Sidekiq features too.

pirj:
Thanks a lot!

PR #4246, opened 2019-09-05
A couple of random small code cleanups

fatkodima:


lib/sidekiq.rb
```diff
@@ -182,6 +181,10 @@ def self.dump_json(object)
     JSON.generate(object)
   end
 
+  def self.stringify_keys(hash) # :nodoc:
```
mperham: I really don't want `stringify_keys` as part of the public Sidekiq API, I prefer it left as is.

fatkodima: Yes, i have the same feeling. Is is used in several places and maybe it will be more cleaner if it will be in another place or was implemented as a monkey patch, but as far as sidekiq is monkey patching free, so let's leave as it was. Reverted this change.

PR #4247, opened 2019-09-05
Stop heartbeat gracefully

fatkodima:


mperham:
Heartbeat does not stop until process death.

fatkodima:
Ah, ok. I was misled by the unreachable line below https://github.com/mperham/sidekiq/blob/2bcf3bfff17866e06e48b111646d1b6f3d08ca7e/lib/sidekiq/launcher.rb#L76

PR #4248, opened 2019-09-07
Add ability to sort 'Enqueued' page on Web UI by enqueued_at time

### LINKED ISSUE #4041: Add the Ability to Sort Enqueued Page Both By "Most Recently Enqueued" As Well As "Next To Process"
This is a feature request from #4037 

I'd like to have the ability to sort the "enqueued" page so that it shows the jobs either by "Most Recently Enqueued" (as it does now) as well as "Next To Process".

The rationale is that I'd like to be able to sit on the first page (especially with Live Poll enabled) and watch which items are about to be run (among other reasons)

fatkodima:
Screen Shot 2019-09-08 at 02 13 47
Also closes #4041

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -5,7 +5,7 @@
 HEAD
 ---------
 
-- Add ability to sort 'Enqueued' page on Web UI by enqueued_at time [#4248]
+- Add ability to sort 'Enqueued' page on Web UI by position in the queue [#4248]
 - Support `Client.push_bulk` with different delays [#4243]
 ```ruby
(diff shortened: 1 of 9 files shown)


mperham:
you aren’t sorting based on enqueued_at but just list index. Should the UI reflect that? Otherwise I see it as a potential source of support issues and bug reports.

fatkodima:
Ah, yes, it is just listing queue in head-tail and tail-head order, wrong column name.

> Should the UI reflect that?

You mean changing column name or something else?

fatkodima:
Updated. Please, review.
Screen Shot 2019-09-08 at 16 40 53

mperham:
Instead of "Position", use "#" which is language independent.

fatkodima:
@mperham Done. Anything else?

mperham:
thank you!

PR #4265, opened 2019-09-13
Shuffle queues in place before fetching

fatkodima:


fatkodima:
`shuffle!` - no
`uniq!` - yes, but it doesn't matter in this case

PR #4268, opened 2019-09-14
Allocate Chain entries lazily

fatkodima:
Since many applications are using bare oss sidekiq without any middlewares, this pr can help to save time and memory.

I have tested using `bin/sidekiqload`.

### Using `memory_profiler`, I got:
**Before**:
```
Total allocated: 999716276 bytes (10908573 objects)
Total retained:  12844675 bytes (1634 objects)
```

**After**:
```
Total allocated: 951803096 bytes (10109336 objects)
Total retained:  12826021 bytes (1330 objects)
```
Speed up in memory = `(999716276  - 951803096) / 999716276.0 * 100 ~ 5%`

### Using `allocation_tracer`, I got:
**Before**:
```
 ["/Users/fatkodima/Desktop/oss/sidekiq/lib/sidekiq/middleware/chain.rb",
  81,
  :T_ARRAY]=>[2, 0, 0, 0, 0, 0],
 ["/Users/fatkodima/Desktop/oss/sidekiq/lib/sidekiq/client.rb",
  216,
  :T_ARRAY]=>[300009, 0, 0, 0, 0, 0],
 ["/Users/fatkodima/Desktop/oss/sidekiq/lib/sidekiq/middleware/chain.rb",
  118,
  :T_IMEMO]=>[200007, 0, 0, 0, 0, 0],
 ["/Users/fatkodima/Desktop/oss/sidekiq/lib/sidekiq/middleware/chain.rb",
  118,
  :T_ARRAY]=>[200006, 0, 0, 0, 0, 0],
 ["/Users/fatkodima/Desktop/oss/sidekiq/lib/sidekiq/middleware/chain.rb",
  126,
  :T_ARRAY]=>[200006, 0, 0, 0, 0, 0],
 ["/Users/fatkodima/Desktop/oss/sidekiq/lib/sidekiq/middleware/chain.rb",
  127,
  :T_IMEMO]=>[400013, 0, 0, 0, 0, 0],
 ["/Users/fatkodima/Desktop/oss/sidekiq/lib/sidekiq/middleware/chain.rb",
  127,
  :T_DATA]=>[400012, 0, 0, 0, 0, 0]
.......
```

**After**:
All those lines roughly creating 1_700_000 objects were gone away.

### Time measurements:
**Before** (avg result):
```
bin/sidekiqload  19.94s user 5.34s system 98% cpu 25.581 total
```

**After**(avg result):
```
bin/sidekiqload  18.97s user 5.14s system 98% cpu 24.278 total
```

Speed up in time = `(25.581 - 24.278) / 25.581 * 100 ~ 5%`

### DIFF
--- a/lib/sidekiq/middleware/chain.rb
+++ b/lib/sidekiq/middleware/chain.rb
@@ -67,7 +67,6 @@ module Sidekiq
   module Middleware
     class Chain
       include Enumerable
-      attr_reader :entries
 
       def initialize_copy(copy)
         copy.instance_variable_set(:@entries, entries.dup)
@@ -78,8 +77,12 @@ module Sidekiq
       end
 
       def initialize
-        @entries = []
+        @entries = nil
         yield self if block_given?
+      end
+
+      def entries
+        @entries ||= []
       end
 
       def remove(klass)
@@ -114,6 +117,10 @@ module Sidekiq
         any? { |entry| entry.klass == klass }
       end
 
+      def empty?
+        @entries.nil? || @entries.empty?
+      end
+
       def retrieve
         map(&:make_new)
       end
@@ -123,6 +130,8 @@ module Sidekiq
       end
 
       def invoke(*args)
+        return yield if empty?
+
         chain = retrieve.dup
         traverse_chain = lambda do
           if chain.empty?


mperham:
I don’t really like the Rails initializer. Seems unnecessary.

fatkodima:
Won't threads compete for `Sidekiq`s module `@client_chain` object instantiation on first access to `Sidekiq.client_middleware`? Can something bad happen if will?

mperham:
The users Sidekiq.rb initializer is the thing which will touch it and Rails boots on a single thread. Also, cli.rb touches the client middleware to log its contents, not sure if that’s done before launching threads.

fatkodima:
Yes, there is a line inside cli.rb which touches server middleware explicitly
https://github.com/mperham/sidekiq/blob/705d6ca7ea2ce73b3244b740fc0d0f8f19db8c60/lib/sidekiq/cli.rb#L71

And 2 lines, which touch both middleware chains, but only in debug mode
https://github.com/mperham/sidekiq/blob/705d6ca7ea2ce73b3244b740fc0d0f8f19db8c60/lib/sidekiq/cli.rb#L77-L78

So, probably, would need to add line for client middleware, similar to first example? 
So for sidekiq server, middlewares are partly eager loaded:
- server middleware - yes
- client middleware - only when in debug mode

> The users Sidekiq.rb initializer is the thing which will touch it and Rails boots on a single thread.

But when I have initializer like this:
```ruby
    config.redis = { :namespace => 'myapp', :size => 25, :url => 'redis://myhost:8877/0' }
  end
```

without explicitly calling `client_middleware`, it stays not initialized. And when I have rails with puma, for example, multiple threads enqueueing jobs may compete for that object creation.

What do you think?

mperham:
It’s a one time startup cost that might lead to one or two more Arrays, right?  is this worth the code complexity?

fatkodima:
Initially I was thinking not about saving a couple of objects, but rather something like atomically `@client_chain` allocation and assignment inside `#client_middleware` before threads starting to use it. But thinking twice, looks like this is not a real problem. 
So removed rails initializer.

mperham:
Nicely done, you’re doing a great job with these re-factorings.

fatkodima:
Thanks @mperham for your amazing work! ❤️

PR #4269, opened 2019-09-14
Reduce allocated objects

fatkodima:
Using `bin/sidekiqload` with `memory_profiler` I got:

**Before**:
```
Total allocated: 952250916 bytes (10115468 objects)
Total retained:  12802918 bytes (2000 objects)
```

**After**:
```
Total allocated: 934688905 bytes (9715635 objects)
Total retained:  12822122 bytes (2185 objects)
```

Memory savings = `(952250916 - 934688905) / 952250916.0 * 100 ~ 2%`
Object allocations reduced by = `(10115468 - 9715635) / 10115468.0 * 100 ~ 4%`
And time savings i have seen are also `~4%`.

Those savings are a bit smaller than in previous pr, but taking into account the simple changes, I think it's worth considering. 

Also line
```ruby
constant.const_defined?(name, false) ? constant.const_get(name, false) : constant.const_missing(name)
``` 
is a bit redundant, since `const_get(name, false)` calls `const_missing` when it can not find a constant.

### DIFF
--- a/lib/sidekiq/logger.rb
+++ b/lib/sidekiq/logger.rb
@@ -9,7 +9,7 @@ module Sidekiq
       ctx.merge!(hash)
       yield
     ensure
-      hash.keys.each { |key| ctx.delete(key) }
+      hash.each_key { |key| ctx.delete(key) }
     end
 
     def ctx
--- a/lib/sidekiq/processor.rb
+++ b/lib/sidekiq/processor.rb
@@ -269,13 +269,15 @@ module Sidekiq
     end
 
     def constantize(str)
+      return Object.const_get(str) unless str.include?("::")
+
       names = str.split("::")
       names.shift if names.empty? || names.first.empty?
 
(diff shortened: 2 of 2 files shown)


mperham:
I like the logger and processor changes.  Not as much the other two.  I really don't like to use the ! methods unless it's in a hot path.

fatkodima:
I'm not a fan of bang methods also, and unlikely that somebody will call `#push_bulk` many times per second, but push-bulking just once an array of 10_000 jobs will save us 2 useless arrays of the same size. Not much, but not little for such change. 1) So should i remove this change or left it as is?

Change in `Enq` class will save us client creation on each poll. Not much savings, but why allocate when we can reuse? 2) So should i remove this change or left it as is?

mperham:
Get rid of the other two.

fatkodima:
Changed as requested.

PR #4280, opened 2019-09-21
Support job tags

### LINKED ISSUE #4073: Support for ad-hoc job tags?
In #4071 we raised the idea of tags or labels for jobs.  Right now the only metadata we have about a job is the class and args, everything else is user-specific.

Just as processes can have labels displayed next to their entry in the Busy tab, it might be useful to show tags or labels on the different pages that show job listings. Of course we'd surface these things in the API also.

> NB: I don't want these features to power application logic.  Application code should not be iterating through queues or sorted sets to find labels / tags, this is for emergency repairs or Web UI usage only.

WDYT?

fatkodima:
Honestly, I can't figure out an example of a use case, which can not be replaced (with some hacks or not) by extending job payload. But it seems dirty for me, for example, to pass unused arguments to just show them in UI. Or, if, for example, job already accepts some `tenant_id` or `client_id`, I don't want to additionally pass `name` argument to just see name (instead of raw id) on ui.

Tags can be used for different use cases: to categorizing jobs, for example, by subdomain, tenant, country, locale, application, version, user/client, "alpha/beta/pro/ent", types of jobs, teams/people responsible for jobs, additional metadata (something like "do not stop this"), etc. 

This can be used as:
```ruby
class SomeApiWorker
  sidekiq_options queue: "api", tags: ["unreliable"]
  ...
end

# or
SomeApiWorker.set(tags: ["unreliable", "team_mc_teamface"]).perform_async(...)
```
Screen Shot 2019-09-22 at 02 41 37

Logs
```
2019-09-21T23:42:48.968Z pid=11430 tid=ovgymqh86 class=SomeApiWorker jid=e28bafeb235a5ab13c572342 tags=unreliable,team_mc_teamface INFO: start
2019-09-21T23:45:22.987Z pid=11430 tid=ovgymqh86 class=SomeApiWorker jid=e28bafeb235a5ab13c572342 tags=unreliable,team_mc_teamface elapsed=154.019 INFO: done
```

Possible considerations: implement tags as a hash instead of array; log tags as `tags=[foo, bar]` or `...[foo] [bar]` (like [ActiveSupport::TaggedLogging](https://api.rubyonrails.org/classes/ActiveSupport/TaggedLogging.html#method-i-tagged))

If this seems reasonable, will also add missing tests and changelog entry.
Closes #4073

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -5,6 +5,17 @@
 HEAD
 ---------
 
+- Support ad-hoc job tags. You can tag your jobs with, for example, subdomain, tenant, country, locale, application,
+  version, user/client, "alpha/beta/pro/ent", types of jobs, teams/people responsible for jobs,
+  additional metadata, etc.
+  Tags are shown on different pages with job listings [fatkodima, #4280]
+```ruby
+class MyWorker
+  include Sidekiq::Worker
+  sidekiq_options tags: ['foo', 'bar']
+  ...
+end
+```
 - Get scheduled jobs in batches before pushing into specific queues.
   This will decrease enqueueing time of scheduled jobs by a third. [fatkodima, #4273]
 ```
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -388,6 +388,10 @@ module Sidekiq
       Time.at(self["created_at"] || self["enqueued_at"] || 0).utc
     end
 
+    def tags
+      self["tags"] || []
+    end
+
     def error_backtrace
       # Cache nil values
       if defined?(@error_backtrace)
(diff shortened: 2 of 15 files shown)


mperham:
The implementation is clean and simple enough that I like it.  DataDog statsd metrics also have the notion of tags but I doubt there would be much confusion... I will probably make the tag clickable in Sidekiq Pro so you could automatically filter based on a tag value.

fatkodima:
> I will probably make the tag clickable in Sidekiq Pro so you could automatically filter based on a tag value.

So should I take this as a green light for this feature and add missing parts(changelog, tests) only?

mperham:
Yep. 👍

fatkodima:
Updated.

PR #4282, opened 2019-09-23
Allow `fetch` to take a `Range` object to help narrow the search for a job

hmcfletch:
Allow `score` to be a timestamp or a `Range` object. We have found this useful for helping us locate a job we know exists in a narrow time range.

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -596,18 +596,25 @@ module Sidekiq
       end
     end
 
+    ##
+    # Fetch jobs that match a given time or Range. Job ID is an
+    # optional second argument.
     def fetch(score, jid = nil)
+      begin_score, end_score =
+        if score.is_a?(Range)
+          [score.first, score.last]
+        else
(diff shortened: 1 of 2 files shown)


lib/sidekiq/api.rb
```diff
@@ -596,24 +596,30 @@ def each
...
-        end
+        data, job_score = element
+        entry = SortedEntry.new(self, job_score, data)
+        result << entry if jid.nil? || entry.jid == jid
       end
     end
 
     ##
     # Find the job with the given JID within this sorted set.
-    # This is a slower O(n) operation.  Do not use for app logic.
```
mperham: Did you mean to remove this comment?

hmcfletch: Nope. Was thinking about `zrangebyscore` for some reason here.

hmcfletch: add back in and rebased

mperham:
Well done, thank you!

PR #4285, opened 2019-09-24
Call run_afters even if throwing from run_befores

### LINKED ISSUE #4284: Dashboard after actions not run if raise/throw comes from a before action
I'd like to be able to use `Sidekiq::WebAction#redirect` in a block I've given to `Sidekiq::WebApplication::before` and still get the corresponding _afters_ run. Without this, `Thread.current[:sidekiq_redis_pool]` doesn't get set to `nil` when I redirect from a _before_ block because the `= nil` happens in an _after._ That was causing some bugs when I started sharding our queues redis (I know, I know). Maybe the right solution is for me to not use redirect in a _before_, but if you can accommodate it, I'd love to open a PR to change

```
resp = catch(:halt) {
  app = @klass
  self.class.run_befores(app, action)
  begin
    resp = action.instance_exec env, &action.block
  ensure
    self.class.run_afters(app, action)
  end

  resp
}
```
to
```
app = @klass
resp = catch(:halt) do
  self.class.run_befores(app, action)
  action.instance_exec env, &action.block
ensure
  self.class.run_afters(app, action)
end
```
so that throwing `:halt` in a _before_ doesn't cause `run_afters` to not be called.

jamesfischer8:
Closes #4284

### DIFF
--- a/lib/sidekiq/web/application.rb
+++ b/lib/sidekiq/web/application.rb
@@ -283,17 +283,13 @@ module Sidekiq
       action = self.class.match(env)
       return [404, {"Content-Type" => "text/plain", "X-Cascade" => "pass"}, ["Not Found"]] unless action
 
-      resp = catch(:halt) {
-        app = @klass
+      app = @klass
+      resp = catch(:halt) do
         self.class.run_befores(app, action)
-        begin
-          resp = action.instance_exec env, &action.block
-        ensure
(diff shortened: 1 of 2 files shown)


mperham:
Does this need a test?

jamesfischer8:
Good idea. Added one.

PR #4287, opened 2019-09-25
Support job-specific log levels

### LINKED ISSUE #4286: Job-specific log levels
Rails just added request-specific log levels. Sidekiq should support per-job log levels so you can turn on debugging for problematic workers. Investigate thread safe loggers.

fatkodima:
Closes #4286

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -5,6 +5,11 @@
 HEAD
 ---------
 
+- Support job-specific log levels. This will allow you to turn on debugging for
+  problematic workers. [fatkodima, #4287]
+```ruby
+MyWorker.set(log_level: :debug).perform_async(...)
+```
 - Support ad-hoc job tags. You can tag your jobs with, e.g, subdomain, tenant, country,
   locale, application, version, user/client, "alpha/beta/pro/ent", types of jobs,
   teams/people responsible for jobs, additional metadata, etc.
--- a/lib/sidekiq/logger.rb
+++ b/lib/sidekiq/logger.rb
@@ -17,8 +17,74 @@ module Sidekiq
     end
   end
 
+  module LoggerThreadSafeLevel
+    LOG_LEVEL_MAP = Hash[Logger::Severity.constants.map { |c| [c.to_s, Logger::Severity.const_get(c)] }]
+    LOG_LEVEL_MAP.default_proc = proc do |_, level|
+      Sidekiq.logger.info("Invalid log level: #{level.inspect}")
+      nil
+    end
+
+    Logger::Severity.constants.each do |severity|
+      define_method("#{severity.downcase}?") do
+        Logger.const_get(severity) >= level
+      end
+    end
+
+    def local_level
+      Thread.current[:sidekiq_log_level]
+    end
+
+    def local_level=(level)
+      case level
+      when Integer
+        Thread.current[:sidekiq_log_level] = level
+      when Symbol, String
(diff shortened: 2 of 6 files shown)


Changes.md
```diff
@@ -5,6 +5,18 @@
 HEAD
 ---------
 
+- Support job-specific log levels. This will allow you to turn on debugging for
+  problematic workers. [fatkodima, #INSERT_NUM_HERE]
```
mperham: INSERT_NUM_HERE

Changes.md
```diff
@@ -5,6 +5,18 @@
...
+```ruby
+class MyWorker
+  include Sidekiq::Worker
+  sidekiq_options log_level: :debug
```
mperham: Should we even document this usage pattern or will most people use it dynamically via `set` on the console?

lib/sidekiq/client.rb
```diff
@@ -224,6 +224,10 @@ def normalize_item(item)
       raise(ArgumentError, "Job class must be either a Class or String representation of the class name") unless item["class"].is_a?(Class) || item["class"].is_a?(String)
       raise(ArgumentError, "Job 'at' must be a Numeric timestamp") if item.key?("at") && !item["at"].is_a?(Numeric)
       raise(ArgumentError, "Job tags must be an Array") if item["tags"] && !item["tags"].is_a?(Array)
+      if item["log_level"] && !valid_log_level?(item["log_level"])
```
mperham: I don't see this validation as worth the code size.

lib/sidekiq/processor.rb
```diff
@@ -119,18 +119,20 @@ def dispatch(job_hash, queue)
...
-              # constantize the worker and instantiate an instance, we have to call
-              # the Reloader.  It handles code loading, db connection management, etc.
-              # Effectively this block denotes a "unit of work" to Rails.
-              @reloader.call do
-                klass = constantize(job_hash["class"])
-                worker = klass.new
-                worker.jid = job_hash["jid"]
-                @retrier.local(worker, pristine, queue) do
-                  yield worker
+          @job_logger.with_log_level(job_hash["log_level"]) do
```
mperham: Can you combine the log_level and hash_context setup into one method since the job_hash is already being passed in?

lib/sidekiq/logger.rb
```diff
@@ -17,8 +17,68 @@ def ctx
...
+    def local_level
+      Thread.current[:sidekiq_log_level]
+    end
+
+    def local_level=(level)
+      case level
+      when Integer
+        Thread.current[:sidekiq_log_level] = level
+      when Symbol, String
+        Thread.current[:sidekiq_log_level] = Logger::Severity.const_get(level.to_s.upcase)
```
mperham: I would setup a Hash with the N levels and the default proc would log a warning and return nil, rather than looking up constants.

fatkodima:
Updated.

jclusso:
@mperham this doesn't appear to be documented anywhere and I was wondering if this actually works as "envisioned" in #4286 or only via set at queue time.

mperham:
@jclusso Minor features sometimes get overlooked and don't make their way to the wiki. You can either add a section to the Logging wiki page yourself or open an issue if you can't get it working.

jclusso:
@mperham gotcha. It appears to work so I added it to the wiki.

Issue #2934, opened 2016-04-16
Suggestion: Easier way to test distinct Mailer jobs?


key88sf:
Currently testing of the delayed extensions for Mailers only looks at the job queue shared by all mailers. This means that if I have code that looks like:
```
if something
   UserMailer.delay.welcome_email(...)
end

if something_else
  AdminMailer.delay.new_signup(...)
end
```

I can't easily write a test that just checks a specific mailer using the suggested `expect {subject}.to change {Sidekiq::Extensions::DelayedMailer.jobs.size}.by(1)`.  Instead I have to either parse out the embedded mailer class from the job args, or maybe run sidekiq inline to actually deliver the mail.

It would be awesome if I could write an expectation for a *specific mailer*'s job queue size:
`expect {subject}.to change {UserMailer.jobs.size}.by(1) # something like this`

Any idea if that's easy/possible? Thanks!

mperham:
I don't know how you can test for a specific worker without have a specific worker for that mailer.

key88sf:
Well I assume it could just equate to a test helper method which checks the DelayedMailer.jobs queue perhaps? One could write this themselves as well, but I thought it could be useful to bake into the gem.

So essentially adding `DelayedMailer.jobs_for(UserMailer)` which returns all jobs for the UserMailer class? Just throwing it out as a possibility...

mperham:
Send me a proof of concept PR and maybe we can get something that works.

key88sf:
Basically it's this (not the prettiest):

testing.rb:
```
      def jobs_for(klass)
        jobs.select{|j| YAML.load(j["args"][0])[0] == klass}
      end

# Usage
# Sidekiq::Extensions::DelayedMailer.jobs_for(UserMailer)

```

PR #4292, opened 2019-09-29
Easier way to test enqueuing specific ActionMailer and ActiveRecord delayed jobs

fatkodima:
I have found this problem was discussed before in [2934](https://github.com/mperham/sidekiq/issues/2934)

I also disabled rubocop check (failing, but not relevant to this pr), because `ensure` can not be used inside curly braces. Personally, I don't see much value in that cop.

### DIFF
--- a/lib/sidekiq/testing.rb
+++ b/lib/sidekiq/testing.rb
@@ -323,6 +323,18 @@ module Sidekiq
       end
     end
   end
+
+  module TestingExtensions
+    def jobs_for(klass)
+      jobs.select do |job|
+        marshalled = job["args"][0]
+        marshalled.index(klass.to_s) && YAML.load(marshalled)[0] == klass
+      end
+    end
+  end
+
+  Sidekiq::Extensions::DelayedMailer.extend(TestingExtensions)
+  Sidekiq::Extensions::DelayedModel.extend(TestingExtensions)
 end
 
 if defined?(::Rails) && Rails.respond_to?(:env) && !Rails.env.test?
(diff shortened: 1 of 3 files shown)


mperham:
Should the other test APIs have the same jobs_for helper?

fatkodima:
Does it make sense, if yes? To me this method looks like `Extensions` specific.

mperham:
You’re right, I misunderstood.

lib/sidekiq/testing.rb
```diff
@@ -323,6 +323,15 @@ def drain_all
       end
     end
   end
+
+  module TestingExtensions
+    def jobs_for(klass)
+      jobs.select { |job| YAML.load(job["args"][0])[0] == klass }
```
mperham: I would add a quick check to prune obviously wrong values and avoid parsing the entire YAML if not necessary, e.g.

```ruby
jobs.select { |job| job["args"].index(klass.to_s) && YAML.load(job["args"][0])[0] == klass }
```

What do you think?

lib/sidekiq/web/application.rb
```diff
@@ -284,7 +284,7 @@ def call(env)
       return [404, {"Content-Type" => "text/plain", "X-Cascade" => "pass"}, ["Not Found"]] unless action
 
       app = @klass
-      resp = catch(:halt) do
+      resp = catch(:halt) do # rubocop:disable Standard/SemanticBlocks
```
mperham: Can you submit this as a separate change?

lib/sidekiq/testing.rb
fatkodima: Good idea. But `job["args"]` is an array, so changed code a bit.

PR #4305, opened 2019-10-02
Extended translations for german

juril33t:
Added translations for strings that were present in [en.yml](https://github.com/mperham/sidekiq/blob/master/web/locales/en.yml) but not in de.yml

Changed "töten" into "vernichten", because "töten" literally means to take someone's life. As a native speaker, I think "vernichten" sounds more appropriate here.


mperham:
Drat, I just released 6.0.1 less than an hour ago.  Oh well, 6.0.2!

PR #4306, opened 2019-10-02
Improve fault tolerance of rescheduling jobs using API

fatkodima:
This eliminates probability of job being lost when `delete` succeeds but `schedule` fails, and reduces redis roundtrips to 1.

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -471,8 +471,9 @@ module Sidekiq
     end
 
     def reschedule(at)
-      delete
-      @parent.schedule(at, item)
+      Sidekiq.redis do |conn|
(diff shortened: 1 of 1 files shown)


mswart:
This change introduced a slight behavior change as the original called [schedule](https://github.com/mperham/sidekiq/blob/4803d87969d88153cbaf178ebbfec3d0da23a5da/lib/sidekiq/api.rb#L574) method did a type conversion (`to_f.to_s`) and the current implementation does not.
Our tests failed after the 6.0.2 upgrade as it is no longer supported to call it with a DateTime object.

mperham:
Yep, looks like we need `at.to_f`.

PR #4309, opened 2019-10-04
Fixed: outdated upstart script examples

lizdeika:
Sidekiq 6 removed `-i` option so current upstart examples are not working.


mperham:
This fix is wrong. You only need to edit the bundle exec line. Upstart supports indexed services for starting multiple copies of a process and we want to use that.

fatkodima:
Not directly relevant to this pr (but this already contains a plenty of commits), but seen in previous (as well as mine), @mperham you probably should configure github's merge button to merge with squashing to make history cleaner for future debuggings and references.

mperham:
Thanks for the reminder, merge button updated.

PR #4331, opened 2019-10-10
Add bug report templates

fatkodima:


mperham:
I don't really see this as necessary and don't want to maintain it years from now.

fatkodima:
This definitely would help to get better bug reports and to get them faster, instead of constantly asking bug reporters to provide ones (some of them maybe not even understand, what do you want from them or how to write one).

I do not see any maintenance needs for this files, as their content will not need to change for years with 95% probability.

PR #4340, opened 2019-10-16
Make the redis server version error clearer

ammar:
Make is clearer that the redis version error message refers to the version of the server, not the redis gem.


mperham:
This might be my first PR from Palestine?  Thank you!

PR #4359, opened 2019-11-04
Use MultiJson to dump/load json, so users can control which json coder is used

knapo:
It would be really great, if users could control which json coder is used. Using `MultiJson` gives this flexibility.

I personally have various issues with default `JSON` coder, and prefer to use `oj`, which is faster and can handle encoding issues and work on non UTF-8 strings.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,8 +1,9 @@
 PATH
   remote: .
   specs:
-    sidekiq (6.0.2)
+    sidekiq (6.0.3)
       connection_pool (>= 2.2.2)
+      multi_json
       rack (>= 2.0.0)
       rack-protection (>= 2.0.0)
       redis (>= 4.1.0)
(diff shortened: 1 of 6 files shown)


mperham:
You need to make a better pitch if you want to make this fundamental of a change.

mperham:
For example, the ex maintainer of multijson does not recommend its use anymore. https://github.com/rails/rails/pull/10576

knapo:
> You need to make a better pitch if you want to make this fundamental of a change.

Well, you should write a few more words, not just closing PR saying that I didn't elaborate enough.

> For example, the ex maintainer of multijson does not recommend its use anymore. rails/rails#10576

Thanks for pointing this out.  I wasn't aware about `multi_json` performance issues. In such case, I agree it doesn't makes sense to bring it to `sidekiq`. I have another solution in mind, less fundamental, and coming with a next PR.

PR #4360, opened 2019-11-04
Allow to use different json coders through json_loader and json_dumper config options.

knapo:
Alternative solution to fix problem described in https://github.com/mperham/sidekiq/pull/4359.

It allows users to use different json coders by specifying  `json_loader` and `json_dumper` through configuration options. Default values are set to `JSON` coder's methods (`JSON.parse` / `JSON.generate`).

Example to use different engine:
```ruby
Sidekiq.options[:json_loader] = ->(string) { Oj.load(string) }
Sidekiq.options[:json_dumper] = ->(object) { Oj.dump(object) }
```

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -35,6 +35,8 @@ module Sidekiq
     dead_max_jobs: 10_000,
     dead_timeout_in_seconds: 180 * 24 * 60 * 60, # 6 months
     reloader: proc { |&block| block.call },
+    json_loader: proc { |string| JSON.parse(string) },
+    json_dumper: proc { |object| JSON.generate(object) },
   }
 
   DEFAULT_WORKER_OPTIONS = {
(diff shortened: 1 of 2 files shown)


mperham:
What about monkeypatching Sidekiq.{load/dump}_json in your own app? That's why those methods exist.

fatkodima:
Similar code changes were proposed several times in the past and were rejected.

knapo:
I stopped considering monkey patching as a good practice years ago and wanted to come with a more elegant solution.

mperham:
I still haven’t heard any compelling reasons why stdlib JSON doesn’t work for you. I’m against pointless abstraction.

knapo:
It fails on invalid encoding strings, when `oj` can handle them with no issues. 

e.g. 
```
2.6.5 :049 > JSON.generate({foo: bar})
...
Encoding::InvalidByteSequenceError ("\xE2" on US-ASCII)
2.6.5 :050 > Oj.dump({foo: bar})
 => "{\"foo\":\"Bar\xE2\"}" 
```

mperham:
I consider that a feature: don't propagate bad data across system boundaries.

knapo:
Me either, but sometimes it to complex and not that easy to achieve.

Issue #4343, opened 2019-10-17
Feature Request: Add Pause button to Sidekiq UI


jakemack:
Having the ability to pause a queue without launching a Ruby shell of some sort seems like a no brainer. It would be particularly useful to pause or unpause a misbehaving queue from a phone when a computer is not easily accessible. I had just such an issue last night where I was out and one queue with newly deployed jobs was draining resources from all other jobs. Simply pausing the queue allowed everything else to proceed as usual until the offending job was fixed, but it all had to wait until I could find a computer.

Would you be open to a PR for this?

mperham:
My belief is that if you want to pause a queue, there's usually a more involved process that needs to happen (alert emails sent, maybe a Slack warning, add a security or audit trail, etc).  You should build your own UI with a button to enable that entire process and use the `Sidekiq::Queue#pause!` API in that button handler.

jakemack:
Ideally, that would be the case, but unfortunately the real world isn't ideal. Such a button is useful for the emergencies where a process hasn't been built around an issue yet. Ideally, one would have time to build that process after the first such emergency but unfortunately, not all businesses cooperate with that logic. It could be that the cost of building such a system around it is not worth the extra cost to the business when a manual, ad-hoc pause suffices.

In addition, a number of the other features in the Web UI also follow such logic: usually there should be a more involved process for quieting or stopping a given worker, there should likely be a more involved process around deciding to run a scheduled job earlier than scheduled, etc. Apart from a few easily viewable metrics, the Web UI is really a collection of conveniences for ad-hoc manipulation of Sidekiq. Adding one more seems to fit right in with the theme. And until you can say there is _always_ a more involved process around an action, there will be situations where an ad-hoc, manual action is appropriate.

mperham:
Ok, I can be flexible here and willing to consider a PR if you want to put something together. Take a look at the Filtering code if you want to see a similar feature.

jakemack:
Awesome, I'll see what I can put together

shayonj:
Came across this multiple times (operationally, mostly) in the past and every time I felt it'd be nice to have this feature. So, I whipped a quick PR :). There are some open questions in the PR as to supporting this behavior cleanly between regular version of sidekiq and sidekiq pro.

mperham:
Now available in Sidekiq 6.0.4.

PR #4374, opened 2019-11-13
Introduce the ability to pause queue from the UI

shayonj:
Considering sidekiq web UI allow to perform complex tasks with ease.
Having the ability to pause a queue during an emergency (or operational intense)
scenarios would come in quite handy.

This PR introduces a new button to under the "Actions" column on `sidekiq/queues` pages,
right next to the `Delete` button. Depending on the state of the queue, it will according
show the `Pause` or `Unpause` text, with appropriate form element that `POST`s to the same
endpoint. The endpoint logic is updated to handle the new cases. Added some relevant unit tests.

Adding mock `Sidekiq::Queue#pause!` and `Sidekiq::Queue#unpause!` functions to
the API, which I believe will be overridden by sidekiq pro.

Original request: https://github.com/mperham/sidekiq/issues/4343

**Examples**
image
image

**Open questions**
- [ ] `Sidekiq::Queue#pause!` is a feature on sidekiq pro. Showing the `pause` button in the UI on regular sidekiq could be misleading, since it would be a no-op? 
  - Maybe it should be hidden through some custom check/feature-flag-esque behavior? And/or ported to sidekiq pro?
- [ ] Added new copy to `en.yml`, translations would need to be accordingly handled?

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -242,17 +242,9 @@ module Sidekiq
       Sidekiq.redis { |con| con.llen(@rname) }
     end
 
-    # Sidekiq Pro overrides pause?, pause! and unpause! functions
+    # Sidekiq Pro overrides this
     def paused?
       false
-    end
-
-    def pause!
-      nil
-    end
-
-    def unpause!
-      nil
     end
 
     ##
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -242,9 +242,17 @@ module Sidekiq
       Sidekiq.redis { |con| con.llen(@rname) }
     end
 
-    # Sidekiq Pro overrides this
+    # Sidekiq Pro overrides pause?, pause! and unpause! functions
     def paused?
       false
+    end
+
+    def pause!
+      nil
+    end
+
+    def unpause!
+      nil
     end
 
     ##
--- a/web/locales/en.yml
+++ b/web/locales/en.yml
@@ -24,9 +24,13 @@ en: # <---- change this to your locale code
   ShowAll: Show All
   CurrentMessagesInQueue: Current jobs in <span class='title'>%{queue}</span>
   Delete: Delete
+  Pause: Pause
+  Unpause: Unpause
   AddToQueue: Add to queue
   AreYouSureDeleteJob: Are you sure you want to delete this job?
   AreYouSureDeleteQueue: Are you sure you want to delete the %{queue} queue?
+  AreYouSurePauseQueue: Are you sure you want to pause the %{queue} queue?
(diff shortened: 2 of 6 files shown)


shayonj:
cc @mperham

web/locales/en.yml
```diff
@@ -24,9 +24,13 @@ en: # <---- change this to your locale code
   ShowAll: Show All
   CurrentMessagesInQueue: Current jobs in <span class='title'>%{queue}</span>
   Delete: Delete
+  Pause: Pause
+  Unpause: Unpause
   AddToQueue: Add to queue
   AreYouSureDeleteJob: Are you sure you want to delete this job?
   AreYouSureDeleteQueue: Are you sure you want to delete the %{queue} queue?
+  AreYouSurePauseQueue: Are you sure you want to pause the %{queue} queue?
```
mperham: We don't need confirmation for these actions.

web/locales/en.yml
```diff
@@ -24,9 +24,13 @@ en: # <---- change this to your locale code
   ShowAll: Show All
   CurrentMessagesInQueue: Current jobs in <span class='title'>%{queue}</span>
   Delete: Delete
+  Pause: Pause
```
mperham: Please put new strings at the bottom of the file so other languages can see an easy diff of which strings are missing.

lib/sidekiq/api.rb
```diff
@@ -242,11 +242,19 @@ def size
       Sidekiq.redis { |con| con.llen(@rname) }
     end
 
-    # Sidekiq Pro overrides this
+    # Sidekiq Pro overrides pause?, pause! and unpause! functions
     def paused?
       false
     end
 
+    def pause!
```
mperham: What happens if the user isn't using Sidekiq Pro?  Should these methods really do nothing?  Should they raise an error or log a warning?  Should the button be disabled in the Web UI with an explanatory tooltip if `!defined?(::Sidekiq::Pro)`?

shayonj: That's good question, I left it under open question so we could discuss some possible options. I am currently of the opinion that we roll it out with minimal change to the regular sidekiq and web (backwards compatible) and as a feature on sidekiq pro. That means, we simply hide the button when pro is not present from the UI and accordingly add proper validation in the api router. Raising exception or tooltips would result in unnecessary friction that didn't exist before. Of course, can be persuaded :). 

I introduced this as a new commit. Its not a huge change. Used `defined?(::Sidekiq::Pro)` as a check to determine whether or not pro is present.

mperham:
Thank you, I'm very happy with this.

shayonj:
woot! thanks for the quick feedback and reviews 🙌

jakemack:
Thanks for picking up my slack here, @shayonj. Awesome work!

PR #4378, opened 2019-11-14
Allow enable/disable Web dark mode

ritxi:


seuros:
`prefers-color-scheme: dark ` do the trick.

If your computer is in dark mode, you will have dark mode in the ui.

PR #4407, opened 2019-12-23
Add Japanese locale texts for pause / unpause

ryohashimoto:
Hi, I added Japanese locale texts for pause / unpause for the Web UI button introduced in #4374.


mperham:
Thank you!

PR #4410, opened 2019-12-28
Removing fetcher and stopped? methods

RicardoTrindade:
Hey, I noticed that the `fetcher` attr_accessor in launcher.rb seems to not be used anymore. I'm proposing to remove it along with the `stopped?` in `manager.rb` method which is not used as well.


mperham:
Given this is an undocumented but public API and there isn’t a compelling reason to remove, I’m reluctant to merge this.

RicardoTrindade:
Completely understand your point, will close it then

PR #4412, opened 2019-12-31
Silence Ruby 2.7 warnings

kddeisz:
Since you can pass kwargs to Logger.new, you need to explicitly handle them in the subclass otherwise you get a warning.


mperham:
🎵 Livin' on the edge! 🎶

Thanks!

PR #4424, opened 2020-01-11
Fix Rack::File deprecation

### LINKED ISSUE #4423: Rack::File deprecation warning
Ruby version: 2.6.3
Sidekiq version(s):  6.0.4

### Issue

Warning displayed:
`Rack::File is deprecated, please use Rack::Files instead.`

### Reproduction code

```ruby
require 'sidekiq/web'
puts "ok"
```

### Output

```
$ ruby test.rb
Rack::File is deprecated, please use Rack::Files instead.
ok
```

### Additional info

- The warning was introduced in Rack 2.1.0 - ref: https://github.com/rack/rack/blob/2.1.0/lib/rack/file.rb

DannyBen:
Closes #4423

### DIFF
--- a/lib/sidekiq/web.rb
+++ b/lib/sidekiq/web.rb
@@ -14,7 +14,7 @@ require "sidekiq/web/application"
 require "rack/protection"
 
 require "rack/builder"
-require "rack/file"
+require "rack/files"
 require "rack/session/cookie"
 
 module Sidekiq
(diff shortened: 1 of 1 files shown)


mperham:
I noted that Rack master has already reverted this deprecation.

DannyBen:
They just removed the warning. `File` is an alias to `Files` since 2012.
https://github.com/rack/rack/pull/305

Should be changed, unless there are evidence that state that `Rack::File` is superior and preferred over `Rack::Files` - I did not find any.

mperham:
I reverted this as it breaks Rack <2.0.8.

mperham:
I would be happy to bump to Rack 2.1+ in Sidekiq 6.1 but I don't want to do it in master for a few weeks until this Rack mess stabilizes.

DannyBen:
Alright, fair enough.

PR #4434, opened 2020-01-16
Add rvm instruction for sidekiq service

the-spectator:
+ This PR adds instructions for using rvm with sidekiq service.


mperham:
Don’t use rvm in production.

mperham:
Running rvm in production is not a community best practice and therefore I don't document it nor encourage it. RVM was useful back before Bundler but chruby is a safer, simpler alternative.

PR #4439, opened 2020-01-28
update version 6 upgrade notes about setting log level

jjb:
noticed this while upgrading

Let me know preferred copy (or just nuke this PR and add directly).


mperham:
That has never been the recommended way to configure the logger so that’s why it’s not in the notes. The recommended way still works.

```ruby
Sidekiq.logger.level = Logger::WARN
```

PR #4449, opened 2020-02-05
Use `unlink`, not `del`, for performance

mperham:
This makes deleting/clearing large queues much faster.


mperham:
This requires Redis 4.0+ which is a requirement for Sidekiq 6.0+

Issue #4409, opened 2019-12-28
[Request] Add config option to force light/dark mode


CHTJonas:
It would be nice to be able to override the CSS `@media` queries for light/dark mode so that, regardless of the user's theme choice in their OS, a specified theme is always served. This could take the form of a config option in `config/sidekiq.yml` for example:

```YAML
---
:theme: light
```

This addresses some of the tail comments from #4231.

mperham:
I disagree with the premise, the user should have control over the UI and theme.

heldopslippers:
Sorry to get into this again. 
There are also teams that really don't like the dark Sidekiq theme, but still are using the "dark" theme on their computers. 
Why would it be so bad to make it possible to be able to force a default if somebody once to.

mperham:
I'd love to have CSS PRs which make the dark theme better. I'm not interested in a switch which forces light/dark on all users.

PR #4231, opened 2019-08-27
Initial pass at dark mode support, courtesy of @silent-e

mperham:
See #4227 for more discussion.

PR #4453, opened 2020-02-07
Add web theme (light/dark) configuration in sidekiq.yml

kihaya:
Please See: https://github.com/mperham/sidekiq/issues/4409

### background
https://github.com/mperham/sidekiq/pull/4231 enables sidekiq web ui to change the theme depending on OS appearence setting (using media query). However, currently no way to force user to see specific theme of web ui (https://github.com/mperham/sidekiq/pull/4231#issuecomment-542296663).

### What
This PR enable sidekiq to configure web theme (light  or dark) in sidekiq.yml. Those who (OS) does not support `prefers-color-scheme` can now enjoy the dark mode (it ease to eyes).


mperham:
I think this is not a good idea. You force dark upon all who want the light theme.

kihaya:
@mperham 

> You force dark upon all who want the light theme.

It's sorry for my poor explanation. Could you tell me for which part do you think so ?

This PR intends to:
+ Enable to switch light or dark theme by configuration.
+ If no specific configuration given, theme follows os settings as usual.

mperham:
Yes, but if the config says "dark", the Web UI will always use "dark", even if the user's environment wants to use "light".

damuz91:
@mperham but shouldn't the user config override what the environment 'wants'?

mperham:
@damuz91 I'm confused. The user's config in their OS / browser will control the color scheme used, always. Remember, teams of people use the Sidekiq UI. One person might want dark, another might want light. The admin setting up sidekiq.yml should have no say in the matter.

jcquarto:
I find this confusing, very much so.  It's one thing to have Sidekiq match a user's OS theme when there's no explicit setting -- which is what a team can do if they want their members to see it. But have you had anyone over the age of 40 look at your dark theme?  red text on black?  This is madness. There absolutely should be a mechanism by which a developer can force a theme on his own machine without having to switch the OS color scheme.  Maybe let the YML file  take a ENV variable ? that way, in development I will have my own ENV file anyway, and I can set the theme however i like it. 

This reminds me of a story from Amazon way way way back in the day. Like, 2001/2002. I bought a book for a relative on Gardening. Thereafter, despite the fact that I never ever had bought a gardening book previously or since, Amazon kept suggesting gardening books to me.   A few years later that algorithm change, thank god.  It a fine example of developers optimizing for something that is none of their business just for the sake of being cool.   It's not Sidekiq's job to care what theme my OS has. Let me worry about that. If you are going to enforce that bit of your-opinion-on-good-design on me, at least make it easy for me to enforce my-own-opinion-on-good-design on myself.

Either that, or I completely have misunderstood whether we can set the theme or not.  Can't find much in the way of docs, though I do see a lot of smoke on the topic. (I'm kinda aghast that any of us are spending any mental cycles kvetching on this no-brainer topic of theme. This isn't WordPress.).  In the meantime, the dark theme is , just IMHO,  a poor choice. I would love for Mike to build me a house; that doesn't mean I want him to pick the paint on my walls.

seuros:
@jcquarto , are you aware you can assemble own house and even have the wall transparent glass if you wish ?

In other word, Mike decided make that dashboard as simple as possible. If your theme is dark, it will be dark, if your theme is light, it will be light.

If you want customization, you can mount your own copy of engine and customize it or just interact with the API directly.
You have to support any extra feature manually then, since you are not following the convention.

mperham:
As a guy who is over 40 years old, I would welcome any PRs with CSS improvements to the dark theme.￼

mperham:
#4674

mperham:
^^^ That took me one hour. @jcquarto Want to try that PR and give me feedback?

PR #4481, opened 2020-03-06
Remove usage of setInterval

### LINKED ISSUE #4480: Ajax uses setInterval instead of setTimeout, leading to server load
Thanks for a great product

Ruby version: 2.6.3
Sidekiq / Pro / Enterprise version(s): Sidekiq 5.2.8 (code for 6.0.5/master on github contains the same issue)

Initializer and error message: not applicable

### Detail:

The javascript at https://github.com/mperham/sidekiq/blob/master/web/assets/javascripts/application.js#L58 uses `setInterval`. This queues a query to the backend via ajax irrespective of whether the last AJAX request has responded or not.

If there are problems with the infrastructure running the Sidekiq web ui (eg, overloaded web servers), this will make those problems worse - since ajax requests will be queued continuously and increase the established connections to the web ui host.

The general practice is to instead use `setTimeout` to queue a new request on completion of the last request. See https://makitweb.com/how-to-fire-ajax-request-on-regular-interval/ for more information.

### Why:

Unfortunately, we had a large number of items incorrectly queued with large message bodies (exceeding the size listed at https://github.com/mperham/sidekiq/blob/455e9d56f46f0299eaf3b761596207e15f906a39/lib/sidekiq/extensions/generic_proxy.rb#L7)

In trying to access the web page, clicking `poll` would bring the management app down. This is because running `lrange queue:low 0 24` would take 6 seconds to respond. In the time it took the first ajax request to respond another 3 ajax requests were queued - which then took 6 seconds to respond each. In short order, this exceeded the number of connections on the web server and it restarted when the load balancer removed it.

mperham:
Use of setInterval can lead to thundering herd under heavily loaded
conditions, fixes #4480.

### DIFF
--- a/web/assets/javascripts/application.js
+++ b/web/assets/javascripts/application.js
@@ -19,7 +19,8 @@ Sidekiq = {};
 $(function() {
   var pollpath = $('body').data('poll-path');
   if (pollpath != "") {
-    updatePage(pollpath);
+    var ti = parseInt(localStorage.timeInterval) || 2000;
+    setTimeout(function(){updatePage(pollpath)}, ti);
   }
 
   $(document).on('click', '.check_all', function() {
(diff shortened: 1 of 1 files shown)


mperham:
Appears to work fine with a simple local smoke test.  Can any JS-savvy people confirm?

andris9:
If ajax call fails (eg. network issues) then `done()` is never called and your timer chain breaks. You probably would want to add `fail()` handler as well to manage that.

mperham:
Thank you, great feedback, added.

PR #4487, opened 2020-03-12
Backwards compatibility for options in rails 5.1

nicholasjstock:
This commit https://github.com/rails/rails/commit/0e64348ccaf513de731f403259ec5b49e7b3f028
 moves from passing a string in the wrapper key to passing a class.
Therefore in rails 5.1 the following does not show a backtrace in the sidekiq UI

```
class FuckedJob < ApplicationJob
  sidekiq_options retry: 0,  backtrace: 50
  def perform
    raise "NOOOOOOOOOO"
  end
end
```

We are a bit weary of moving to rails 6 on this one client project because code coverage isn't all that great yet. 
Maybe someone else has the same problem so here you go!


mperham:
This solution breaks at least two edge cases that I can think of.  This feature can't be backported or else I would have.

PR #4489, opened 2020-03-14
systemd notification integration

### LINKED ISSUE #4488: Add systemd notify and watchdog support
As Sidekiq is meant to be a long-running daemon, it should support the notification and watchdog functionality that systemd allows.

* [x] Add READY=1 notification upon startup
* [x] Add STOPPING=1 notification upon terminate
* [x] Add WATCHDOG=1 loop

See also https://github.com/agis/ruby-sdnotify
https://www.freedesktop.org/software/systemd/man/sd_notify.html

mperham:
Fixes #4488

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -5,6 +5,8 @@
 HEAD
 ---------
 
+- Integrate with systemd's watchdog and notification features [#4488]
+  See `Type=notify` in [systemd.service](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Options)
 - Fix edge case where a job can be pushed without a queue.
 
 6.0.5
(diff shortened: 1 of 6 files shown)


misdoro:
Nice, glad to see it in the standard sidekiq setup :heart:

agis:
Hey @mperham, it's great to see sidekiq leveraging sd_notify. I'm curious, how come you didn't use the actual gem instead? Would be interested to know the reasoning. Thanks!

mperham:
@agis The actual implementation is one .rb file and quite simple. I prefer to minimize 3rd party dependencies as it makes long-term maintenance much easier. Notice that Sidekiq only depends on the `rack` and `redis` gems (and connection_pool but that's my gem too).

PR #4491, opened 2020-03-15
Prefer Hash#transform_keys & Regex#named_captures

the-spectator:
+ Uses `Hash#transform_keys` to stringify.
+ Uses `Regex#named_captures` to get hash of captured data.


the-spectator:
@mperham Thank you for the merge!

PR #4500, opened 2020-03-23
Use media attribute for prefers-color-scheme

ykzts:
Use the media attribute of the link element for prefer-color-scheme. This will reduce unnecessary HTTP requests.

see https://web.dev/prefers-color-scheme/


mperham:
Nice fix!

Issue #4498, opened 2020-03-19
Jobs processed during shutdown not appearing in stats 


ouranos:
Ruby version: ruby-2.5.1
Sidekiq version: 6.0.5

Hi, 
To test deployment and graceful restart of Sidekiq workers, I've created a dummy jobs that just sleeps and output to stdout:
```ruby
class SleepyWorker < BaseWorker
  SLEEP = 30

  def perform steps = 10
    steps.times do |i|
      puts "Sleeping for #{SLEEP}s (#{i}/#{steps})"
      sleep SLEEP
    end
    puts "I'm done"
  end
end
```

My initializer is pretty standard:
```ruby
redis_config = { url: ENV['REDIS_SIDEKIQ_URL'] }

# Use redis authentication if available
redis_config[:password] = ENV['REDIS_SIDEKIQ_PASSWORD'] if ENV['REDIS_SIDEKIQ_PASSWORD'].present?

Sidekiq.configure_server do |config|
  config.redis = redis_config
end

Sidekiq.configure_client do |config|
  config.redis = redis_config
end
```

I've increased the shutdown timeout to 8 minutes in `config/sidekiq.yml`:
```yml
:timeout: 480 # 8m, default is 25s
```

I'm stopping Sidekiq with the `TERM` signal after having enqueued a 5min sleeping job, I'm expecting:
- Sidekiq to stop accepting new jobs -> worker shows as quiet in the UI ✅ 
- Sidekiq to wait for the job to be done before shutting down ✅ 
- The numbers of processed jobs to be increased by 1 => Busy is at 1 while processing but Processed is not increased :x:

Am I missing something?

mperham:
The job is not done so it is not considered processed yet. Right?

ouranos:
It's actually finishing processing as it takes shorter than the 8 minutes timeout (and my actual job update a record in the db at the end which is correctly updated)
So processing is fine, it's more an issue with the stats reporting I think.

mperham:
Did you refresh the Web page?  You’re right, I’d expect that to work. A terminating process still reports its stats every five seconds.

ouranos:
Yes, I've tested multiple times as I thought it was an issue with my restart process first.

The logs are:
```
2020-03-19 00:17:22,005 DEBG 'sidekiq_worker' stdout output:
2020-03-19T00:17:22.005Z pid=43 tid=go8g0fhej INFO: Pausing to allow workers to finish...
...
[...]

PR #4502, opened 2020-03-23
Do not connect to redis at exit if not needed

misdoro:
A fix for https://github.com/mperham/sidekiq/issues/4498 introduced a callback that is executed at ruby VM exit and tries to connect to redis.

This callback is also executed in CI runs and in all other cases where sidekiq is loaded but not used with real redis.

As a work-around, keep the at_exit hook but communicate to redis only if there is something to send, i.e. the number of processed or failed jobs is greater than zero

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -5,6 +5,7 @@
 HEAD
 ---------
 
+- Fix: Do not connect to redis at ruby vm exit when not needed. [#4502] 
 - Remove Redis connection naming [#4479]
 
 6.0.6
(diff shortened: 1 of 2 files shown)


misdoro:
@mperham WDYT ?

mperham:
The root problem for many people is that they are using the `sidekiq-cron` gem and it has a bug where it is loading some Sidekiq internals (`sidekiq/launcher`) when it should not.

aaronjensen:
Ah, we are using that, but we're also enterprise, so I'll work on moving us over to enterprise cron.

PR #4528, opened 2020-04-11
Add Vietnamese locale for web

luong-komorebi:


mperham:
This is great, thank you. The Sidekiq Pro and Enterprise locales are in the issues if you want to translate those, they are pretty small.

luong-komorebi:
We are using Sidekiq Pro at our Org. Will try to look at the issue and find a way to contribute. Thank you, Mike

Issue #4440, opened 2020-01-29
Truncated response body using Sidekiq dashboard w/New Relic


pelargir:
Ruby version: 2.6.5p114
Sidekiq / Pro / Enterprise version(s): 6.0.4

We're using newrelic_rpm 6.8.0.360 to add JavaScript instrumentation to our Rails 5.2.4.1 app. After upgrading to Rack 2.1.2 the dashboard went blank. Viewing the page source, the HTML stops in the middle of the New Relic JS code near the top of the page (see screenshot).

Apparently, Rack::Response no longer overwrites the Content-Length header, which means an incorrectly configured Content-Length header is now exposed. See https://github.com/rack/rack/issues/1520#issuecomment-579838704

Is it possible Sidekiq is setting Content-Length prior to New Relic adding its instrumentation, resulting in an incomplete page load since the New Relic code takes up the entire Content-Length before the remaining body can be sent?

As soon as I remove New Relic from our app the dashboard starts working again.

![view-source_https___www_candlescience_com_admin_sidekiq](https://user-images.githubusercontent.com/3756/73385068-772c2500-429a-11ea-8406-8bfa85596b5e.png)

mperham:
Should NewRelic be instrumenting Sidekiq's pages at all?  Maybe it should exclude /sidekiq/*.

See https://github.com/mperham/sidekiq/blob/3510de7df3c2eaacec9061184666aa853a0abb13/lib/sidekiq/web/application.rb#L317

Added here with no explanation:

https://github.com/mperham/sidekiq/commit/9ae228c521699eee43d7b2c371ecb86ca28468f9

I just tested removing those two lines (315 and 317) and the Sidekiq UI appears to work without them in both Rack 2.0.8 and 2.1.2. Unfortunately I have no idea whether this is a good idea or not.

pelargir:
Yes, New Relic Browser instrumentation should probably be excluded from Sidekiq::Web by default. They give examples of how to do this in a Rails controller and a Sinatra app, but not a pure Rack app. I'm thinking a middleware might work.

pelargir:
Still trying to turn off New Relic for Sidekiq::Web using this:
https://docs.newrelic.com/docs/agents/ruby-agent/api-guides/ignoring-specific-transactions#page-load-timing-rum

In the meantime, to get running I've created a monkey patch that deletes the Content-Type header from responses coming from Sidekiq::Web. Hope this helps someone:

```ruby
require 'sidekiq/web'

[...]

PR #4541, opened 2020-04-22
Ensure Rack::ContentLength is loaded as middleware

f3ndot:
6.0.5 removed `Sidekiq::WebApplication`'s own Content-Length header calculation, expecting Rack to do it instead via some middleware (https://github.com/mperham/sidekiq/commit/675e40c32684a7e89987673b82b5241545763fcb). This was to fix an issue where New Relic would inject content into Sidekiq's Web UI after the hand crafted content length was calculated (#4440). While such middleware exists, it's not enabled by default (at least in a Rails app) and so the HTTP response doesn't have the header at all.

The absence of the header has caused unexpected behaviour in an nginx and HAProxy setup we use in production systems, causing each page load to take exactly 20 seconds. The suspicion is that, without the Content-Length response header, HAProxy waits a preset amount of time before deciding the body must have surely by finished and then responds.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -5,6 +5,7 @@
 Unreleased
 ---------
 
+- Ensure `Rack::ContentLength` is loaded as middleware for correct Web UI responses [#4541]
 - Avoid exception dumping SSL store in Redis connection logging [#4532]
 
 6.0.7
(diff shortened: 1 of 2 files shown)


mperham:
Please add a changelog entry.

f3ndot:
@mperham all set. Thanks for your review.

mperham:
The next version of Sidekiq will revert this as I see Content-Length as orthogonal to the Web UI. I believe Rails will handle it automatically. If you are running the Web UI directly, you can ‘use Rack::ContentLength’ beforehand. If neither of these is applicable, let me know and we can figure out something.

PR #4532, opened 2020-04-18
Remove SSL parameters from Redis connection logging to avoid exception

geoffharcourt:
In 3f9c4bf9 the Redis connection options began to be cloned (via dumping
and re-marshalling) to avoid issues with password redaction in logging
altering the connection options and breaking authentication with Sentinels.

Unfortunately, this change caused an exception on boot for users of
Redis over SSL. The `OpenSSL::X509::Store` object used for SSL certs is
not yet dumpable in the bundled OpenSSL wrapper for current Rubies
(although it does in master as of ruby/openssl#281).

The fix here prunes the `ssl_params` options out of the Redis
configuration options before the dumping and marshalling. It's probably
better not to include those in logging anyway for privacy purposes.

Fix #4531

PR #4545, opened 2020-04-29
Exclude :ssl_params from Redis options clone without ActiveSupport

geoffharcourt:
In dd0a8476 we started filtering out SSL params from the Redis options
cloning for logging purposes to prevent an undumpable
`OpenSSL::X509::Store` from raising an exception. The fix used
ActiveSupport's `Hash#except` to exclude the `:ssl_params` key from the
hash before cloning the options hash by dumping and reloading.

As ActiveSupport is not a requirement for Sidekiq, we need a more
generally-available solution. This change removes the key using
`Hash#slice` by including every top-level options key _except_
`:ssl_params`.

See https://github.com/mperham/sidekiq/pull/4532#issuecomment-620894721

cc: #4532

### DIFF
--- a/lib/sidekiq/redis_connection.rb
+++ b/lib/sidekiq/redis_connection.rb
@@ -97,7 +97,12 @@ module Sidekiq
         redacted = "REDACTED"
 
         # deep clone so we can muck with these options all we want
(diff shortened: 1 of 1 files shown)


mperham:
Thank you for the quick fix!

PR #4550, opened 2020-05-03
Add tests for invalid concurrency and timeout setting

manojmj92:
This change adds (missing) test cases for `Sidekiq::CLI` when

1. An invalid (`<=0`) value for `concurrency` is supplied.
1. An invalid (`<=0`) value for `timeout` is supplied.


mperham:
Thank you!

PR #4555, opened 2020-05-06
Add Queue#delete_jobs_from

robinclart:
This makes it a bit easier to cleanup and repair a queue from a bad job. Which tends to be the most common case for using the `Sidekiq::Queue#each` method in my experience. This is crucial when you're in a hurry and said bad job is hammering a datastore to the point where it's impacting your whole application.

Before this addition, you had to do the following:

```ruby
queue.each { |job| job.delete if job.klass == "BadWorker" }
```

While definitely not bad, this is a bit of a contortion. My goal for this new API was to be shorter to type than the original, have no block or conditional and be easier to remember. Hence the following:

```ruby
queue.delete_jobs_from "BadWorker"
```

This also avoids having to know about the `Sidekiq::Job` API.

---

@mperham I notice too late that you prefer having an issue first before opening a PR. But given I already made the commit, I thought I would open this as a PR immediately anyway and we can discuss the legitimacy of this new feature here. Feel free to close if not in line with your design goals.

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -261,6 +261,12 @@ module Sidekiq
       end
     end
 
+    def delete_jobs_from(klass)
+      each do |job|
+        job.delete if job.klass == klass
+      end
+    end
+
     ##
     # Find the job with the given JID within this queue.
     #
--- a/test/test_api.rb
+++ b/test/test_api.rb
@@ -234,6 +234,17 @@ describe 'API' do
       assert_equal 0, q.size
     end
 
+    it 'can delete jobs by klass' do
+      q = Sidekiq::Queue.new
+      Time.stub(:now, Time.new(2012, 12, 26)) do
+        10.times { ApiWorker.perform_async(1, 'mike') }
+        assert_equal 10, q.size
+
+        q.delete_jobs_from('ApiWorker')
+        assert_equal 0, q.size
+      end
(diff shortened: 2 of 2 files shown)


mperham:
This is part of Sidekiq Pro so I can’t accept it into OSS. https://github.com/mperham/sidekiq/wiki/Pro-API

robinclart:
Didn't notice that. I'll let my team know (We're on Enterprise version so I guess it's there as well). And I guess they didn't notice as well 😂

MXfive:
@mperham I wonder what you think about restructuring the docs to have all the API across all versions on the same page, with a clear separation that the API below a certain point is in the pro version only.

I was the one dealing with a production issue yesterday and needed to kill all jobs of a certain class, I knew there was a way to do this with the API and the OSS API was the page Google obviously took me to first.  As I was under pressure from the production incident, it didn't occur to me to check the Pro API page also as I had already found a solution to the very immediate and time sensitive issue we were experiencing.

mperham:
I would like a public RDoc site but it's been so low priority that I
haven't put any time into it.

robinclart:
How about adding the methods and let them raise NotImplementedError in the OSS version and overwrite them in Pro and Ent versions such that there's no need for a different RDoc?

mperham:
Sure, send me a PR?

PR #4559, opened 2020-05-07
Avoid pushing TIMEOUT information to the end of the array of queues

manojmj92:
## Background

To fetch job data contained in the queues, Sidekiq uses the `redis` command `brpop`, which blocks the connection when there are no elements to pop from the given list of queues.

However, we do not want to always block the connection (Sidekiq will just work on forever in that case), so we also supply a `TIMEOUT`  (of 2 seconds) to prevent this.

## Change

Right now, `TIMEOUT` is also pushed to the end of array once, if the queues are strictly ordered.
In case of queues that are not strictly ordered, `TIMEOUT` data is being pushed to the end of the array every time a call to the `queues_cmd` method is made. 

While inserts to the end of the array are `O(1)`, we could avoid pushing the timeout information to this array completely, and instead just pass the timeout information as an argument to the redis client 
directly. This appears to be the recommended way as per the [docs](https://github.com/redis/redis-rb/blob/master/lib/redis.rb#L1220).

PS: It also appears that passing `timeout` information as the last element in the array *could* be deprecated by Redis client sometime in the future. [Source](https://github.com/redis/redis-rb/blob/f597f21a6b954b685cf939febbc638f6c803e3a7/lib/redis/distributed.rb#L409)

### DIFF
--- a/lib/sidekiq/fetch.rb
+++ b/lib/sidekiq/fetch.rb
@@ -27,14 +27,11 @@ module Sidekiq
     def initialize(options)
       @strictly_ordered_queues = !!options[:strict]
       @queues = options[:queues].map { |q| "queue:#{q}" }
-      if @strictly_ordered_queues
-        @queues.uniq!
-        @queues << TIMEOUT
-      end
+      @queues.uniq! if @strictly_ordered_queues
     end
 
     def retrieve_work
-      work = Sidekiq.redis { |conn| conn.brpop(*queues_cmd) }
+      work = Sidekiq.redis { |conn| conn.brpop(*queues_cmd, timeout: TIMEOUT) }
       UnitOfWork.new(*work) if work
     end
 
@@ -47,9 +44,7 @@ module Sidekiq
       if @strictly_ordered_queues
         @queues
       else
-        queues = @queues.shuffle!.uniq
-        queues << TIMEOUT
-        queues
+        @queues.shuffle!.uniq
       end
     end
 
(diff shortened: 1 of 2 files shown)


mperham:
This conflicts with Pro’s super_fetch so I can’t take it.

mperham:
I might be wrong. Looks like this just refactors BasicFetch which should be independent of SuperFetch.

lib/sidekiq/fetch.rb
```diff
@@ -27,14 +27,11 @@ def requeue
...
       @queues = options[:queues].map { |q| "queue:#{q}" }
-      if @strictly_ordered_queues
-        @queues.uniq!
-        @queues << TIMEOUT
-      end
+      @queues.uniq! if @strictly_ordered_queues
     end
 
     def retrieve_work
-      work = Sidekiq.redis { |conn| conn.brpop(*queues_cmd) }
```
mperham, line 37: This does have the effect of allocating a hash.  Is it better to use a possibly deprecated API and save a hash allocation or not?

manojmj92: Thanks, Mike,

I ran some benchmarks using `benchmark-ips` and the new method call is around 3x slower than the current method, so, purely if the decision is purely performance-based - probably not a good idea to merge.

But if the deprecation happens at some point, we'd be forced to use this anyway. Given the fact that the deprecation does not have a fixed date (it has been that way for at least a couple of years now), I do not have a strong opinion, to be honest 🙂

PR #4561, opened 2020-05-10
Define Reloader class before callback that uses it

eugeneius:
If the application has already been loaded before this file is required, the `after_initialize` callback runs immediately, but the Reloader class isn't defined yet.

In our case, this happens because we're preloading our application code before invoking Sidekiq Enterprise's sidekiqswarm command, to maximise the copy-on-write memory savings.

Also remove a comment about `cache_classes` that has been irrelevant since https://github.com/mperham/sidekiq/pull/3235.


mperham:
Rookie mistake: always define a class before referencing it, even lexically.

PR #4562, opened 2020-05-13
Allow newer rack version due to directory traversal security bug

ur5us:
[[CVE-2020-8161] Directory traversal in Rack::Directory](https://groups.google.com/d/msg/ruby-security-ann/T4ZIsfRf2eA/h32aSPjcBAAJ)


ur5us:
@mperham The motivation to loosen the version requirement and thus allow a slightly newer version of rack gem to be used is so that Sidekiq v5 can be used on Ubuntu 14.04 LTS which only ships Redis 2.8.

arothian:
Is there an ETA for a release of the 5.x line with this change?

mperham:
No but you can run github master in your Gemfile to pick up the fix.

mperham:
...and by master I mean the 5.x branch.

PR #4569, opened 2020-05-19
Elaborating a bit on reload/restart service

thomasdarde:
With this comment I think I would have avoid some struggle configuring sidekiq 6


mperham:
I think I would rather kill any usage of reload. It's a terrible idea in hindsight because reload == quiet but that's not obvious from the word.

PR #4572, opened 2020-05-20
update json

marekls:


mperham:
Sidekiq 3 has not been supported for years. You’ll need to upgrade.

PR #4575, opened 2020-05-22
Optimize stringify item before pushing it

dsantosmerino:
We were debugging something when found this piece of code in the `Sidekiq::Worker` class:
```ruby
# stringify
item.keys.each do |key|
  item[key.to_s] = item.delete(key)
end
```

Since the oldest supported version now is Ruby v2.5 and this version added the `Hash#transform_keys` method, we considered changing it to use the standard method, easier to read. Also, it looks like it brings some performance improvements, from 0 to 20% depending on the execution.

Here the benchmark code:
```ruby
# frozen_string_literal: true

require "benchmark/ips"

def stringify(item)
  item.keys.each do |key|
    item[key.to_s] = item.delete(key)
  end
end

def stringify_optimized(item)
  item.transform_keys(&:to_s)
end

item = {:queue=>"low", "class"=>"RandomWorker", "args"=>["foo", "bar"], "at"=>1590166226.951523}

10.times do |i|
  puts("Report ##{i}")
  Benchmark.ips do |x|
    x.report("stringify") do
      stringify(item)
    end
    x.report("stringify_optimized") do
      stringify_optimized(item)
    end
    x.compare!
  end
end
```

And one of the reports:
```txt
Report #9
           stringify    75.241k i/100ms
 stringify_optimized    74.384k i/100ms
           stringify    869.794k (±15.0%) i/s -      4.213M in   5.034072s
 stringify_optimized      1.056M (± 4.5%) i/s -      5.281M in   5.011056s
Comparison:
 stringify_optimized:  1056329.3 i/s
           stringify:   869793.5 i/s - 1.21x  slower
```

### DIFF
--- a/lib/sidekiq/redis_connection.rb
+++ b/lib/sidekiq/redis_connection.rb
@@ -8,16 +8,14 @@ module Sidekiq
   class RedisConnection
     class << self
       def create(options = {})
-        options.keys.each do |key|
-          options[key.to_sym] = options.delete(key)
+        symbolized_options = options.transform_keys(&:to_sym)
+
+        if !symbolized_options[:url] && (u = determine_redis_provider)
+          symbolized_options[:url] = u
         end
 
-        if !options[:url] && (u = determine_redis_provider)
-          options[:url] = u
-        end
-
-        size = if options[:size]
-          options[:size]
+        size = if symbolized_options[:size]
+          symbolized_options[:size]
         elsif Sidekiq.server?
           # Give ourselves plenty of connections.  pool is lazy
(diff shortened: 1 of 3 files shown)


mperham:
I believe there are several copies of this in the codebase. Can you find any more instances?

dsantosmerino:
> I believe there are several copies of this in the codebase. Can you find any more instances?

Yes, I found two more, one in the other way, symbolizing. Already changed both, let me know if you anything else to consider.

mperham:
Well done, that's all I could find too.

PR #4576, opened 2020-05-22
Fix SimpleCov

santib:
SimpleCov 0.18+ is not yet supported by CodeClimate
https://docs.codeclimate.com/docs/configuring-test-coverage#supported-languages-and-formats


mperham:
When you lock a specific version, please add a comment to the Gemfile explaining why.

mperham:
Just realized I can edit the PR, so I took care of it. Thanks for the fix!

PR #4578, opened 2020-05-26
Validate that job arguments being passed to `push_bulk` is an Array of Arrays

manojmj92:
## Background

Sidekiq allows to bulk push jobs into Redis via the `push_bulk` methods. One of the requirements for using this feature is that the job arguments being passed to this method call should be an "Array of Arrays".

## Problem

For some reason, the validation being done to make sure that the argument is an Array of Arrays was only checking if the first element of the array was an array. 

It is currently not checking if the `2nd` to `nth` element are arrays too.

## Change

This change adds:
- Validation that all elements of the array are of `Array` datatype.
- Tests to make sure that the new validation works as expected.

### DIFF
--- a/lib/sidekiq/client.rb
+++ b/lib/sidekiq/client.rb
@@ -90,16 +90,16 @@ module Sidekiq
     # Returns an array of the of pushed jobs' jids.  The number of jobs pushed can be less
     # than the number given if the middleware stopped processing for one or more jobs.
     def push_bulk(items)
-      arg = items["args"].first
-      return [] unless arg # no jobs to push
-      raise ArgumentError, "Bulk arguments must be an Array of Arrays: [[1], [2]]" unless arg.is_a?(Array)
+      args = items["args"]
+      raise ArgumentError, "Bulk arguments must be an Array of Arrays: [[1], [2]]" unless (args.is_a?(Array) && args.all?(Array))
+      return [] if args.empty? # no jobs to push
 
       at = items.delete("at")
       raise ArgumentError, "Job 'at' must be a Numeric or an Array of Numeric timestamps" if at && (Array(at).empty? || !Array(at).all?(Numeric))
 
       normed = normalize_item(items)
-      payloads = items["args"].map.with_index { |args, index|
-        copy = normed.merge("args" => args, "jid" => SecureRandom.hex(12), "enqueued_at" => Time.now.to_f)
+      payloads = args.map.with_index { |job_args, index|
+        copy = normed.merge("args" => job_args, "jid" => SecureRandom.hex(12), "enqueued_at" => Time.now.to_f)
         copy["at"] = (at.is_a?(Array) ? at[index] : at) if at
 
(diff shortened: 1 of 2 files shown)


lib/sidekiq/client.rb
```diff
@@ -90,15 +90,15 @@ def push(item)
     # Returns an array of the of pushed jobs' jids.  The number of jobs pushed can be less
     # than the number given if the middleware stopped processing for one or more jobs.
     def push_bulk(items)
-      arg = items["args"].first
-      return [] unless arg # no jobs to push
-      raise ArgumentError, "Bulk arguments must be an Array of Arrays: [[1], [2]]" unless arg.is_a?(Array)
+      arg = items["args"]
+      return [] unless arg.present? # no jobs to push
```
mperham, line 94: Nope, this is a Rails method.

lib/sidekiq/client.rb
```diff
@@ -90,15 +90,15 @@ def push(item)
     # Returns an array of the of pushed jobs' jids.  The number of jobs pushed can be less
     # than the number given if the middleware stopped processing for one or more jobs.
     def push_bulk(items)
-      arg = items["args"].first
-      return [] unless arg # no jobs to push
-      raise ArgumentError, "Bulk arguments must be an Array of Arrays: [[1], [2]]" unless arg.is_a?(Array)
+      arg = items["args"]
```
mperham, line 93: `args`, not `arg`.

lib/sidekiq/client.rb
```diff
@@ -90,16 +90,16 @@ def push(item)
     # Returns an array of the of pushed jobs' jids.  The number of jobs pushed can be less
     # than the number given if the middleware stopped processing for one or more jobs.
     def push_bulk(items)
-      arg = items["args"].first
-      return [] unless arg # no jobs to push
-      raise ArgumentError, "Bulk arguments must be an Array of Arrays: [[1], [2]]" unless arg.is_a?(Array)
+      args = items["args"]
+      raise ArgumentError, "Bulk arguments must be an Array of Arrays: [[1], [2]]" unless (args.is_a?(Array) && args.all?(Array))
+      return [] if args.empty? # no jobs to push
```
manojmj92, line 95: adding `.empty?` because `return [] unless args` continues execution even if the array is empty.
And this is now added after checking that arg is indeed an Array, to make sure that checks like `nil.empty?` never happen.

lib/sidekiq/client.rb
manojmj92: Corrected 🙂

lib/sidekiq/client.rb
manojmj92: Apologies, you caught me before I could force push after immediately realizing my mistake 😄

I had tested in my local before pushing, and this did not error out, since `activesupport` was available.

I have now changed this to `.empty?` 🙂

mperham:
Thank you!

PR #4581, opened 2020-05-27
Use error handler(s) from CLI launch failure

beatak:
**Problem**

Sometimes, the sidekiq workers of our app fail to launch by the Redis connectivity issue, and if that happens, it spits multiple lines of stack trace, regardless of configured `Sidekiq.error_handlers`. This patch will use the error handlers, if it's configured.

### DIFF
--- a/bin/sidekiq
+++ b/bin/sidekiq
@@ -31,7 +31,12 @@ begin
   cli.run
 rescue => e
   raise e if $DEBUG
-  STDERR.puts e.message
-  STDERR.puts e.backtrace.join("\n")
+  if Sidekiq.error_handlers.length === 0
+    STDERR.puts e.message
+    STDERR.puts e.backtrace.join("\n")
+  else
+    cli.handle_exception e
+  end
+
   exit 1
 end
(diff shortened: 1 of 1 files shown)


mperham:
I think CLI launch failures should always output to STDERR. Exception handlers are for job errors and other things that can be recovered from once Sidekiq has booted up successfully. Confirm/deny?

mperham:
Oh, I see. Once exception_handler.rb is required, we have something which will print out to the Terminal anyways...

bin/sidekiq
```diff
@@ -31,7 +31,12 @@ begin
   cli.run
 rescue => e
   raise e if $DEBUG
-  STDERR.puts e.message
-  STDERR.puts e.backtrace.join("\n")
+  if Sidekiq.error_handlers.length === 0
```
mperham, line 34: Why ===? I think you are doing too much JavaScript? 😂

beatak: ah sorry this is my bad. I'll fix.

beatak:
I wanted to have a consistent way to collect errors and I thought `Sidekiq.error_handlers` is a good place. In my case, I'm sending the errors to Datadog via `Sidekiq.error_handlers`.

mperham:
Thank you!

PR #4588, opened 2020-06-03
Vendor rack-protection/authenticity_token middleware

### LINKED ISSUE #4587: Remove dependency on `rack-protection`?
Can we remove the dependency on `rack-protection`? afaik we only use the CSRF protection. Can we vendor that?

seuros:
Closes #4587

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -4,7 +4,6 @@ PATH
     sidekiq (6.0.7)
       connection_pool (>= 2.2.2)
       rack (~> 2.0)
-      rack-protection (>= 2.0.0)
       redis (>= 4.1.0)
 
 GEM
@@ -117,8 +116,6 @@ GEM
       byebug (~> 11.0)
       pry (~> 0.10)
     rack (2.2.2)
-    rack-protection (2.0.8.1)
-      rack
     rack-test (1.1.0)
       rack (>= 1.0, < 3)
     rails (6.0.3.1)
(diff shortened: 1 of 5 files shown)


Gemfile.lock
```diff
@@ -211,3 +208,6 @@ DEPENDENCIES
   sqlite3
   standard
   toxiproxy
+
```
mperham, line 211: Remove this and Bundler version doesn’t matter

seuros: Done.

lib/sidekiq/web/authenticity_token.rb
```diff
@@ -0,0 +1,231 @@
+require 'securerandom'
+require 'base64'
+require 'rack/request'
+
+module Sidekiq
+  class Web
+    ## Shamelessly copied from Rack-protection gem
```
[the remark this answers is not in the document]
mperham: https://github.com/sinatra/sinatra/blob/master/rack-protection/License

mperham:
Well done! I wonder if we could remove Rack and just have Sidekiq::Web quack like a Rack app... 🤔

natematykiewicz:
> Well done! I wonder if we could remove Rack and just have Sidekiq::Web quack like a Rack app... 🤔

A safer option is probably to add a soft dependency on Rack. Like how Rails does with various database drivers.

https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L3-L4

Basically, you still require a specific Rack version _**if**_ they attempt to use mount the Web UI.

PR #4599, opened 2020-06-16
Upgrade Redis to 4.2.1.

tgxworld:
Pulling in fix for `Redis#exists?` https://github.com/redis/redis-rb/commit/525a93edbb2e65ea94634f30153d38cf85e3fe7f


mperham:
There’s no need for this. The gemspec is >= so you can upgrade if you need that fix.

PR #4602, opened 2020-06-17
Fetch API refactor, WIP

mperham:
The existing fetch API grew organically to its current state but is less than optimal: you pass a class reference via the global options, making it hard to configure a fetcher instance and requiring some test gymnastics to set up that global config.

I'd like to change the fetch API to be instance-based. `bulk_requeue` will become an instance method. Feedback welcome.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -12,6 +12,7 @@ HEAD
 - Remove rack-protection, reimplement CSRF protection [#4588]
 - Require redis-rb 4.2 [#4591]
 - Update to jquery 1.12.4 [#4593]
+- Refactor internal fetch logic and API [#4602]
 
 6.0.7
 ---------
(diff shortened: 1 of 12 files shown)


jhoffner:
It would be ideal if either a class or instance could be passed, for backwards compatibility with 3rd party gems.

vinc:
Shouldn't this have been a major version change as it's not backwards compatible? I didn't expect a version change from 6.1.0 to 6.1.1 to break anything. A fix as suggested by @jhoffner to be released in version 6.1.2 would be great!

mperham:
The Fetch API has never been documented, it's not intended as a public, "user-facing" API. Internals can change.

Furthermore I want to point out that Sidekiq is not semantic versioned. I do my best to limit any possible breaking change to minor or major releases based on the expected impact, patch releases try not to break anything.

vinc:
Thanks for the explanation, got it 👍

cubbiu:
guys, maybe you can create 3.4.1 tag, so we can just update the gem's version in gemfile?

PR #4617, opened 2020-07-01
Remove wrong comment on instance method

mobilutz:
Within #4602 this method was changed from a class- to instance-method. The comment should be removed as the method is not a class-method anymore.


mperham:
thank you!

PR #4656, opened 2020-07-28
JSON Log Formatter following the recommend patterns of ECS

pedrobachiega:
https://www.elastic.co/guide/en/ecs/current/ecs-base.html

Using keys `@timestamp`, `severity` and `message` that are being broadly used at internet.

### DIFF
--- a/lib/sidekiq/logger.rb
+++ b/lib/sidekiq/logger.rb
@@ -148,11 +148,11 @@ module Sidekiq
       class JSON < Base
         def call(severity, time, program_name, message)
           hash = {
-            ts: time.utc.iso8601(3),
+            "@timestamp" => time.utc.iso8601(3),
             pid: ::Process.pid,
             tid: tid,
-            lvl: severity,
-            msg: message
+            severity: severity,
+            message: message
           }
           c = ctx
           hash["ctx"] = c unless c.empty?


mperham:
Why should we treat ECS as canonical? Are there other de facto standards or conventions we can adopt instead? E.g. does systemd's journald have an opinion?

pedrobachiega:
> Why should we treat ECS as canonical? Are there other de facto standards or conventions we can adopt instead? E.g. does systemd's journald have an opinion?

Hey Mike!
I and my teams are seeing the raise of ECS around the community, basically because it is an open source specification.
Some references:
 - It's a suggestion at kubernetes documentation: https://kubernetes.io/docs/tasks/debug-application-cluster/logging-elasticsearch-kibana/
 - Code42 already published that they are using it too: https://redblue42.code42.com/implementation-considerations-for-the-elastic-common-schema/

But you are right about questioning others opinions. Unfortunately, there isn't a de facto standard yet.
What is a better approach here at the repository, maybe create an issue to discuss?

mperham:
I think the right approach here is to create an ECS-specific logging plugin ([per the 6.0 release notes](https://github.com/mperham/sidekiq/blob/master/6.0-Upgrade.md#whats-new)). If there is good demand, we can pull it into core at some point. If you need changes in core to make it possible, open a specific issue or send us a PR.

Issue #4660, opened 2020-08-04
Push jobs after DB transactions committed


donaldong:
### Motivation
In an ActiveRecord DB transaction, people might want to push Sidekiq jobs to Redis. If the transaction failed, all the database side operations will rollback; but we cannot rollback the API calls to Redis.

It would be helpful if Sidekiq provides an option to defer job pushing until DB transactions have committed successfully -- if the transaction failed, the jobs would be canceled.

I'd be happy to submit a PR, but I noticed the `raw_push` method could be overwritten in Sidekiq-Pro for the Pro features. I can still check if the method is overwritten in the opensource version, tho checking for a Pro overwrite in the opensource version seems a bit odd.

Thoughts?

mperham:
I'd be curious to see a PR if you want to submit something. If possible it should work with Rails 5+. don't worry about Sidekiq Pro, that would be my work if we decide to proceed.

axos88:
@donaldong Take a look at the `after_commit` gem

PR #4661, opened 2020-08-05
Push jobs after DB transactions committed

donaldong:
### Motivation
This addresses #4660 

Delayed jobs can be pushed in transactions safely since they often live in the same database. When using sidekiq, we still want to push jobs in a transaction. Ideally, if the DB transaction fails, the jobs need to be canceled somehow to preserve the original behavior.

This defers the job pushing until a transaction successfully completes. If the transaction failed, the jobs won't get pushed (similar to rolling back a delayed job).

### Approach
When jobs are pushed in a DB transaction, instead of pushing the jobs immediately, we register a callback to push to the jobs. The callback will be fired only if the transaction has committed successfully.

To enable this feature, in the sidekiq initializer:
```ruby
  # If you're using transactional DB cleaner in the test environment:
  # disable this feature or the callbacks won't get fired until the end of the test cases
  config.push_after_commit! unless Rails.environment.test?
  # ...
end
```

**Note**: I added a method `raw_push_immediately` which can be integrated into Sidekiq Pro's [client reliability](https://github.com/mperham/sidekiq/wiki/Pro-Reliability-Client). If the pushing processes failed, the pro client could still automatically retry in the background by extending this method.

### Test Plan
- Added new test cases to test the new behavior.
- No test failures locally.
- The patch has also been tested on rails 5.2.

### DIFF
--- a/lib/sidekiq/client.rb
+++ b/lib/sidekiq/client.rb
@@ -67,12 +67,14 @@ module Sidekiq
     #   push('queue' => 'my_queue', 'class' => MyWorker, 'args' => ['foo', 1, :bat => 'bar'])
     #
     def push(item)
-      normed = normalize_item(item)
-      payload = process_single(item["class"], normed)
+      after_commit do
+        normed = normalize_item(item)
+        payload = process_single(item["class"], normed)
 
-      if payload
-        raw_push([payload])
-        payload["jid"]
+        if payload
+          raw_push([payload])
+          payload["jid"]
+        end
       end
     end
 
@@ -98,17 +100,19 @@ module Sidekiq
       raise ArgumentError, "Job 'at' must be a Numeric or an Array of Numeric timestamps" if at && (Array(at).empty? || !Array(at).all?(Numeric))
       raise ArgumentError, "Job 'at' Array must have same size as 'args' Array" if at.is_a?(Array) && at.size != args.size
 
-      normed = normalize_item(items)
-      payloads = args.map.with_index { |job_args, index|
-        copy = normed.merge("args" => job_args, "jid" => SecureRandom.hex(12), "enqueued_at" => Time.now.to_f)
-        copy["at"] = (at.is_a?(Array) ? at[index] : at) if at
+      after_commit do
+        normed = normalize_item(items)
+        payloads = args.map.with_index { |job_args, index|
+          copy = normed.merge("args" => job_args, "jid" => SecureRandom.hex(12), "enqueued_at" => Time.now.to_f)
+          copy["at"] = (at.is_a?(Array) ? at[index] : at) if at
 
-        result = process_single(items["class"], copy)
-        result || nil
-      }.compact
+          result = process_single(items["class"], copy)
+          result || nil
+        }.compact
 
-      raw_push(payloads) unless payloads.empty?
-      payloads.collect { |payload| payload["jid"] }
+        raw_push(payloads) unless payloads.empty?
+        payloads.collect { |payload| payload["jid"] }
+      end
     end
 
     # Allows sharding of jobs across any number of Redis instances.  All jobs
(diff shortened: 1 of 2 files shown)


mperham:
I think you need to integrate this in as a Railtie, ie in sidekiq/rails.rb.

mperham:
But a very promising start! Is there any reason why someone wouldn't want this behavior and might want to disable it?

donaldong:
Thanks :)

> I think you need to integrate this in as a Railtie, ie in sidekiq/rails.rb.

Okay! I will separate the callback class.

> Is there any reason why someone wouldn't want this behavior and might want to disable it?

One typical use case is to disable it in the test environment. Some projects use transactional database cleaner -- all the test cases are wrapped in transactions. In this scenario, we'd want to disable this behavior to push jobs in test cases.

mperham:
I'm not a fan of config switches for features that the user has to enable. Convention over configuration. That said, this has the potential to  change code behavior or break tests. Is it a small change that has big effects?

1. Will this change break tests or test suites?
1. Should it default to "off" in the test environment and "on" in other environments?
1. To that end, are there aspects or APIs that should go in Sidekiq::Testing?
1. Any issues or concerns you can think of with non-Rails apps, e.g. someone running Hanami + Sequel?
1. Should it be part of Sidekiq 7.0 as a major, possibly breaking, change?
1. If the user has to fix their test suite due to this change, is this change worth the pain to the user?

I just realized this change will break client middleware:

1. User calls `push`.
1. Client middleware runs, puts associated data (e.g. a unique lock) in Redis.
1. DB transaction rolls back so job is not actually pushed
1. Now there is a unique lock without an associated job

As you can see, there's lots of semantic land mines here. We need to tread very carefully.

donaldong:
> 1. Will this change break tests or test suites?
> 2. Should it default to "off" in the test environment and "on" in other environments?
> 3. To that end, are there aspects or APIs that should go in Sidekiq::Testing?

Seems when using active record, we could pretty reliable workaround testing: [after_commit_everywhere](https://github.com/Envek/after_commit_everywhere/blob/be8602f9fbb8e40b0fc8a04a47e4c2bc6b560ad5/lib/after_commit_everywhere.rb#L93). I updated the implementation and removed the option of configuring this feature.

> 6. If the user has to fix their test suite due to this change, is this change worth the pain to the user?

😅  it will break the tests if the test cases are checking if the jobs are pushed within a transaction (or something similar). I think it'd make sense for people to rewrite the test cases in favor of the new behavior -- should we offer some options to adopt this new behavior in the test environment incrementally?

> I just realized this change will break client middleware:

Oh yeah very true! I updated the implementation to patch from the top (push, push_bulk).

> 4. Any issues or concerns you can think of with non-Rails apps, e.g. someone running Hanami + Sequel?

I have not used Sequel. In my implementation, I assumed the connection is an active record connection. Perhaps we can offer an option to let people specify the SQL ORM lib they're using? (we now support ActiveRecord, later we can add support to other libs) It might get more interesting when people have multiple ORM libs / multiple DB connections.

> 5. Should it be part of Sidekiq 7.0 as a major, possibly breaking, change?

Perhaps? that'd make sense to me. 😃

mperham:
This code seems really self-contained but has big semantic effects. I wouldn't envision pulling this into Sidekiq before 7.0 and that might be another 6-12 months away. If you put out a gem and get bug reports from users for a few months, people can start using it today and I think you'd have a very strong case to pull it into core Sidekiq. What do you think about making it a 3rd party gem that people can opt into?

dzunk:
I don't think this will work in a multi-DB Rails application, as not all models use the connection pool from `ActiveRecord::Base`.

janko:
I believe it would make the most sense to pull this into a gem, e.g. `sidekiq-activerecord`. That way you can also use dependencies such as after_commit_everywhere, which people would pull only if using Active Record.

Also, in these days we shouldn't assume Active Record / Rails anymore, as this supports a monoculture that Active Record is *the* ORM, when in fact it's just *a* ORM – people are using Sequel and ROM more and more (and there is also Mongoid etc). And people are using Sidekiq in other web frameworks.

donaldong:
Thanks, everyone! Yea, let's make this a separate gem. I will link it here once it's ready.

christos:
I always push jobs to Sidekiq in `after_commit` (instead of `after_save`) hooks because often enough Sidekiq is faster processing a job than the DB is committing the transaction. 

When that happens, the job might fail to see the newly created (or updated) record in the DB the first time it runs.

I actually have a module that I include in the models I want to ensure are committed before pushing a job to Sidekiq, that wraps the `perform_async` and delays pushing it, until after commit. Just like this PR

Having said all that, I think some explicit mention in the docs about the fact that a job can be pushed and performed before the record that triggered it is even committed to the DB would be useful.

As would a gem that enables this behaviour on a per model basis (say by doing an `include Sidekiq::Transactional` or something like that.

PR #4673, opened 2020-08-23
Allow middleware to yield arguments

eugeneius:
I noticed that our application has several middleware that wrap the job in a method that yields:

```ruby
class TaggedLoggingMiddleware
  def call(*)
    Rails.logger.tagged("my cool job") { yield }
  end
end
```

To save a stack frame, I rewrote them to pass the given block directly to the method:

```ruby
class TaggedLoggingMiddleware
  def call(*, &block)
    Rails.logger.tagged("my cool job", &block)
  end
end
```

I even made a change to RuboCop (https://github.com/rubocop-hq/rubocop/pull/8455) so that they could be rewritten automatically.

When I deployed the change, every job started to fail with:

> ArgumentError: wrong number of arguments (given 1, expected 0)

😩 

This happened because [lambdas have strict argument checking](https://ruby-doc.org/core-2.6.6/Proc.html#class-Proc-label-Lambda+and+non-lambda+semantics), and `Rails.logger.tagged` [yields the logger](https://github.com/rails/rails/blob/v6.0.3.2/activesupport/lib/active_support/tagged_logging.rb#L80). We had unit tests for the middleware, but they called it with a proc, so they didn't catch the issue.

I don't think strict argument checking makes sense here; any yielded arguments should be accepted and ignored.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -6,6 +6,7 @@ HEAD
 ---------
 
 - Fix Web UI crash with corrupt session [#4672]
+- Allow middleware to yield arguments [#4673, @eugeneius]
 
 6.1.1
 ---------
--- a/lib/sidekiq/middleware/chain.rb
+++ b/lib/sidekiq/middleware/chain.rb
@@ -133,7 +133,7 @@ module Sidekiq
         return yield if empty?
 
         chain = retrieve.dup
-        traverse_chain = lambda do
+        traverse_chain = proc do
           if chain.empty?
             yield
           else
--- a/test/test_middleware.rb
+++ b/test/test_middleware.rb
@@ -41,6 +41,12 @@ describe Sidekiq::Middleware do
     end
   end
 
+  class ArgumentYieldingMiddleware
+    def call(*args)
+      yield 1
+    end
+  end
+
   class AnotherCustomMiddleware
     def initialize(name, recorder)
       @name = name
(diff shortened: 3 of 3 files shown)


mperham:
I have a lot of respect for people trying to remove stack frames. 😎 this isn't something I'd thought of before, it's quite subtle but nice to see.￼

mperham:
Diving deeper, I'm not sure I like this change.  What does it mean to yield "1" or `logger` to a middleware API that is designed to have `worker, job, queue` arguments?

eugeneius:
The yielded values are always ignored, which isn't very useful if the middleware is yielding them directly; it's only helpful when the author of the middleware doesn't control the behaviour, as in the case I described above.

I have no way to tell the `tagged` method _not_ to yield the logger, so I can't pass the block without wrapping it. In general, it's not easy to understand whether the block needs to be wrapped or not; `Rails.logger.tagged` doesn't even [document](https://api.rubyonrails.org/v6.0.3.2/classes/ActiveSupport/TaggedLogging.html) what it yields! When combined with the fact that this problem is very difficult to catch with a test unless you're aware that Sidekiq passes a lambda, I thought it was worth allowing the block to be passed to any method that yields. There's my pitch 😄

mperham:
Ok, diving even deeper, I see the middleware chain controls the args passed to each middleware so this does add a bit of nice flexibility without risk.

PR #4712, opened 2020-10-07
Add timeout value to process info

igas:
I'm trying to introspect `Sidekiq::ProcessSet.new`, it already has `concurrency` value, but I also need `timeout`.

### DIFF
--- a/lib/sidekiq/launcher.rb
+++ b/lib/sidekiq/launcher.rb
@@ -188,6 +188,7 @@ module Sidekiq
           "pid" => ::Process.pid,
           "tag" => @options[:tag] || "",
           "concurrency" => @options[:concurrency],
+          "timeout" => @options[:timeout],
           "queues" => @options[:queues].uniq,
           "labels" => @options[:labels],
           "identity" => identity


mperham:
Why? What’s the use case? I’m not against this, I just want to understand what you are trying to do.

igas:
Hi @mperham,

Thanks for a prompt response.

I have a couple of use-cases, currently, we use elastic beanstalk and it's a bit of magic involved. We got our timeout set, but it does not look like it's correctly set. For debug one of my colleagues used:
```ruby
class SidekiqOptions < BaseService
  def execute
    pp Sidekiq.options
  end
end
```
To get the options, if we had this we'd just use `Sidekiq::ProcessSet.new`.

Moving forward we'll need a more complicated "manager", where we'll have different queues with different timeouts. Having queue and timeout date we'll be able to decide on when to quite/kill process.

Hope it makes sense.

mperham:
You can turn on -v verbose. Sidekiq prints the options on boot in debug mode.

igas:
It would work I guess, but it requires at least 2 redeploys, while introspection is not. We can solve the first problem one way or another, so it's less of the problem.

How do you propose solving the following?

> Moving forward we'll need a more complicated "manager", where we'll have different queues with different timeouts. Having queue and timeout we'll be able to decide on when to quite/kill process.

mperham:
> It would work I guess, but it requires at least 2 redeploys, while introspection is not. We can solve the first problem one way or another, so it's less of the problem.

I thought this was a one-time bug with EBS configuration that you were trying to fix. If so, two deploys is not unreasonable.

> How do you propose solving the following?

Different Sidekiq processes. Each process can have different queues and shutdown timeouts.

igas:
I'm happy to close it for now and revise when I have an example.

Issue #4671, opened 2020-08-19
Sidekiq UI responds with internal server error when session cookie is invalid


basherru:
Ruby version: ruby 2.7.0p0
Sidekiq / Pro version(s): 6.1.1, 5.1.1
Rails version: 6.0.3.2

We are using Rails 6.0.3.2 with Sidekiq Pro. When the `rack.session` cookie is corrupted and the digest verification in [Rack::Session::Cookie](https://github.com/rack/rack/blob/2-2-stable/lib/rack/session/cookie.rb#L142) fails, it returns an empty hash as a session data. This leads to sidekiq web trying to get the `:csrf` key in the empty hash and getting `nil`, which is then passed to `Base64.strict_encode64` sending the `unpack1` message to it.

Steps to reproduce:

1. Visit Sidekiq web UI with default session store
2. Change like a bit in the `rack.session` cookie digest (after `--` delimiter)
3. Try to perform any action that require `POST` request with form-data

```
NoMethodError: undefined method `unpack1' for nil:NilClass
  from base64.rb:74:in `strict_decode64'
  from sidekiq/web/csrf_protection.rb:145:in `decode_token'
  from sidekiq/web/csrf_protection.rb:141:in `compare_with_real_token'
  from sidekiq/web/csrf_protection.rb:105:in `valid_token?'
  from sidekiq/web/csrf_protection.rb:76:in `accept?'
  from sidekiq/web/csrf_protection.rb:41:in `call'
  from rack/session/abstract/id.rb:266:in `context'
  from rack/session/abstract/id.rb:260:in `call'
  from rack/content_length.rb:17:in `call'
  from rack/urlmap.rb:74:in `block in call'
  from rack/urlmap.rb:58:in `each'
  from rack/urlmap.rb:58:in `call'
  from rack/builder.rb:244:in `call'
  from sidekiq/web.rb:105:in `call'
  from sidekiq/web.rb:110:in `call'
  from action_dispatch/routing/mapper.rb:19:in `block in <class:Constraints>'
  from action_dispatch/routing/mapper.rb:48:in `serve'
  from action_dispatch/journey/router.rb:49:in `block in serve'
  from action_dispatch/journey/router.rb:32:in `each'
  from action_dispatch/journey/router.rb:32:in `serve'
  from action_dispatch/routing/route_set.rb:834:in `call'
  from rack/etag.rb:27:in `call'
  from rack/conditional_get.rb:40:in `call'
[...]

PR #4731, opened 2020-11-04
Fix handling of no csrf session token

szechyjs:
The previous fix for #4671 added a guard for a completely empty session. However, it seems that even if you delete the `rack.session` cookie, the session will still contain `session_id`, thus its not actually `empty?`. This changes the guard to actually check for the csrf token.

### DIFF
--- a/lib/sidekiq/web/csrf_protection.rb
+++ b/lib/sidekiq/web/csrf_protection.rb
@@ -90,12 +90,10 @@ module Sidekiq
         end
 
         sess = session(env)
-
-        # Checks that Rack::Session::Cookie did not return empty session
-        # object in case the digest verification failed
-        return false if sess.empty?
-
         localtoken = sess[:csrf]
+
+        # Checks that Rack::Session::Cookie actualy contains the csrf toekn
+        return false if localtoken.nil?
 
         # Rotate the session token after every use
         sess[:csrf] = SecureRandom.base64(TOKEN_LENGTH)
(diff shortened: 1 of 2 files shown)


lib/sidekiq/web/csrf_protection.rb
```diff
@@ -90,13 +90,11 @@ def valid_token?(env, giventoken)
...
         sess = session(env)
-
-        # Checks that Rack::Session::Cookie did not return empty session
-        # object in case the digest verification failed
-        return false if sess.empty?
-
         localtoken = sess[:csrf]
 
+        # Checks that Rack::Session::Cookie actualy contains the csrf toekn
+        return false if localtoken.blank?
```
mperham, line 96: `blank?` is Rails-specific.

szechyjs: good catch, updated to plain ruby

lib/sidekiq/web/csrf_protection.rb
```diff
@@ -90,13 +90,11 @@ def valid_token?(env, giventoken)
...
         sess = session(env)
-
-        # Checks that Rack::Session::Cookie did not return empty session
-        # object in case the digest verification failed
-        return false if sess.empty?
-
         localtoken = sess[:csrf]
 
+        # Checks that Rack::Session::Cookie actualy contains the csrf toekn
+        return false if localtoken.to_s.strip.empty?
```
mperham, line 96: Why not just `return false if localtoken.nil?`

szechyjs: true, i guess protecting against an empty string isn't necessary here

mperham:
thank you!

PR #4751, opened 2020-11-28
ensure `wrapped` is of type string, representing the AJ class name

rcugut:
This addresses a change in Rails codebase related to the Sidekiq adapter: https://github.com/rails/rails/commit/eb4f36c13f63fcabcd211ca8c669420e46c19a34 as explained in [this commit comment](https://github.com/rails/rails/commit/eb4f36c13f63fcabcd211ca8c669420e46c19a34#commitcomment-44607834).

Sidekiq and related libs that are built around the ecosystem ([like scout_apm](https://github.com/scoutapp/scout_apm_ruby/issues/342)) expect the `wrapped` job argument to be consistent, of type string, as the `name` of the underlying ActiveJob class.

### DIFF
--- a/lib/sidekiq/client.rb
+++ b/lib/sidekiq/client.rb
@@ -243,6 +243,9 @@ module Sidekiq
 
       raise(ArgumentError, "Job must include a valid queue name") if item["queue"].nil? || item["queue"] == ""
 
+      # ensure the wrapped class, if exists, is serialized as its name
+      item["wrapped"] = item["wrapped"].name if item["wrapped"].respond_to?("name")
+
       item["class"] = item["class"].to_s
       item["queue"] = item["queue"].to_s
       item["jid"] ||= SecureRandom.hex(12)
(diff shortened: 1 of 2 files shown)


mperham:
Don’t monkey patch JSON if it’s not 100% compatible. You can still use the Yajl APIs directly. https://github.com/brianmario/yajl-ruby#json-gem-compatibility-api

rcugut:
hmm, I'm not trying to... I think? 🤔 

I intentionally enabled Yajl's JSON gem compatibility (globally per project), so that my entire project would use Yajl under the hood. And doing so overrides `JSON.generate`, which is used [by Sidekiq to serialize objects to send to redis](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq.rb#L184), and which breaks the serialization of that `wrapped` field.

I ended up removing the yajl-ruby gem...

mperham:
I've added an enty to the troublesome gems section of the Problems and Troubleshooting wiki page which warns about using Oj or Yajl-ruby.

PR #4752, opened 2020-12-02
Warn on boot if maxmemory policy is not 'noeviction'

odlp:
This PR proposes logging a warning when the CLI boots if the maxmemory policy is not 'noeviction'.

Inspired by [Nate Berkopec's tweet the other day](https://twitter.com/nateberkopec/status/1333863949238931464) (I've also had clients who had this misconfiguration):

> Running Sidekiq from a Redis datastore with a volatile eviction policy is probably the most common Sidekiq error I see in the wild w/clients. Switch to noeviction, you want Sidekiq to fail very loudly when you run out of memory, not silently drop jobs.

### DIFF
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -62,6 +62,11 @@ module Sidekiq
       ver = Sidekiq.redis_info["redis_version"]
       raise "You are connecting to Redis v#{ver}, Sidekiq requires Redis v4.0.0 or greater" if ver < "4"
 
+      maxmemory_policy = Sidekiq.redis_info["maxmemory_policy"]
+      if maxmemory_policy != "noeviction"
+        logger.warn "'noeviction' maxmemory policy is recommended (current policy: '#{maxmemory_policy}'). See: https://github.com/mperham/sidekiq/wiki/Using-Redis#memory"
+      end
+
       # Since the user can pass us a connection pool explicitly in the initializer, we
       # need to verify the size is large enough or else Sidekiq's performance is dramatically slowed.
       cursize = Sidekiq.redis_pool.size
--- a/test/test_cli.rb
+++ b/test/test_cli.rb
@@ -481,6 +481,32 @@ describe Sidekiq::CLI do
           assert_includes @logdev.string, "Booted Rails #{::Rails.version} application in production environment"
         end
       end
+
+      describe 'checking maxmemory policy' do
+        it 'warns if the policy is not noeviction' do
+          redis_info = { "maxmemory_policy" => "allkeys-lru", "redis_version" => "6" }
+
+          Sidekiq.stub(:redis_info, redis_info) do
+            subject.stub(:launch, nil) do
+              subject.run
+            end
+          end
+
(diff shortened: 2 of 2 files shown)


odlp:
Test failure on Ruby 2.5 appears to be unrelated. It's a temporal assertion which failed (CI running slow?):

```
Failure:
API::with an empty database#test_0024_can enumerate retries [/home/runner/work/sidekiq/sidekiq/test/test_api.rb:484]:
Expected |1606907616.4180453 - 1606907616.3701653| (0.047879934310913086) to be <= 0.02.

rails test /home/runner/work/sidekiq/sidekiq/test/test_api.rb:472
```

lib/sidekiq/cli.rb
```diff
@@ -62,6 +62,11 @@ def run(boot_app: true)
       ver = Sidekiq.redis_info["redis_version"]
       raise "You are connecting to Redis v#{ver}, Sidekiq requires Redis v4.0.0 or greater" if ver < "4"
 
+      maxmemory_policy = Sidekiq.redis_info["maxmemory_policy"]
+      if maxmemory_policy != "noeviction"
```
mperham, line 66: There will be a lot of these in development.  I'd suggest you only print this if env != "development".

odlp: Taking a look at `redis.conf` in various versions, I believe `noeviction` is the default:

- [4.0](https://github.com/redis/redis/blob/ff6db5f176caa426135bf9e603a1eb6f6dea4f83/redis.conf#L588-L590)
- [5.0](https://github.com/redis/redis/blob/a767d84a7273ea05bbfbb23f4eef63af7f4a613d/redis.conf#L595-L597)
- [6.0](https://github.com/redis/redis/blob/25214bd7dc2f4c995d76020e95180eb4e6d51672/redis.conf#L889-L891)

But appreciate that it's a somewhat meaningless check in development. I'll add the additional condition 👍

mperham: Good catch, if it’s the default in the brew formula then I’m ok with it as is.

odlp: It appears that the Homebrew formulae don't modify `maxmemory_policy`.

Here's `4.0`:
https://github.com/Homebrew/homebrew-core/blob/fdb83fcfb482e5ed1f1c3c442a85b99223fcabeb/Formula/redis%404.0.rb#L27-L31

Here's `6.0`:
https://github.com/Homebrew/homebrew-core/blob/fdb83fcfb482e5ed1f1c3c442a85b99223fcabeb/Formula/redis.rb#L30-L34

mperham:
Thank you, great idea!

Dantemss:
I have a question about why this warning is necessary when using volatile-* policies:
If the maxmemory policy is volatile-* then keys without a TTL will never be evicted...
So I imagine those policies would work fine unless Sidekiq was setting a TTL on important keys or if the client somehow managed to configure redis-rb to add a TTL to all keys.
I've seen that configuration option on the Rails cache side, but I don't think redis-rb itself supports it.
So what would be causing those keys to expire?

PR #4767, opened 2020-12-18
Spike next-gen sidekiq config and runner API

mperham:
There are several usecases to which Sidekiq is not well-suited today:

1. Kick off a Sidekiq process via Lambda, process jobs until queue is empty or 10 minutes have passed.
2. Embed a Sidekiq instance within a larger Ruby/Rails process, e.g. Discourse.

Refactor and prototype a new API for configuring and running Sidekiq purely from Ruby, rather than requiring a binstub, config YML, CLI args, etc.

This is still very hypothetical, nothing is sure or documented.


courtsimas:
Love the direction this is going!

PR #4770, opened 2020-12-23
Fix logger with from deleting existing context

mmrobins:
If using `with` to add logging context, any context that has the same
keys as existing context will disappear after the with block completes,
instead of reverting back to the original context.

This fixes that by restoring the original context directly instead of
just deleting the keys from the `with` context.

### DIFF
--- a/lib/sidekiq/logger.rb
+++ b/lib/sidekiq/logger.rb
@@ -6,10 +6,11 @@ require "time"
 module Sidekiq
   module Context
     def self.with(hash)
+      orig_context = current.dup
       current.merge!(hash)
       yield
     ensure
-      hash.each_key { |key| current.delete(key) }
+      Thread.current[:sidekiq_context] = orig_context
     end
 
     def self.current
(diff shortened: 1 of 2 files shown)


mperham:
Thanks, can you explain how you triggered this issue? Were you using this API directly or is there another issue to fix too?

mmrobins:
Yeah, I have a sidekiq job that I call with `perform_async`, and then it synchronously calls another sidekiq job that *sometimes* gets called async too.  I wanted the logs in the synchronously called job to have more appropriate context than the one from the parent job, so I was wrapping code in a `with` block, and was surprised that at the end the calling job didn't have any context left.

pseuocode

```
class ParentJob
  include Sidekiq::Worker

  def perform
    ChildJob.new.perform
    # now the class context is gone when the `done` log happens
  end
end

class ChildJob
  include Sidekiq::Worker

  def perform
    if Sidekiq.context[:class] != self.class
      Sidekiq::Context.with(calling_class: Sidekiq.context[:class], class: self.class) do
        logger.info "in the child class doing stuff synchronously"
      end
    end
  end
end
```

mperham:
Thanks for the context!

PR #4787, opened 2021-01-24
spell corrected in sidekiq.service

AsbahIshaq:
changed 'quiet' to 'quit'


mperham:
Quiet is the correct term here.

PR #4804, opened 2021-02-09
Improve Web UI session experience

mperham:
For years many people have struggled to configure a Rack session with the rewritten Web UI introduced in v4.2.0. "Forbidden" is a common complaint and there are many hacks around this area in many apps. This change is an attempt to clean up this problem and hopefully guide people back onto the happy path and away from those hacks.

The Web UI no longer exposes any session configuration. Anything like `Sidekiq::Web.sessions = false` and `Sidekiq::Web.session_secret = ...` is no longer available.

The Web UI only supports two modes of operation now:

1. Running within a Rails app and reusing the existing session.
2. Running as a bare Rack app, user must explicitly configure a session.

## Rails

For Rails, as long as you mount the Web UI within the routes, it should provide a session for anything within the routes.

```ruby
# config/routes.rb
require 'sidekiq/web'

Rails.application.routes.draw do
  mount Sidekiq::Web => "/sidekiq"
  ....
end
```

## Rack

For Rack, a basic cookie session is straightforward to enable:

1. First create a shared secret for all Sidekiq processes to use and commit it to git:
```ruby
require 'securerandom'; File.open(".session.key", "w") {|f| f.write(SecureRandom.hex(32)) }
```
2. Update your Rack app to use a session middleware before the Web UI:
```ruby
use Rack::Session::Cookie, secret: File.read(".session.key"), same_site: :strict, max_age: 86400
run Sidekiq::Web
```


bf4:
@mperham you may want to update https://github.com/mperham/sidekiq/wiki/Monitoring#forbidden (I presume you'd rather do it than me :)

nice release

PR #4806, opened 2021-02-09
Add process/thread count summary to Busy page

mperham:
Sadly my design skills are minimal so it's not very attractive. CSS help appreciated if anyone wants to polish.

Screen Shot 2021-02-09 at 3 49 51 PM

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -5,7 +5,7 @@
 HEAD
 ---------
 
-- Add process/thread count summary to Busy page [#]
+- Add process/thread count summary to Busy page [#4806]
 
 6.1.3
 ---------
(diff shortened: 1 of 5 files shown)


mperham:
Here's another option for UI: a new "Cluster Summary" area at the top:

Screen Shot 2021-02-18 at 10 22 19 AM

Yet another option:

Screen Shot 2021-02-18 at 10 32 11 AM

mperham:
Latest

Screen Shot 2021-02-18 at 1 22 55 PM

sjahu:
🎉 latest version using `stats-container`/`stat` looks good!

Another simple approach would be just adding a "total" row to the existing table, since all the stats are derived from it anyway.
image
Totally depends on how prominent you want the summary to be, though; your way's probably better!

mperham:
I thought about that. This change request comes from someone running hundreds of Sidekiq processes so finding the end of that table is actually tough, that’s why I was trying to put this info at the top. 🤷‍♂️

sjahu:
Makes sense 👍

mperham:
I think I will add a sum of RSS as another stat to fill out the row; that might help people detect memory changes.

PR #4819, opened 2021-02-18
Change the background colour for the RSS info badge to be less eye-catching

sjahu:
3b5ae30c4e5e9e760268243ab5c14664a2f8d236 added a process RSS column to the table on the busy page. The red badge beside column header is useful for linking users to more information, but it's quite eye-catching. Since the badge doesn't indicate a notification or status change, it would be nice if it stood out a little less from the background.

This PR changes the background colour to light grey in normal mode and dark grey in dark mode. I also changed the class name from "circled" to "info-circle" since, with the more discreet colouring, I feel like it's less generally useful.

Before:
image
image

After:
image
image

### DIFF
--- a/web/assets/stylesheets/application-dark.css
+++ b/web/assets/stylesheets/application-dark.css
@@ -147,3 +147,14 @@ input {
   fill: #ddd;
   color: #ddd;
 }
+
+.info-circle {
+  color: #282828;
+  background-color: #555555;
+  border-radius: 50%;
+  text-align: center;
+  vertical-align: middle;
+  padding: 3px 7px;
+  font-size: 0.7em;
+  margin-left: 5px;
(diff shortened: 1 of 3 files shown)


mperham:
This is really well done, I'll pull this into an element in Sidekiq Enterprise too which uses an info-circle too.

#4806 needs some CSS polish too if I could ask for your help. I'll merge it and you can run master with `gem 'sidekiq', github: 'mperham/sidekiq'` to see the change on the Busy page. If you don't have time, I'll probably update the summary numbers to be info-circles too, if only to give them a little visual flair.

mperham:
That should be:

> gem 'sidekiq', github: 'mperham/sidekiq', branch: 'busy_summary'

PR #4861, opened 2021-04-07
middleware upgrade

dpep:
I'm a big fan of sidekiq and was looking through the sidekiq middleware for inspiration.  I noticed a few areas that could be optimized, if you're open to it?

```ruby
remove(klass) if exists?(klass)
```
seems redundant, since each do an O(n) check for the existence of klass

```ruby
retrieve.dup
```
also seems redundant, since each call to `retrieve` will create brand new Entry instances

### DIFF
--- a/lib/sidekiq/middleware/chain.rb
+++ b/lib/sidekiq/middleware/chain.rb
@@ -77,6 +77,7 @@ module Sidekiq
       end
 
       def initialize
+        @entries = nil
         yield self if block_given?
       end
 
@@ -117,7 +118,7 @@ module Sidekiq
       end
 
       def empty?
-        !instance_variable_defined?(:@entries) || entries.empty?
+        @entries.nil? || @entries.empty?
       end
 
       def retrieve
--- a/lib/sidekiq/middleware/chain.rb
+++ b/lib/sidekiq/middleware/chain.rb
@@ -77,7 +77,6 @@ module Sidekiq
       end
 
       def initialize
-        @entries = nil
         yield self if block_given?
       end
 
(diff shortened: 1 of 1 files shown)


lib/sidekiq/middleware/chain.rb
```diff
@@ -144,6 +143,8 @@ def invoke(*args)
       end
     end
 
+    private
```
dpep, line 146: this might make future changes easier?

lib/sidekiq/middleware/chain.rb
```diff
@@ -118,7 +117,7 @@ def exists?(klass)
       end
 
       def empty?
-        @entries.nil? || @entries.empty?
+        !instance_variable_defined?(:@entries) || entries.empty?
```
natematykiewicz, line 120: You can still keep this as `@entries.nil?`. An undefined instance variable is nil. Also, if the variable _is_ defined, but is nil, this method will return `true`.

dpep: I tried that originally, but there was a bunch of log spew:
```
sidekiq/lib/sidekiq/middleware/chain.rb:121: warning: instance variable @entries not initialized
```

I agree someone was likely trying to avoid allocating the array unless needed (which seems right).  this certainly isn't as clean, so happy to revert this part of it

dpep:
@natematykiewicz what's our next step?

natematykiewicz:
Hey @dpep! I'm not a maintainer, so I'm not sure. I was just trying to help with one thing I noticed.

mperham:
LGTM. Thanks!

PR #4862, opened 2021-04-07
Fix incorrect button rounding

### LINKED ISSUE #4858: "Quiet" button awkward border radius
Ruby version: 2.5.1
Rails version: 6.0.3.6
Sidekiq / Pro / Enterprise version(s): 6.2.0

Something in 6.2.0 made the left half of the button "Quiet" button not round, and it looks a bit awkward.

# 6.1.3
Screen Shot 2021-04-05 at 12 18 19 PM

# 6.2.0
Screen Shot 2021-04-05 at 12 18 15 PM

natematykiewicz:
.btn-group uses :first-child and :last-child to determine which buttons should have rounded corners and which ones should have square corners. The "Quiet" button had a square left corner, because the first child was a CSRF hidden field tag.

Keeping only .btn elements inside the .btn-group means that the 1 or 2 buttons are rounded as expected. When there's no "Quiet" button, the "Stop" button has 4 round corners. When both buttons are present, the corners between the buttons are square.

Fixes #4858

I traced this problem to this diff. https://github.com/mperham/sidekiq/commit/f0ddebc7406fe4b78333fbc5f14488a38b9ccebc#diff-d68d7f0b05d0be529c7d2493225a3cd8e5358f0c6327b7dfc8ead5f2927a1ba1L50-L59

The buttons now look like this:

When both buttons are present
Screen Shot 2021-04-06 at 10 49 55 PM

When only Stop is present
Screen Shot 2021-04-06 at 10 55 46 PM

### DIFF
--- a/web/views/busy.erb
+++ b/web/views/busy.erb
@@ -78,9 +78,10 @@
         <td><%= process['busy'] %></td>
         <td>
           <form method="POST">
+            <%= csrf_tag %>
+            <input type="hidden" name="identity" value="<%= process['identity'] %>"/>
+
             <div class="btn-group pull-right flip">
-              <%= csrf_tag %>
-              <input type="hidden" name="identity" value="<%= process['identity'] %>"/>
               <% unless process.stopping? %><button class="btn btn-warn" type="submit" name="quiet" value="1"><%= t('Quiet') %></button><% end %>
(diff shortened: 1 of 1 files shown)


mperham:
Sharp detective work!

natematykiewicz:
Thanks for the quick merge!

PR #4866, opened 2021-04-08
Make Dashboard and Busy page similar to other pages

mobilutz:
This adds some small styling changes to the Dashboard and Busy page.

Dashboard looks like this afterwards - NOTE white boxes are because of strange policies ...
Screenshot 2021-04-08 at 11 55 57

Changes are:
* make `History` a headline row similar to other pages
* make `Redis` a headline row similar to other pages
* add `Redis` to locale files so it can be translated if needed

Busy page looks like this:
Screenshot 2021-04-08 at 11 54 59

Changes are:
* center stats boxes just like on dashboard

### DIFF
--- a/web/assets/stylesheets/application.css
+++ b/web/assets/stylesheets/application.css
@@ -277,7 +277,7 @@ table .table-checkbox label {
 
   .navbar .nav a {
     text-align: center;
-  }  
+  }
 
 }
 
@@ -400,7 +400,7 @@ img.smallogo {
   padding: 10px 0;
 }
 
-.redis-wrapper {
+.stats-wrapper {
   width: 100%;
   text-align: center;
 }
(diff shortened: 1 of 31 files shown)


mperham:
Redis is a name, it is not a translated word anymore than “Mike” would be translated. 

Please verify changes with a narrow “mobile” browser window.

mobilutz:
@mperham I removed Redis from translation.

And here are screenshots from the two pages in different sizes.
The code only reuses existing classes and existing HTML, so nothing new was added.
Screenshot 2021-04-09 at 11 42 59
Screenshot 2021-04-09 at 11 43 13
Screenshot 2021-04-09 at 11 43 39
Screenshot 2021-04-09 at 11 44 07

mperham:
There are still locale changes. Please remove them all.

Can you look through the rest of the tab pages and make all headers consistent? I would prefer them all to use a semantic tag like <h3> rather than a marked up <div>.

mobilutz:
Thanks @mperham for using the change and sorry for not getting back to your requests.

Two kids in quarantine makes live a tiny bit more complicated 😉

PR #4879, opened 2021-04-19
Prefer warn over puts so these can be suppressed

qrush:
So we're using `Sidekiq::Testing.inline!` for some interesting things @wistia - I'd love to be able to suppress the `puts` statements since we now require this file in development + staging environments. 🙏🏼

### DIFF
--- a/lib/sidekiq/testing.rb
+++ b/lib/sidekiq/testing.rb
@@ -338,7 +338,7 @@ module Sidekiq
 end
 
 if defined?(::Rails) && Rails.respond_to?(:env) && !Rails.env.test? && !$TESTING
-  warn("**************************************************")
-  warn("⛔️ WARNING: Sidekiq testing API enabled, but this is not the test environment.  Your jobs will not go to Redis.")
-  warn("**************************************************")
+  warn("**************************************************", uplevel: 1)
+  warn("⛔️ WARNING: Sidekiq testing API enabled, but this is not the test environment.  Your jobs will not go to Redis.", uplevel: 1)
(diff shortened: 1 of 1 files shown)


mperham:
Long time, Nick, welcome!

whaddya think about adding these keywords that can act as more context?

https://ruby-doc.org/core-3.0.1/Kernel.html#method-i-warn

qrush:
@mperham it's been too long 😁 

Which `uplevel` would you prefer?

mperham:
Whatever prints the best output for your usage. What's reasonable?

qrush:
`uplevel: 1` sounds good to me - the only example I can find in my bundle is via rexml:

https://github.com/ruby/rexml/blob/e941ff17ed3dad428d946b15524bb3529e684266/lib/rexml/cdata.rb#L60-L62

mperham:
In Ruby 3.0 I think you can silence warnings by category. What category? :sidekiq? :sidekiq_testing?

mperham:
I’m trying to figure out what the semantics of category are. It’s different from what I thought. Looks like there’s only two categories: deprecated and experimental. Neither fits this use.

https://bugs.ruby-lang.org/issues/17122

qrush:
Looks that way to me. We are still on 6.0 so need to spend a bit more time to get up to this patch- almost there!

mperham:
Also, I think we should remove the lines of stars. It's unnecessary output.

qrush:
Thanks @mperham !!

mperham:
Thanks for putting up with my comments. I love a gleaming, highly polished bike shed.

PR #4881, opened 2021-04-21
Fix a typo in lt locale file

redross:
As you can see from the translation of `Stop`, the correct spelling should be `Sustabdyti`. There is no such word as `Sustbadyti` (also `badyti` means `stabbing` :dagger: , which is kind of funny)


mperham:
Stab All. Ha! Thanks.

PR #4882, opened 2021-04-21
Memoize process count in Poller

jonhyman:
Avoids scanning the processes set and trying to cleanup on every Poller#wait call

### DIFF
--- a/lib/sidekiq/scheduled.rb
+++ b/lib/sidekiq/scheduled.rb
@@ -117,7 +117,10 @@ module Sidekiq
         # As we run more processes, the scheduling interval average will approach an even spread
         # between 0 and poll interval so we don't need this artifical boost.
         #
-        if process_count < 10
+        # Memoize the process count to avoid calling SCARD/HGET on all the processes everytime #wait is called
+        # since there may be a lot of processes online.
+        @process_count ||= process_count
+        if @process_count < 10
           # For small clusters, calculate a random interval that is ±50% the desired average.
           poll_interval_average * rand + poll_interval_average.to_f / 2
         else


mperham:
I’m uncomfortable with either calling every time or only calling it once. The process count might change over time. Should we cache this for five minutes or some reasonable length?

jonhyman:
I don't know the history of why you added the special case. To me, it seems arbitrary -- why 10 processes, why not 20 -- that's meant to help smooth out random numbers. Processes can run for a long time so even with <10 processes running for 24h that's a lot of wait() calls and I'd assume that a case where you end up with N*M messing consistently messing you up is probably rare because you have `rand` there. While you do provide an example of 2 processes sleeping, you will get some jitter by the fact that Ruby is single-threaded and your jobs are going to interfere with your randoms anyway. Over a short period of time the jitter will separate the processes. I've written a quick example below.

> Should we cache this for five minutes or some reasonable length?

Overall if that's what you're most comfortable with, 5m would be fine by me.

```ruby
# Fibonacci will simulate a job doing work
def fib(n)
  if n == 0 || n == 1
    return n
  end
  return fib(n - 1) + fib(n - 2)
end

# This simulates the poller, it sleeps for 5 seconds and we measure how long it takes between runs.
# Because there's work happening on another thread, the time difference between runs will vary slightly from 5 seconds.
t1 = Thread.new do
  while true
    if @et
      puts((@et - Time.now).round(2))
    end
    @et = Time.now
    sleep(5)
  end
end

t2 = Thread.new do
  while true
    fib(Random.rand(60))
    sleep(2)
  end
end
```

YMMV but here's an output from me running it for a few seconds. So if you have two processes here and they're doing some CPU-based work, the `wait` calls will diverge from each other within like a minute and then you should get closer to your averages. Even if they both sleep for 8 seconds, they'll probably be off by enough that you still hit your 5 second average most of the runs in a long time period.

```
-5.0
-5.11
-5.0
-5.0
-5.0
-5.11
-5.11
-5.13
-5.12
-5.11
-5.21
-5.2
-5.12
-5.06
-5.1
-5.01
-5.02
```

mperham:
What do you think about caching the value for 10 calls. We guarantee a 90% reduction in calls and still check regularly.

mperham:
Ping?

jonhyman:
Wow sorry I totally missed all the comments due to going on PTO shortly after this. Apologies, but I think you ended up in a good spot.

PR #4884, opened 2021-04-23
Add updated translations for spanish language

andreslazo:
This PR adds some missing translations for Spanish language:

Captura de Pantalla 2021-04-23 a la(s) 10 55 28

Also adds three words (Utilization, Process and Name) that were missing from locale files.


mperham:
Thank you!

PR #4886, opened 2021-04-27
Update sidekiq.service

cassiompf:
Adding configuration to work with the gem 'capistrano/sidekiq' to be able to deploy automatically and restart the sidekiq service.

Repository gem 'capistrano/sidekiq': https://github.com/seuros/capistrano-sidekiq

### DIFF
--- a/examples/systemd/sidekiq.service
+++ b/examples/systemd/sidekiq.service
@@ -42,13 +42,19 @@ WatchdogSec=10
 
 WorkingDirectory=/opt/myapp/current
 # If you use rbenv:
-# ExecStart=/bin/bash -lc 'exec /home/deploy/.rbenv/shims/bundle exec sidekiq -e production'
+# ExecStart=/bin/bash -lc 'exec /home/deploy/.rbenv/shims/bundle exec sidekiq -e production' -C config/sidekiq.yml
 # If you use the system's ruby:
-# ExecStart=/usr/local/bin/bundle exec sidekiq -e production
+# ExecStart=/usr/local/bin/bundle exec sidekiq -e production -C config/sidekiq.yml
 # If you use rvm in production without gemset and your ruby version is 2.6.5
-# ExecStart=/home/deploy/.rvm/gems/ruby-2.6.5/wrappers/bundle exec sidekiq -e production
+# ExecStart=/home/deploy/.rvm/gems/ruby-2.6.5/wrappers/bundle exec sidekiq -e production -C config/sidekiq.yml
 # If you use rvm in production with gemset and your ruby version is 2.6.5
-ExecStart=/home/deploy/.rvm/gems/ruby-2.6.5@gemset-name/wrappers/bundle exec sidekiq -e production
+ExecStart=/home/deploy/.rvm/gems/ruby-2.6.5@gemset-name/wrappers/bundle exec sidekiq -e production -C config/sidekiq.yml
+
+# To work with the gem 'capistrano/sidekiq'
+# If you are not using this gem, just remove the two lines just below
+# although it will not have any effect on the service if you keep these two lines
+ExecReload=/bin/kill -TSTP $MAINPID
+ExecStop=/bin/kill -TERM $MAINPID
 
 # Use `systemctl kill -s TSTP sidekiq` to quiet the Sidekiq process
 
--- a/examples/systemd/sidekiq.service
+++ b/examples/systemd/sidekiq.service
@@ -42,19 +42,19 @@ WatchdogSec=10
 
 WorkingDirectory=/opt/myapp/current
 # If you use rbenv:
-# ExecStart=/bin/bash -lc 'exec /home/deploy/.rbenv/shims/bundle exec sidekiq -e production' -C config/sidekiq.yml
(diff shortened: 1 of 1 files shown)


examples/systemd/sidekiq.service
```diff
@@ -42,13 +42,19 @@ WatchdogSec=10
 
 WorkingDirectory=/opt/myapp/current
 # If you use rbenv:
-# ExecStart=/bin/bash -lc 'exec /home/deploy/.rbenv/shims/bundle exec sidekiq -e production'
+# ExecStart=/bin/bash -lc 'exec /home/deploy/.rbenv/shims/bundle exec sidekiq -e production' -C config/sidekiq.yml
```
mperham, line 45: This is the default so there is no need to explicitly configure it.

examples/systemd/sidekiq.service
```diff
@@ -42,13 +42,19 @@ WatchdogSec=10
...
+# ExecStart=/home/deploy/.rvm/gems/ruby-2.6.5/wrappers/bundle exec sidekiq -e production -C config/sidekiq.yml
 # If you use rvm in production with gemset and your ruby version is 2.6.5
-ExecStart=/home/deploy/.rvm/gems/ruby-2.6.5@gemset-name/wrappers/bundle exec sidekiq -e production
+ExecStart=/home/deploy/.rvm/gems/ruby-2.6.5@gemset-name/wrappers/bundle exec sidekiq -e production -C config/sidekiq.yml
+
+# To work with the gem 'capistrano/sidekiq'
+# If you are not using this gem, just remove the two lines just below
+# although it will not have any effect on the service if you keep these two lines
+ExecReload=/bin/kill -TSTP $MAINPID
+ExecStop=/bin/kill -TERM $MAINPID
```
mperham, line 57: This is the default, no need.

examples/systemd/sidekiq.service
```diff
@@ -42,13 +42,19 @@ WatchdogSec=10
...
-# ExecStart=/home/deploy/.rvm/gems/ruby-2.6.5/wrappers/bundle exec sidekiq -e production
+# ExecStart=/home/deploy/.rvm/gems/ruby-2.6.5/wrappers/bundle exec sidekiq -e production -C config/sidekiq.yml
 # If you use rvm in production with gemset and your ruby version is 2.6.5
-ExecStart=/home/deploy/.rvm/gems/ruby-2.6.5@gemset-name/wrappers/bundle exec sidekiq -e production
+ExecStart=/home/deploy/.rvm/gems/ruby-2.6.5@gemset-name/wrappers/bundle exec sidekiq -e production -C config/sidekiq.yml
+
+# To work with the gem 'capistrano/sidekiq'
+# If you are not using this gem, just remove the two lines just below
+# although it will not have any effect on the service if you keep these two lines
+ExecReload=/bin/kill -TSTP $MAINPID
```
mperham, line 56: I think it’s a bad idea to change the semantics of reload to map it to quiet.

cassiompf: I got this from the systemd service template in the "sidekiq/capistrano" repository. The gem uses the reload command to reload the sidekiq service and by default the sidekiq service does not have the "reload" command, it only has the start and stop commands. Para fazer funcionar, peguei "ExecReload" e o adicionei ao serviço sidekiq.

In the link below you can see how the gem uses ExecReload: https://github.com/seuros/capistrano-sidekiq/blob/master/lib/generators/capistrano/sidekiq/systemd/templates/sidekiq.service.capistrano.erb

examples/systemd/sidekiq.service
cassiompf: Ok

examples/systemd/sidekiq.service
cassiompf: By default does it use the action mailer queue? I added this because I had to set up an extra queue on the sidekiq with the name "mailers" so that I can make the system for sending emails using the method (deliver_later). Without it, he only creates the queue with the name "default", if I'm not mistaken.

My file sidekiq.yml

My file sidekiq.yml:

`:concurrency: 1
:queues:
  - default
  - mailers`

examples/systemd/sidekiq.service
cassiompf: I removed that line

examples/systemd/sidekiq.service
cassiompf: I found it interesting to leave this information in the service example to better guide people. In my case I was a little lost to solve the problem using the gem 'capistrano/sidekiq'. I believe that if you leave this information visible here, people will not have to go through the same problem that and many other people have already.

mperham: https://stackoverflow.com/a/28504367 How to set the AM queue to “default”.

examples/systemd/sidekiq.service
mperham: I don't like the idea of mapping `reload` to `quiet`. These two words mean very different things. I would suggest that capistrano-sidekiq move away from reload and use `systemctl kill -TSTP` instead.

mperham:
Sorry, I don't see anything I want to keep.

examples/systemd/sidekiq.service
cassiompf: Ok

PR #4904, opened 2021-05-17
Better format class in `JobLogger`

BuonOmo:
# Commit

Since we already have `Sidekiq::Job#display_class` for formatting correctly the displayed class, I used it in `Sidekiq::JobLogger` to avoid _quite_ useless log information such as:

```
... class=Sidekiq::Extensions::DelayedClass ...
```

# Discussion

Hi Mike,

IMHO, this pull request has a small naming drawback, `Foo.call` or `Mailer#welcome` are not really classes. In the UI, you use `job` to better describe them.

Hence I've reported a few alternatives I have in mind, please tell me if you prefer any of those over the current proposal.

| proposition | breaking change | issue |
| ----------- | :-------------: | ----- |
| 1. Keep the old class and add a new `job` to the log. | | Logs too long |
| 2. Remove the old class and add a new `job` to the log. | yes, large | |
| 3. (current patch) Keep the class key, change the behaviour | yes, small | Naming |

Since it is my first contribution here, please tell me if I have overlooked any part of the PR (changelog, test, doc, perf, anything..).

Best


mperham:
The Sidekiq server process itself does not load or use the API at all so you can't require and reuse that Web UI display code. I'm ok with (3) but I don't have a sense for how important this change is vs if it will break anyone's code.

BuonOmo:
Thanks for the quick reply!

> I'm ok with (3) but I don't have a sense for how important this change is vs if it will break anyone's code.

Yeah it can hurt some logs. Maybe next major is OK (we've monkey patched locally, anyone else can do so while waiting). And I guess if one if willing for that feature sooner they can :+1: the PR or/and add a comment ?

> The Sidekiq server process itself does not load or use the API at all so you can't require and reuse that Web UI display code.

I'm not very much familiar with the architecture of sidekiq, I guess I could either:

1. Extract `#display_class` to `lib/sidekiq/util.rb` (not sure it is ok though, and it is maybe too much for the web ui)
2. Extract `Sidekiq::Job` to a `lib/sidekiq/api/job.rb` file, which would be small enough to be loaded by server. (only issue is it requires `zlib`, but this is also required by `lib/sidekiq/job_retry.rb` hence LGTM)
3. extract `#display_class` to `lib/sidekiq/job_logger.rb` and require it from `lib/sidekiq/api.rb` (however this creates weird relations, and add some useless code to the ui)
4. create a new `lib/sidekiq/job_util.rb` (or `../util/job.rb` file containing `#display_class` and maybe more.
5. copy paste and add a comment on both sides, which is simpler now, but less maintainable of course..

For now, (2.) is my favorite, but I don't know if it would benefit the codebase enough vs the needed size. I like (4.) as well.

BuonOmo:
@mperham I've chosen the option `(2.)` for now since it is my favorite and I didn't want the PR to stay in an unacceptable state. I can still implement another solution if need be.

mperham:
Sorry for the delay. I would extract the display_* methods to a Module and include it in Sidekiq::Job and Sidekiq::JobLogger. I think Sidekiq::JobUtils is a reasonable name.

BuonOmo:
@mperham sorry I forgot to ping you, changes done! I've written a `JobUtils` class. 

Since both `Job` and `JobLogger` does not store the job in the same way, I've used class methods directly rather than some magic module having tight coupling with both classes.

mperham:
I appreciate your hard work on this. I think there is a very different and possibly simpler method.

What if you start over and adjusted generic_proxy.rb so it added `"display_class" => "#{target}.#{name}"` to the job hash and then used that in api.rb and job_logger.rb? I think that would be way less code change.

BuonOmo:
@mperham I agree this is less code change, however, it looks like this will add overhead in a lot of places for a log (e.g: a bigger hash).

And we would still have to handle the `ActiveJob` case separately, hence there would still be a great amount of code.

If you are still sure you want that I should have some time to do it in the next 2 weeks

mperham:
No worries, I took care of it but gave you credit because your issue was the inspiration. Thank you!

BuonOmo:
LGTM, do you consider releasing it soon?

mperham:
Probably not for a while.

PR #4908, opened 2021-05-20
Remove double space character

olimart:


olimart:
Hi @mperham do you mind explaining? did I miss something?

mperham:
What’s the point? It’s a meaningless change.

olimart:
whaoo. wasn't expecting such rude reply.
I guess the definition of meaningful/meaningless could be added to the contributing guidelines then to avoid such situation.

PR #4909, opened 2021-05-26
Add alert-danger and alert-info support for dark mode

noaelad:
In our usage of sidekiq, we've added some custom tabs to the UI and we're using `alert-danger` and `alert-info` on those pages. The css for these does not look great in dark mode, this PR offers a dark-mode friendly style for those tags.

Here's an example of an `alert-danger` with the proposed style:
image

Here's an example of an `alert-info` with the proposed style:
image

### DIFF
--- a/web/assets/stylesheets/application-dark.css
+++ b/web/assets/stylesheets/application-dark.css
@@ -64,6 +64,14 @@ table {
   background-color: #484;
 }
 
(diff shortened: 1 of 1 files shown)


mperham:
Since these are classes that are in the base CSS, it makes sense to add dark support for them too. Thank you!

PR #4919, opened 2021-06-04
Add missing japanese locales

h-shima:
Hi @mperham, thanks for providing such nice gems.

I noticed that some Japanese translations were missing, so I added them.


mperham:
Thank you!

PR #4935, opened 2021-07-01
Feature api for restricting stats calls performed

mijoharas:
`Sidekiq::Stats` currently makes 7 + number_of_processes + number_of_queues redis calls. (NOTE: due to pipelining, it's only 4 network trips, so it's not too bad but if you're running hundreds of processes it'd still be nice to skip 100 redis calls if you don't need them).

This Pr proposes an API to allow you to optionally remove `number_of_processes` redis calls from stats initialization. We could additionally allow the user to `skip_enqueued` to reduce the number of redis calls to just 7 at the users discretion.

Let me know what you think, and I can make changes + add documentation as relevant, and could add the `skip_enqueued` as well if you think that would also be useful.

Open questions:
1. Do we want this (my company does, but we can obviously just maintain a fork).
2. Is `skip_workers: true` the right name for this variable?
3. Do we want to save this `skip_workers` argument as an instance variable, so subsequent calls to `fetch_stats!` don't need to explicitly provide this, or leave the api as-is.
4. Do we want to extend this to skip a roundtrip to redis for other things? if so, which things? queues seems like an obvious option, as if a user does each, they'll skip 

(it seemed faster to make a PR to just show what I meant rather than an issue suggesting the change, though obviously it's just a proposal)

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -7,8 +7,8 @@ require "base64"
 
 module Sidekiq
   class Stats
-    def initialize
-      fetch_stats!
+    def initialize(skip_workers: false)
+      fetch_stats!(skip_workers: skip_workers)
     end
 
     def processed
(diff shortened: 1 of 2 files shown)


mijoharas:
makes sense, is something like https://github.com/mperham/sidekiq/pull/4936 what you were thinking of? (closing this branch)

PR #4936, opened 2021-07-01
Quick stats api.

mijoharas:
follow on from https://github.com/mperham/sidekiq/pull/4935

This api seems simpler (I can just call `Sidekiq::Stats::QuickStats.new.fetch_stats!` for my usecase, and we reduce duplication with the main `Stats` api, since there's an overlap).

What do you think?

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -32,7 +32,6 @@ module Sidekiq
     end
 
     def enqueued
-      maybe_fetch_stats_slow!
       stat :enqueued
     end
 
@@ -41,7 +40,6 @@ module Sidekiq
     end
 
     def workers_size
-      maybe_fetch_stats_slow!
       stat :workers_size
     end
 
@@ -139,13 +137,11 @@ module Sidekiq
 
     private
 
-    def maybe_fetch_stats_slow!
-      return unless stat(:enqueued).nil? || stat(:workers_size).nil?
-
-      fetch_stats_slow!
-    end
-
     def stat(s)
+      if (s == :enqueued || s == :workers_size) &&
+         @stats.present? && @stats[s].nil?
+        fetch_stats_slow!
+      end
       @stats[s]
     end
 
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -8,7 +8,7 @@ require "base64"
 module Sidekiq
   class Stats
     def initialize
-      fetch_stats!
+      fetch_stats_fast!
     end
 
     def processed
@@ -32,6 +32,7 @@ module Sidekiq
     end
 
     def enqueued
+      maybe_fetch_stats_slow!
       stat :enqueued
     end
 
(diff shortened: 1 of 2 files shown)


lib/sidekiq/api.rb
```diff
@@ -52,17 +52,7 @@ def queues
...
-          conn.get("stat:processed")
-          conn.get("stat:failed")
-          conn.zcard("schedule")
-          conn.zcard("retry")
-          conn.zcard("dead")
-          conn.scard("processes")
-          conn.lrange("queue:default", -1, -1)
-        end
-      }
+      pipe1_res = QuickStats.new.fetch_stats!
```
mijoharas, line 55: note, this isn't the same as the previous `pipe1_res` we threw it into a hash to and did a few basic conversions for convenience.

mijoharas:
Happy to do that, but it does mean a fair of duplication, as all it's doing is extracting the first section of the `fetch_stats!` call. Are you sure?

If so, I guess there's not much benefit from having it rolled into the sidekiq api. (I was thinking the advantage would be keeping it in sync with changes to sidekiq's redis data representation, and reducing duplication. I guess it would still accomplish the first one of those)

mperham:
Sidekiq's redis data format is pretty ossified at this point. If you want to split Stats into a fast method and a slow method where the fast method is called in initialize but the slower stats are lazy loaded, I'd be fine with that.

mijoharas:
Sounds great. I'm heading away from my computer right now but I'll make that change tomorrow. Thanks as always for being so responsive. :pray:

mijoharas:
How does this look? If it's good I'll copy the test I wrote from the other PR as I don't see an existing one on `worker_size`. Should I also add this to the changelog or do anything else?

lib/sidekiq/api.rb
```diff
@@ -51,7 +53,8 @@ def queues
       Sidekiq::Stats::Queues.new.lengths
     end
 
-    def fetch_stats!
+    # O(1) redis calls
```
mperham, line 56: Good comment

lib/sidekiq/api.rb
```diff
@@ -32,6 +32,7 @@ def dead_size
     end
 
     def enqueued
+      maybe_fetch_stats_slow!
```
mperham, line 35: I don't think you need these explicit calls. Just make `stat` smarter?

mijoharas: done in https://github.com/mperham/sidekiq/pull/4936/commits/a3028fc6e5d14179e0599e2001c92816e8ef985b

mperham:
Great work, thank you!

ZackMattor:
Howdy @mperham and @mijoharas. It appears this may break the pre-existing `fetch_stats!` API. It previously returned an object and now it appears to returns an integer. 

We upgraded from 6.2.1 -> 6.2.2 and noticed this breaking change.

mperham:
Thanks for pointing that out. APIs will be refactored, in this case for performance reasons. `fetch_stats!` is not an API that I consider documented for public consumption, per the API wiki page. If you are using more "internal" APIs, it's possible that things will break between versions.

I'm also happy to see PRs which make more API methods private, so this is clearer to people in the future.

Issue #4955, opened 2021-08-06
Rename Sidekiq::Worker to Sidekiq::Job


mperham:
The term `worker` within the Sidekiq ecosystem is a net negative. Replace instances in the wiki and codebase with more precise terminology, e.g.

Sidekiq::Worker -> Sidekiq::Job
worker -> process, thread, job depending on context

seuros:
This is going to be a major breaking change. Most addons of Sidekiq will need to be updated to work with the next version.  Can you explain what the net negative in the term 'worker' ?

mperham:
I know, the plan is to have a very long backward compatibility period. Ideally we make migration as simple as search/replace "Sidekiq::Worker" -> "Sidekiq::Job". My initial thought is:

* For 6.3.0, S::Worker will be an alias for S::Job
* For 6.4.0 it will print a deprecation warning
* For 7.0.0 it will break.

Restoring compatibility is as simple as `Sidekiq::Worker = Sidekiq::Job` in your initializer.

"worker" is ambiguous, people use it to refer to processes ("I started a new worker"), types of jobs, executing threads ("I've got 10 workers"), jobs ("I enqueued 10 workers"), etc. It's a mess.

Does that make more sense?

mperham:
The only major issue I've seen so far is that ActiveJob owns the `app/jobs` directory whereas Sidekiq owns `app/workers`. I don't have a solution for that but I'm not sure it's a major issue: is it strange to see `app/workers/hard_job.rb`?

seuros:
Hmm, I think this change will cause a major refactoring effort , a lot of gems reference to that module. 
For example, it will render [ActiveJob](https://github.com/rails/rails/blob/main/activejob/lib/active_job/queue_adapters/sidekiq_adapter.rb#L39) incompatible with Sidekiq 7 unless we monkey patch it .

For folder naming, if i had to not use app/workers and since we agreed that app/jobs if for ActiveJob wrappers, i will use app/async_tasks or app/sidekiq_tasks.

mperham:
I appreciate this feedback, you always have good advice.

I do want to create and document Sidekiq::Job. You are right that we can't realistically drop Sidekiq::Worker anytime soon, possibly forever. I will make the changes to use Sidekiq::Job everywhere but I will keep Sidekiq::Worker as a supported public API for the foreseeable future.

mperham:
[...]

PR #4956, opened 2021-08-07
Add Sidekiq::Job

mperham:
* Provide Sidekiq::Worker alias for backwards compat
* Rename API's Sidekiq::Job to Sidekiq::JobRecord
* Adjust internals to use Sidekiq::JobRecord

Implements #4955


seuros (approved):

Changes.md
```diff
@@ -5,6 +5,15 @@
 HEAD
 ---------
 
+- Add `Sidekiq::Job` alias for `Sidekiq::Worker` [#4995]
```
seuros, line 8: - Rename `Sidekiq::Worker` to `Sidekiq::Job` [#4995]
- Alias `Sidekiq::Worker` to `Sidekiq::Job` to keep backward compatibility. [#4995]

seuros: That way it clear that Job is just the old Worker instead of a new version with same API.

mperham:
We'll do this slower, in multiple stages.

PR #4957, opened 2021-08-07
Allow sidekiq_retry_in to return a range

maurycy:
The idea is to make it easier to implement (and reason about) exponential (and not only!) [backoffs with jitter](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/).

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -5,6 +5,7 @@
 HEAD
 ---------
 
+- Sidekiq::Worker `sidekiq_retry_in` block can now return a range [#4957]
 - Minimize scheduler load on Redis at scale [#4882]
 - Improve logging of delay jobs [#4904, BuonOno]
 - Minor CSS improvements for buttons and tables, design PRs always welcome!
--- a/lib/sidekiq/job_retry.rb
+++ b/lib/sidekiq/job_retry.rb
@@ -215,7 +215,12 @@ module Sidekiq
 
     def delay_for(worker, count, exception)
       if worker&.sidekiq_retry_in_block
-        custom_retry_in = retry_in(worker, count, exception).to_i
+        range_or_interval = retry_in(worker, count, exception)
+        custom_retry_in = if range_or_interval.is_a?(Range)
+          rand(range_or_interval)
+        else
+          range_or_interval
+        end.to_i
         return custom_retry_in if custom_retry_in > 0
       end
       seconds_to_delay(count)
--- a/test/test_retry.rb
+++ b/test/test_retry.rb
@@ -287,6 +287,8 @@ describe Sidekiq::JobRetry do
             Sidekiq::JobRetry::USE_DEFAULT_RETRY_FORMULA
           when ArgumentError
             count * 4
+          when ZeroDivisionError
+            (count..count * 2)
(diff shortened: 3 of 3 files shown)


maurycy:
@mperham Any reason why? It would make it easier to reason about exponential backoffs with jitter. (Thank you for your work on sidekiq!)

mperham:
“No description provided.”

if you make no effort to explain your PR to others, I make no effort to review it.

maurycy:
@mperham Updated. You closed it in 5 minutes while I was still pondering the best description. Please let me know if you want me to expand on this description.

mperham:
👍🏻

mperham:
I appreciate the feedback. In the future I’ll give people a while to finalize the PR before taking any actions.

mperham:
What happens if the block returns nil? 0?

maurycy:
Yes:

```
irb(main):003:0> nil.is_a?(Range)
=> false
irb(main):004:0> nil.to_i
=> 0
```

hence `> 0`:

https://github.com/mperham/sidekiq/pull/4957/files#diff-a2a78bd5586e1300341bb91dc8be382aa98126dcde11bd507fe9886d2e643ee7L219

maurycy:
Other reasons why I think this change makes sense:

 * random retries should be encouraged, even `seconds_to_delay` already uses `rand`,
 * `sidekiq_retry_in` is always a bit probabilistic, because there is no way to guarantee a retry exactly at a given time.

mperham:
Well I wonder then why let the user return a range but rather build jitter
into sidekiq directly. So if the user returns 30 we would actually jitter
it between 20 and 40 for instance.

maurycy:
Why not both? Some users might want to retry within a short period, others prefer a longer one, and `range` seems like the easiest and the most intuitive way to configure this, without breaking anything and ready to use immediately. Moving in the direction of building in explicit randomness by default for everyone needs some consideration: [which jitter algorithm](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/)? leave [the current seconds_to_delay formula](https://github.com/mperham/sidekiq/blob/3330df0ee37cfd3e0cd3ef01e3e66b584b99d488/lib/sidekiq/job_retry.rb#L225) as is? how to let users configure the interval? what API?  I like the direction, though. What do you think?

mperham:
I strongly believe software needs less knobs, not more. I'd prefer the user return a single Integer and we add jitter internally. That's how Sidekiq's retry subsystem works today. Is the root issue here "sidekiq_retry_in should include jitter so lots of jobs don't pile up at the exact same time"?

maurycy:
> I strongly believe software needs less knobs, not more. I'd prefer the user return a single Integer and we add jitter internally.

There's still a need for configuration I think.

> Is the root issue here "sidekiq_retry_in should include jitter so lots of jobs don't pile up at the exact same time"?

Yes.

mperham:
Ok, I think the simpler solution is to add jitter to the return value from sidekiq_retry_in. perform_in can't add jitter as people really depend on that timing but retries are fair game and already do.

PR #4967, opened 2021-08-14
Add private to Web UI Cache-Control to prevent request collapsing

shaunbennett:
As a follow up to #4966, I realized it's also important to add `private` to the Cache-Control header as well to prevent [request collapsing](https://developer.fastly.com/learning/concepts/request-collapsing/). Even with `no-store`, Fastly (and I assume other places) can still collapse requests.

Note that this is only important if the Web UI is behind some form of authentication, since it provides the possibility of sending authenticated data to non-authenticated users under very specific circumstances.

This matches the example [here](https://developer.fastly.com/solutions/examples/prohibit-browser-caching) that fastly shows to prevent downstream caching + request collapsing.

### DIFF
--- a/lib/sidekiq/web/action.rb
+++ b/lib/sidekiq/web/action.rb
@@ -68,7 +68,7 @@ module Sidekiq
     end
 
     def json(payload)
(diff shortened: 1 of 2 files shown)


mperham:
Thanks for the followup!

PR #4976, opened 2021-08-27
Persist live poll status in browser localStorage

sjahu:
For reference:
Screen Shot 2021-08-27 at 15 20 55
Screen Shot 2021-08-27 at 15 21 03

This PR changes the way that the live polling feature is implemented so that the status (i.e. whether polling is enabled or disabled) is persisted across sessions and through page changes. Currently, if live polling is on but you navigate to a different Sidekiq tab (e.g. from "Busy" to "Enqueued"), live polling becomes disabled and you have to manually turn it back on. By leveraging the browser's `localStorage` property similarly to how the live polling interval is already handled, the status can be persisted.

I considered building on the existing `?poll=true` strategy, but I thought tacking that parameter on to every link on the page would be less elegant than just using `localStorage`.

I've added comments in the files view in case the diffs alone are insufficient to explain changes.

### DIFF
--- a/test/test_web.rb
+++ b/test/test_web.rb
@@ -428,12 +428,6 @@ describe Sidekiq::Web do
     assert_match(/#{msg['args'][2]}/, last_response.body)
   end
 
-  it 'calls updatePage() once when polling' do
-    get '/busy?poll=true'
-    assert_equal 200, last_response.status
-    assert_equal 1, last_response.body.scan('data-poll-path="/busy').count
-  end
-
   it 'escape job args and error messages' do
     # on /retries page
     params = add_xss_retry
@@ -646,12 +640,6 @@ describe Sidekiq::Web do
       get '/queues/foo'
       assert_equal 200, last_response.status
       assert_match(/#{params.first['args'][2]}/, last_response.body)
-    end
-
-    it 'handles bad query input' do
-      get '/queues/foo?page=B<H'
-      assert_equal 200, last_response.status
-      assert_match(/B%3CH/, last_response.body)
     end
   end
 
--- a/web/assets/javascripts/application.js
+++ b/web/assets/javascripts/application.js
@@ -16,13 +16,9 @@
 
 Sidekiq = {};
 
-$(function() {
-  var pollpath = $('body').data('poll-path');
-  if (pollpath != "") {
-    var ti = parseInt(localStorage.timeInterval) || 2000;
-    setTimeout(function(){updatePage(pollpath)}, ti);
-  }
+var livePollTimer = null;
 
+$(function() {
   $(document).on('click', '.check_all', function() {
     $('input[type=checkbox]', $(this).closest('table')).prop('checked', this.checked);
   });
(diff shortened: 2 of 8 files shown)


test/test_web.rb
```diff
@@ -647,12 +641,6 @@ def perform(a, b)
       assert_equal 200, last_response.status
       assert_match(/#{params.first['args'][2]}/, last_response.body)
     end
-
-    it 'handles bad query input' do
```
sjahu, line 651: This test depended on the link generated for the live poll button being present in the response body. Rewritten and moved to `test/test_web_helpers.rb`.

web/assets/javascripts/application.js
```diff
@@ -35,7 +31,16 @@ $(function() {
     $($(this).attr('data-target')).toggle();
   });
 
+  $(document).on("click", ".live-poll", function() {
+    localStorage.poll ^= 1;
```
mperham, line 35: Better to make this more descriptive so local storage viewers have some sense of what this variable is for, e.g. `sidekiq_live_poll`?

And maybe this is the Gopher in me but I don't like seeing numeric operations here and boolean operations down below, e.g. `poll = true`. Is this idiomatic?

mperham (commented):
Does live poll happen on the root dashboard URL? It should not as the graphs automatically poll for data and having the page refresh will break the graphs.

web/assets/javascripts/application.js
sjahu: > make this more descriptive

Good call; I added a commit changing this to `localStorage.sidekiqLivePoll`, and the existing `localStorage.timeInterval` to `localStorage.sidekiqTimeInterval` .

Re the numeric/boolean mixing: this was a sloppy way of coping with JS comparisons and the way that localStorage stores values as strings (after setting `localStorage.foo = true;`, `localStorage.foo` returns `"true"` and `"true" != true` but `"1" == true` 🤦). Anyway, I changed it to use "enabled" and "disabled", which is much more clear. Thanks for pointing out the potential for confusion

sjahu:
> Does live poll happen on the root dashboard URL?

Thanks for catching that. I've changed it so that the live poll timer is only set up when the toggle button is present (and the button's hidden on the dashboard page).

web/assets/javascripts/application.js
```diff
@@ -50,27 +66,44 @@ function updateFuzzyTimes(locale) {
...
+    replacePage
+  ).complete(
+    scheduleLivePoll
+  )
+}
 
-    var $page = $data.filter('#page')
-    $('#page').replaceWith($page)
+function scheduleLivePoll() {
+  let ti = parseInt(localStorage.sidekiqTimeInterval) || 5000;
```
sjahu, line 93: Changed to `5000` to match the default value for polling on the dashboard page (both use `localStorage.sidekiqTimeInterval` when it's set).

mperham:
Thank you! You can run main directly if you want to smoke test the final result. I recently released a new gem version so it'll probably be 2-3 months before another release as I try not to release too frequently unless there is a high priority reason.

PR #4995, opened 2021-09-20
Warn if client concurrency cannot be determined

DylanReile:
Hi there!

We ran into a problem where our Puma processes had 20+ threads sharing the default pool of 5 Redis connections. This would occasionally throw confounding `ConnectionPool::Timeout` errors as the threads competed for a pool that was too shallow.

Unfortunately we were using `ENV['PUMA_MAX_THREADS']` instead of `ENV['RAILS_MAX_THREADS']` to set our thread count. Had we been using the latter, the pool would have been right-sized for us automatically.

I've added this warning as it would have helped us debug. Throwing an error—`Web concurrency cannot be determined. Please set RAILS_MAX_THREADS`—would have helped us catch it even before it hit production, but I don't want to break anything here.

### DIFF
--- a/lib/sidekiq/redis_connection.rb
+++ b/lib/sidekiq/redis_connection.rb
@@ -26,7 +26,17 @@ module Sidekiq
           5
         end
 
-        verify_sizing(size, Sidekiq.options[:concurrency]) if Sidekiq.server?
+        if Sidekiq.server?
+          verify_sizing(size, Sidekiq.options[:concurrency])
+        elsif !ENV["RAILS_MAX_THREADS"]
+          Sidekiq.logger.warn <<~WARN
+            Sidekiq uses the RAILS_MAX_THREADS envvar to set the size of its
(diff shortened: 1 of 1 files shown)


mperham:
I don't want to make this environment variable mandatory. IME it's unusual to require so many client connections and there are benefits (esp in environments where Redis connections are tightly limited) to keeping the pool size smaller than thread count.

DylanReile:
@mperham Agreed, I'm actually trying to convince my team to reduce to 5 threads. Just wondering, since this size is a max and the connections are lazily created, wouldn't they only be created if they really were needed? What triggers that?

mperham:
Right. 10 threads is the default, you might just stick to that. Sharing 5
connections across 10 threads seems very reasonable. The connection_pool
gem does the connection mgmt.

DylanReile:
@mperham Are you talking about the Puma default thread count?

PR #5001, opened 2021-09-27
Preserve behaviour of fetch_stats! returning stats.

mijoharas:
Preserve behaviour of previous `fetch_stats!` call.

While `fetch_stats!` might be better private, I think the api as-is is a bit cleaner if we make this change and preserve the previous behaviour.

I completely understand if you want to close instead though given your previous comment. Just thought I'd throw out the solution here so it's an easy option.

Follow on from https://github.com/mperham/sidekiq/pull/4936#issuecomment-924378724


mperham:
I'm 💯 with preserving behavior if it's possible. Thanks!

PR #5018, opened 2021-10-11
Fix rbenv usage example in systemd service

apauly:
Fixed rbenv usage example in the systemd service. There was an additional `exec` (probably copy/paste from "bundle exec")


apauly:
Good point, didn't think about that 👍

PR #5036, opened 2021-10-27
Refresh javascripts

mperham:
The Web UI's JS stack is 5 years old at this point. Modernize web/assets/javascripts/{dashboard,application}.js.

- [x] Remove jQuery?
- [x] Remove dropdown.js?
- [x] Upgrade D3/Rickshaw.
- [x] Upgrade timeago.

Need to verify desktop and mobile views.


mperham:
Rickshaw is reasonably current and locks the D3 version so I'm fine with leaving them as-is. timeago ain't broken so I'm not going to touch it either.

PR #5043, opened 2021-11-02
Use Redis.pipelined instead of Redis.multi when pushing jobs

engwan:
We don't need the commands here to be atomic so we use command pipelining instead
to save us some overhead from `MULTI`.

https://github.com/mperham/sidekiq/commit/ac69c7ee5e42357a0f87957d17d38e5512ed66bb references Sidekiq Pro so I don't know if this affects Sidekiq Pro's atomic batches. But maybe this could be overridden in Pro to use `multi`?

We have been running with this patch on GitLab.com for a week: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/72473

Effects on Redis CPU were inconclusive but I think it's a good change to make anyway?

### DIFF
--- a/lib/sidekiq/client.rb
+++ b/lib/sidekiq/client.rb
@@ -186,7 +186,7 @@ module Sidekiq
 
     def raw_push(payloads)
       @redis_pool.with do |conn|
-        conn.multi do
+        conn.pipelined do
           atomic_push(conn, payloads)
         end
       end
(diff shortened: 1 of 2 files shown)


mperham:
I can't believe I overlooked that. Yeah, that's an easy win. Pro/Ent don't change the semantics.

engwan:
Looks like I need to update some tests 🙂

PR #5044, opened 2021-11-02
Change scheduler to use Lua-based script

engwan:
This allows us to pop jobs atomically so that we remove redundant ZREMs
from multiple processes trying to process the same job.

When we (GitLab) upgraded to Sidekiq 6, [we saw that the number of ZREM operations on our Redis server increased significantly](https://gitlab.com/gitlab-com/gl-infra/scalability/-/issues/1179#note_695764650). This had a negative effect on Redis CPU.

We tracked this down to https://github.com/mperham/sidekiq/commit/9b75467b33759888753191413eddbc15c37a219e. That change fetches 100 jobs, then tries to remove them one at a time. This change makes sense for setups with a few number of processes since the maximum number of overlapping processes would be low.

In our case, we added logging and found out that [~44% of ZREM operations were redundant](https://gitlab.com/gitlab-com/gl-infra/scalability/-/issues/1179#note_704111452). We reverted to the Sidekiq 5 scheduler, but later on improved it further by introducing this patch.

It removed the redundant ZREM operations which restored Redis CPU back to Sidekiq 5 levels and [reduced enqueue latency](https://gitlab.com/gitlab-com/gl-infra/scalability/-/issues/1179#note_711058749) at the same time. 

In some cases, the current Sidekiq 6 scheduler may be faster. Like when there's a single process and there are a lot of scheduled jobs. But I think it is safe to assume that as instances process more scheduled jobs, there would also be more processes. So overall, I think this is still an improvement over the current scheduler.

### DIFF
--- a/lib/sidekiq/scheduled.rb
+++ b/lib/sidekiq/scheduled.rb
@@ -9,28 +9,47 @@ module Sidekiq
     SETS = %w[retry schedule]
 
     class Enq
+      LUA_ZPOPBYSCORE = <<~LUA
+        local key, now = KEYS[1], ARGV[1]
+        local jobs = redis.call("zrangebyscore", key, "-inf", now, "limit", 0, 1)
+        if jobs[1] then
+          redis.call("zrem", key, jobs[1])
+          return jobs[1]
+        end
+      LUA
+
+      def initialize
+        @lua_zpopbyscore_sha = nil
+      end
+
       def enqueue_jobs(now = Time.now.to_f.to_s, sorted_sets = SETS)
         # A job's "score" in Redis is the time at which it should be processed.
         # Just check Redis for the set of jobs with a timestamp before now.
         Sidekiq.redis do |conn|
           sorted_sets.each do |sorted_set|
-            # Get next items in the queue with scores (time to execute) <= now.
-            until (jobs = conn.zrangebyscore(sorted_set, "-inf", now, limit: [0, 100])).empty?
-              # We need to go through the list one at a time to reduce the risk of something
-              # going wrong between the time jobs are popped from the scheduled queue and when
-              # they are pushed onto a work queue and losing the jobs.
-              jobs.each do |job|
-                # Pop item off the queue and add it to the work queue. If the job can't be popped from
-                # the queue, it's because another process already popped it so we can move on to the
-                # next one.
-                if conn.zrem(sorted_set, job)
-                  Sidekiq::Client.push(Sidekiq.load_json(job))
-                  Sidekiq.logger.debug { "enqueued #{sorted_set}: #{job}" }
-                end
-              end
+            # Get next item in the queue with score (time to execute) <= now.
+            # We need to go through the list one at a time to reduce the risk of something
+            # going wrong between the time jobs are popped from the scheduled queue and when
+            # they are pushed onto a work queue and losing the jobs.
+            while (job = zpopbyscore(conn, keys: [sorted_set], argv: [now]))
+              Sidekiq::Client.push(Sidekiq.load_json(job))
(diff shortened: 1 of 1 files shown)


lib/sidekiq/scheduled.rb
```diff
@@ -9,29 +9,45 @@ module Scheduled
...
+      LUA_ZPOPBYSCORE = <<~EOS
+        local key, now = KEYS[1], ARGV[1]
+        local jobs = redis.call("zrangebyscore", key, "-inf", now, "limit", 0, 1)
+        if jobs[1] then
+          redis.call("zrem", key, jobs[1])
+          return jobs[1]
+        end
+      EOS
+
+      LUA_ZPOPBYSCORE_SHA = Digest::SHA1.hexdigest(LUA_ZPOPBYSCORE)
```
mperham, line 21: The expected hash is returned by `eval`. We should not be calculating the SHA directly but using the eval result instead.

engwan:
> I believe Gitlab has a Sidekiq Pro license. Have you tried the reliable scheduler in Pro? It uses Lua and does the job enqueue in Lua too.

We can't use it due to licensing issues. And even if we could, we need some client middleware to run on the actual enqueue.

lib/sidekiq/scheduled.rb
engwan: I think `eval` just returns the return value of the script and `SCRIPT LOAD` is the one that returns a SHA.

We could do that too, though the current code is just being optimistic that the code is already loaded because in most cases it will be. And even if we did `SCRIPT LOAD` at the start here and memoize it, there's no assurance that the script lives in Redis throughout this process' lifetime, so we'd need the error handling / rescue anyway.

I think this is a common pattern also seen in other open-source projects like Discourse: https://github.com/discourse/discourse/blob/1472e47aae5bfdfb6fd9abfe89beb186c751f514/lib/rate_limiter.rb#L198-L206

The Redis docs also says:

> The client library implementation can always optimistically send EVALSHA under the hood even when the client actually calls EVAL, in the hope the script was already seen by the server. If the NOSCRIPT error is returned EVAL will be used instead.

Though the `redis-rb` client does not do this so we have to do this manually.

mperham: Sorry, it's `script load` that returns the SHA.

mperham: Right, I still don't like it. If we calculate the SHA differently from what Redis calculates (like a stripped linefeed), it will always be wrong and we will not error. Please use script load to be safe.

engwan: Sure. I'll try to get back to this tomorrow.

rainerborene: @mperham @engwan What do you think about a new module `Sidekiq::Scripting` for dealing with lua stuff just like what Sidekiq Pro does? Just an idea for refactoring.

engwan: If Sidekiq Pro already has a pattern for loading a script and running it using EVALSHA, it would be great to extract that.

I was also wondering if EVALSHA is worth it here. We could also just call EVAL every time since Redis also does some optimizations already. The script is short anyway.

But if Pro does that too and we could extract it into a common module, then it's a no-brainer to just use it.

engwan: I have updated the current PR to use `SCRIPT LOAD`. Let me know if you think we should extract this into a separate module as suggested above.

mperham:
I'm fine with the code as is: compact and self-contained. Lua was originally Pro-only as it required newer versions of Redis but Lua has been in Redis for many years now; I think Sidekiq can start to use it where useful.

Once we have multiple Lua call points, we'll extract a utility method at that point in time.

PR #5055, opened 2021-11-10
Replace worker generator with job generator

marcelolx:
As announced here https://www.mikeperham.com/2021/11/07/whats-new-in-sidekiq-6.3/ what sidekiq encourages now is using the term Job instead of Worker, that said I think we could also remove the worker generator and replace it with a job generator. 

Or if desired keep both and remove the worker generator in the future.


mperham:
The issue is that, by convention, ActiveJob "owns" the `app/jobs` directory. Sidekiq owns `app/workers`. I don't want native Sidekiq jobs and ActiveJobs to reside in the same directory. Does that make sense? Thoughts?

mperham:
What about `app/sidekiq`?

marcelolx:
We put everything under `app/jobs` even all our jobs being sidekiq workers, but it is a fair point, I'm ok with `app/sidekiq`

marcelolx:
Changed to generate under `app/sidekiq` 🙌

rainerborene:
> The issue is that, by convention, ActiveJob "owns" the app/jobs directory.

I disagree. That would be the same thing saying that `app/models` is owned by ActiveRecord, which is not true. Discourse, for example, uses `app/jobs` folder for `sidekiq` jobs. I don't think it's common to see both `ActiveJob` and `Sidekiq::Worker` being used at the same thing or what's the point of ActiveJob abstraction anyway? But I could be wrong here.

TonsOfFun:
I have actually had some of my clients mixing `ActiveJob` with Sidekiq set as the queue adapter and `Sidekiq::Worker` classes in their `app/jobs`. 

I think an important distinction here should be the difference between using Sidekiq as an ActiveJobs queue adapter with `Rails.application.config.active_job.queue_adapter = :sidekiq` vs using Sidekiq standalone with classes that `include Sidekiq::Job`. I personally prefer to use Sidekiq standalone in `app/jobs` since the ActiveJob abstract adds overhead with little developer benefit. 

When Sidekiq is used as an ActiveJob queue adapter I think it should remain in `app/jobs` as that is the expected behavior for `ActiveJob::QueueAdapters`. This is where it is similar to `ActiveRecord::ConnectionAdapters` and active record models being in `app/models`. 

That said there is something else to consider, as @rainerborene mentioned, is many developers do define classes in `app/models` that don't inherit from ActiveRecord.

dilraj-vidyard:
Is it possible to `bundle exec rails g sidekiq:worker Users::HardWorker` but specify the directory?

mperham:
Looks pretty hardcoded to me. I don't know if generators allow you to override that path somehow. Nothing a simple `mv` can't fix.

PR #5056, opened 2021-11-10
Continue scheduling jobs when backlog jobs exist

engwan:
This is another attempt to further improve scheduled jobs' enqueue latency. This does 2 changes:

1. It terminates the scheduler gracefully inside the `enqueue_jobs` loop and does not wait for the current poll to complete.
 
   This is important when the scheduled jobs backlog is high and it takes a long time to go through them. We could go over the graceful timeout and lose jobs if we're in the middle of enqueuing them. I think this is a rare case but I also needed this for the next change below.

1. This makes the scheduler keep on scheduling as long as there are jobs that are due instead of waiting for the next poll.

   This is useful when there are a lot of scheduled jobs where each poll could take a few seconds. By the time the poller is done, there are usually more jobs that are due. With the previous code, these jobs would only be processed on the next poll which would be ~5 seconds away (by default).

   Also, the longer each poll takes, the actual poll interval becomes larger too. That's because the sleep starts after the poll is done. So jobs get more delayed as backlog piles up.

   This does increase the chance that polls from different processes overlap because there's still remaining work. But I think that's a good thing because with the Lua scheduler, they actually help each other efficiently now.

### DIFF
--- a/test/test_scheduled.rb
+++ b/test/test_scheduled.rb
@@ -79,6 +79,30 @@ describe Sidekiq::Scheduled do
       end
     end
 
+    it 'should not enqueue jobs when terminate has been called' do
+      created_time  = Time.new(2013, 2, 3)
+      enqueued_time = Time.new(2013, 2, 4)
+
+      Time.stub(:now, created_time) do
+        @retry.schedule (enqueued_time - 60).to_f, @error_1.merge!('created_at' => created_time.to_f)
+        @scheduled.schedule (enqueued_time - 60).to_f, @future_1.merge!('created_at' => created_time.to_f)
+      end
+
+      Time.stub(:now, enqueued_time) do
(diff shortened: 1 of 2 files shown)


test/test_scheduled.rb
```diff
@@ -79,6 +79,30 @@ def call(worker_class, job, queue, r)
       end
     end
 
+    it 'should not enqueue jobs when terminate has been called' do
```
engwan, line 82: I just added this test for the terminate case because I was struggling to find a clean way to write other tests that stub and test the behavior inside the `enqueue_jobs` loop.

I think that would be testing implementation details anyway so maybe this is fine?

mperham:
LGTM

engwan:
Thanks for the quick review @mperham! 🙇

Issue #5058, opened 2021-11-12
Discussion: Lazy enumerators being passed to `.perform_bulk`


kellysutton:
This issue is opening the discussion on the expectations of Sidekiq and how it handles lazy enumerators. 

Here's a toy example that we can use to discuss this issue:

```ruby
class Worker
  include Sidekiq::Worker

  def perform(id)
  end
end

# Then, some time later...
args = (1...1001).lazy.map { |id| [id] }
Worker.perform_bulk(args)
```

Interesting, executing the last line in that code snippet enqueues **zero** Sidekiq jobs. Why? Because of the implementation of `.perform_bulk`. Let's see how that looks:

```ruby
def perform_bulk(args, batch_size: 1_000)
  args.each_slice(batch_size).flat_map do |slice|
    Sidekiq::Client.push_bulk(@opts.merge("class" => @klass, "args" => slice))
  end
end
```

Because the chain of `.each_slice.flat_map` retains the laziness of the enumerator, the block that calls `.push_bulk` is never executed. The following change to the original _does_ cause the code to be executed:

```ruby
Worker.perform_bulk(args).to_a

# OR

Worker.perform_bulk(args.to_a)
```

That little `.to_a` (or a similar method that evaluates the contents of the enumerator) causes our important block that calls `Sidekiq::Client.push_bulk` to be called. 

Here's where I'd like to open the discussion:

- Are we okay with this? Does this go into the bucket of "Well yeah, lazy enumerators have some sharp edges and this is one of them"? OR...
- Should we make `.perform_bulk` a bit "smarter" when it comes to dealing with lazy enumerators. The whole purpose of `.perform_bulk` is to inflict a side effect (i.e. call `Sidekiq::Client.push_bulk`). We could say that it should do that in all cases, regardless of what type of enumerator (lazy or not) gets passed to it.

Happy to open a PR if we go with the second option!

---

I should also mention that myself and others were the ones that proposed `.perform_bulk` in Sidekiq here. I realize the irony that we're reporting a bug in something we contributed last week :)

mperham:
I had thought about that too but I don't want to prematurely add functionality no one will use. It would be nice to support streaming `batch_size` jobs at a time, but only if someone is actually going to use it.

kellysutton:
Cool. I'll create a PR that calls `.force` on the enumerator if it's lazy, just so the behavior is explicit and straight-forward for now, i.e. if you call `.perform_bulk`, Redis will get called N times.

[...]

PR #5060, opened 2021-11-13
Make `.perform_bulk` immediately evaluate lazy enumerators.

kellysutton:
Because `.perform_bulk`'s job is to enqueue Sidekiq jobs, it can bit a little surprising to not have jobs enqueued when it is called with a lazy enumerator until the returned JID enumerable has `.to_a` or `.force` called on it.

This PR changes the behavior of `.perform_bulk` to evaluate a lazy enumerable to ensure jobs are enqueued as soon as `.perform_bulk` is called and not some indeterminate time in the future (or never).

More background here: https://github.com/mperham/sidekiq/issues/5058

### DIFF
--- a/lib/sidekiq/worker.rb
+++ b/lib/sidekiq/worker.rb
@@ -192,10 +192,11 @@ module Sidekiq
       end
 
       def perform_bulk(args, batch_size: 1_000)
-        args = args.force if args.is_a?(Enumerator::Lazy)
-        args.each_slice(batch_size).flat_map do |slice|
+        result = args.each_slice(batch_size).flat_map do |slice|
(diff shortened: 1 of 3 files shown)


mperham:
Rather than force, can we stream the jobs in batch_size chunks?

kellysutton:
Yep. I'll get a few commits put together today that does that.

kellysutton:
@mperham Took another crack at implementing this to make sure we lazily load values from the enumerator `batch_size` values at a time. We decided to not return a lazy enumerator, since the JIDs will need to be in memory somewhere regardless.

Let us know what you think!

mperham:
Amazing. Ruby magic. 🪄

PR #5061, opened 2021-11-15
Support `perform_inline`

hasan-ally:
Calling `WorkerClass.new.perform(*args)` will skip the client and server middlewares

This PR adds support for `perform_inline` method e.g., `WorkerClass.perform_inline(*args)` would call the job's perform method after passing through available client and server middlewares

### DIFF
--- a/lib/sidekiq/client.rb
+++ b/lib/sidekiq/client.rb
@@ -67,13 +67,17 @@ module Sidekiq
     #   push('queue' => 'my_queue', 'class' => MyWorker, 'args' => ['foo', 1, :bat => 'bar'])
     #
     def push(item)
-      normed = normalize_item(item)
-      payload = process_single(item["class"], normed)
+      payload = payload_for_push(item)
 
       if payload
         raw_push([payload])
         payload["jid"]
       end
+    end
+
+    def payload_for_push(item)
+      normed = normalize_item(item)
+      process_single(item["class"], normed)
     end
 
     ##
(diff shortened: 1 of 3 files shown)


lib/sidekiq/worker.rb
```diff
@@ -241,6 +257,11 @@ def perform_async(*args)
         client_push("class" => self, "args" => args)
       end
 
+      # Inline execution of job's perform method after passing through Sidekiq.client_middleware and Sidekiq.server_middleware
+      def perform_inline(*args)
+        set({}).perform_inline(*args)
```
mperham, line 262: Is there a way to get rid of this unnecessary `set`?

lib/sidekiq/client.rb
```diff
@@ -67,15 +67,19 @@ def initialize(redis_pool = nil)
...
-      payload = process_single(item["class"], normed)
+      payload = payload_for_push(item)
 
       if payload
         raw_push([payload])
         payload["jid"]
       end
     end
 
+    def payload_for_push(item)
```
mperham, line 78: This creates a new public API. Can't do that.

hasan-ally: One option I can think of is to move this method to `worker.rb` but that may not be maintainable as logic for private method `normalize_item` would be in 2 places.
Any suggestions on how to proceed?

lib/sidekiq/worker.rb
hasan-ally: Would moving the logic to `Sidekiq::Client` be an acceptable approach?

mperham:
I'll look into this tomorrow. I think the current code structure makes it difficult to implement as cleanly as I'd like so I need to ponder a bit.

mperham:
Thanks for your help here. I will update the changelog to credit you.

PR #5062, opened 2021-11-16
Make `__FILE__` and `__dir__` behave as expected in ERB configs

sambostock:
This enables the use of `__FILE__` and `__dir__` in ERB config files by setting [`ERB#filename`](https://ruby-doc.org/stdlib-3.0.2/libdoc/erb/rdoc/ERB.html#filename) before computing its `results`. This allows `__FILE__` to provide a useful value, and `__dir__` works automatically if `__FILE__` is set correctly.

### Use Case

While in production it may be useful to have granular workers, in development it may be convenient (and less expensive) to have a worker capable of processing all queues, so as not to require starting up all workers individually.

Furthermore, it may be desirable to dynamically read in queue names from the other configs, so as not to have to worry about "drift" in the development config file (e.g. if a new queue is added).

Having access to `__FILE__` and `__dir__` allows such code to be written without worrying about  exact filenames or locations:

```yml
---
:queues: <%=
  (Dir.glob(File.join(__dir__, 'sidekiq*.yml')) - [__FILE__]).flat_map do |config_path|
    YAML.load_file(config_path).fetch(:queues, 'default')
  end.uniq.to_json
%>
```

### DIFF
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -380,7 +380,8 @@ module Sidekiq
     end
 
     def parse_config(path)
-      opts = YAML.load(ERB.new(File.read(path)).result) || {}
+      erb = ERB.new(File.read(path))
+      opts = YAML.load(erb.result) || {}
 
       if opts.respond_to? :deep_symbolize_keys!
         opts.deep_symbolize_keys!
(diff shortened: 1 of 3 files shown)


mperham:
Frankly that YAML sample seems byzantine but the code change is a definite improvement. LGTM

Issue #5070, opened 2021-11-24
[Proposal] Primitive Checks for Job Args


kellysutton:
The goal of this issue is to propose what runtime arg-checking might look like for Sidekiq. If it's interesting, I can put together a PR implementing things.

Essentially, we'd encode Sidekiq's [first best practice of making job parameters "small and simple."](https://github.com/mperham/sidekiq/wiki/Best-Practices#1-make-your-job-parameters-small-and-simple)

Here's the sketch:

- On every set of args, run the check snippet (below)
- This would be configurable per-environment, so that we can fail in dev/test but pass-through and warn in production.
- We'd need to make sure this wouldn't add undue performance overhead.
- By making it configurable, this also opens the option of changing the default in a future major version of Sidekiq. (It would be a breaking change.)
- (Optional) Provide a callback when a complex argument is received.

**Snippet:**

```ruby
# any time we get `args`
klasses = [String, Numeric, TrueClass, FalseClass, NilClass] # would be stored in a constant somewhere
args.each do |arg|
  if klasses.none? { |klass| arg.is_a?(klass) }
    if Sidekiq.raise_on_complex_arguments
      raise Sidekiq::ComplexArgumentError
    else
      Sidekiq.logger.warn "Receive a complex argument"
    end
  end
end
```

That's the sketch. We've been bitten by this a few times as we've scaled where the documentation and knowledge of it is not evenly distributed, but Sidekiq usage is.

sobrinho:
@kellysutton would you create a PR?

kellysutton:
Yep, I'll try to put something together this week. Thanks!

Issue #2870, opened 2016-03-06
Should Sidekiq Error when `perform_async` is called with poorly serializable arguments?


EndangeredMassa:
Many hours have been lost by those I know (and many more beyond, I'm sure) to the quirk of sidekiq's implementation over redis that causes arguments passed to `perform_async` to be serializable to json and back. This causes symbols and some other values to be mutated between scheduling a worker and it executing, from the perspective of the caller.

Should Sidekiq force the arguments to be perfectly serializable and deserializable to the same value? Or at least, could there be an option to enable this behavior?

This could be accomplished by adding a guard to `perform_async` such as:

```ruby
def perform_async(*args)
  if args != JSON.load(JSON.dump(args))
    # TODO: craft a better error message
    raise StandardError.new("BAD ARGS")
  end
  client_push('class' => self, 'args' => args)
end
```

mperham:
It would be a good idea to run that in non-production.

EndangeredMassa:
I'm perfectly happy to have this default to off.

Personally, however, I would actually like to run this in production. I'd rather have a more explicit error from this check than a more vague error (like a nil reference somewhere) down in the execution of the worker.

Would you consider a PR that adds this under a sidekiq config value that defaults to false?

mperham:
I would like it non-production because this is the type of error that should be caught during testing.  It has a significant performance penalty so I'd prefer it not enabled in production.

It also has the potential to break existing user code so I don't want to break production.  No, I don't want to make it user-configurable.

EndangeredMassa:
Ah, yeah. It could be a performance problem. Good call.

So, always on when environment is not `production`?

mperham:
Right.

mperham:
I looked into this but ActiveJob breaks this check so I can't enable it.

PR #5071, opened 2021-11-29
Implement strict argument checking

kellysutton:
This PR implements a runtime type check for arguments as a follow-up from the discussion here: https://github.com/mperham/sidekiq/issues/5070. The idea is to provide some safeguards in non-production environments for folks about to do something dangerous (i.e. use arguments in a job that cannot (de)serialize reliably to/from JSON).

Open Questions:

- Is `#validate` or `#normalize_item` the right place for this logic?
- Are there other types we'd like to include in the allow list?
- This check is relatively strict as written (no symbols, no hashes). Do we want to instead take the approach explored in #2870?

Notes:

- It looks like this idea has been considered before and has been commented out for some time: https://github.com/mperham/sidekiq/issues/2870 and ...
- https://github.com/mperham/sidekiq/blame/848df46c67da926587cd76ab7d78939b4d79714b/lib/sidekiq/client.rb#L222

### DIFF
--- a/lib/sidekiq/job_util.rb
+++ b/lib/sidekiq/job_util.rb
@@ -6,8 +6,6 @@ module Sidekiq
     # These functions encapsulate various job utilities.
     # They must be simple and free from side effects.
 
-    JSON_SERIALIZABLE_TYPES = [String, Numeric, TrueClass, FalseClass, NilClass].freeze
-
     def validate(item)
       raise(ArgumentError, "Job must be a Hash with 'class' and 'args' keys: `#{item}`") unless item.is_a?(Hash) && item.key?("class") && item.key?("args")
       raise(ArgumentError, "Job args must be an Array: `#{item}`") unless item["args"].is_a?(Array)
@@ -16,17 +14,12 @@ module Sidekiq
       raise(ArgumentError, "Job tags must be an Array: `#{item}`") if item["tags"] && !item["tags"].is_a?(Array)
 
       if Sidekiq.options[:raise_on_complex_arguments]
-        item["args"].each do |arg|
-          if JSON_SERIALIZABLE_TYPES.none? { |klass| arg.is_a?(klass)}
-            raise(ArgumentError, "Job argument must be a type safe to serialize to JSON, i.e. String, Numeric, TrueClass, FalseClass, NilClass")
-          end
-        end
+        raise(ArgumentError, "Arguments must be native JSON types, see https://github.com/mperham/sidekiq/wiki/Best-Practices") unless JSON.load(JSON.dump(item['args'])) == item['args']
       end
     end
 
     def normalize_item(item)
       validate(item)
-      # raise(ArgumentError, "Arguments must be native JSON types, see https://github.com/mperham/sidekiq/wiki/Best-Practices") unless JSON.load(JSON.dump(item['args'])) == item['args']
 
       # merge in the default sidekiq_options for the item's class and/or wrapped element
       # this allows ActiveJobs to control sidekiq_options too.
--- a/lib/sidekiq/job_util.rb
+++ b/lib/sidekiq/job_util.rb
@@ -20,7 +20,7 @@ module Sidekiq
           To disable this error, remove `Sidekiq.strict_mode!` from your initializer.
         EOM
         raise(ArgumentError, msg)
-      elsif Sidekiq.options[:environment] == "development" && !json_safe?(item)
+      elsif !Sidekiq.options[:disable_complex_argument_warning] && Sidekiq.options[:environment] == "development" && !json_safe?(item)
         Sidekiq.logger.warn <<~EOM
           Job arguments do not serialize to JSON safely. This will raise an error in Sidekiq 7.0.
 
--- a/lib/sidekiq/job_util.rb
+++ b/lib/sidekiq/job_util.rb
@@ -6,6 +6,8 @@ module Sidekiq
     # These functions encapsulate various job utilities.
     # They must be simple and free from side effects.
 
+    JSON_SERIALIZABLE_TYPES = [String, Numeric, TrueClass, FalseClass, NilClass].freeze
+
     def validate(item)
       raise(ArgumentError, "Job must be a Hash with 'class' and 'args' keys: `#{item}`") unless item.is_a?(Hash) && item.key?("class") && item.key?("args")
       raise(ArgumentError, "Job args must be an Array: `#{item}`") unless item["args"].is_a?(Array)
@@ -14,9 +16,8 @@ module Sidekiq
       raise(ArgumentError, "Job tags must be an Array: `#{item}`") if item["tags"] && !item["tags"].is_a?(Array)
 
       if Sidekiq.options[:raise_on_complex_arguments]
-        klasses = [String, Numeric, TrueClass, FalseClass, NilClass]
         item["args"].each do |arg|
-          if klasses.none? { |klass| arg.is_a?(klass)}
+          if JSON_SERIALIZABLE_TYPES.none? { |klass| arg.is_a?(klass)}
             raise ArgumentError, 'Out of luck!'
           end
         end
--- a/lib/sidekiq/job_util.rb
+++ b/lib/sidekiq/job_util.rb
@@ -13,8 +13,8 @@ module Sidekiq
       raise(ArgumentError, "Job 'at' must be a Numeric timestamp: `#{item}`") if item.key?("at") && !item["at"].is_a?(Numeric)
       raise(ArgumentError, "Job tags must be an Array: `#{item}`") if item["tags"] && !item["tags"].is_a?(Array)
 
-      if Sidekiq.options[:raise_on_complex_arguments]
-        raise(ArgumentError, "Arguments must be native JSON types, see https://github.com/mperham/sidekiq/wiki/Best-Practices") unless job_is_json_safe(item)
+      if Sidekiq.options[:raise_on_complex_arguments] && !job_is_json_safe(item)
+        raise(ArgumentError, "Arguments must be native JSON types, see https://github.com/mperham/sidekiq/wiki/Best-Practices. To disable this error, remove Sidekiq.strict_mode! from your initializer.")
       elsif Sidekiq.options[:environment] == "development" && !job_is_json_safe(item)
         Sidekiq.logger.warn <<~EOM
           Job arguments do not serialize to JSON safely. This will raise an error in Sidekiq 7.0.
--- a/lib/sidekiq/job_util.rb
+++ b/lib/sidekiq/job_util.rb
@@ -18,7 +18,7 @@ module Sidekiq
       if Sidekiq.options[:raise_on_complex_arguments]
         item["args"].each do |arg|
           if JSON_SERIALIZABLE_TYPES.none? { |klass| arg.is_a?(klass)}
-            raise ArgumentError, 'Out of luck!'
+            raise(ArgumentError, "Job argument must be a type safe to serialize to JSON, i.e. String, Numeric, TrueClass, FalseClass, NilClass")
           end
         end
       end
--- a/lib/sidekiq/job_util.rb
+++ b/lib/sidekiq/job_util.rb
@@ -24,7 +24,7 @@ module Sidekiq
         Sidekiq.logger.warn <<~EOM
           Job arguments do not serialize to JSON safely. This will raise an error in Sidekiq 7.0.
 
-          see https://github.com/mperham/sidekiq/wiki/Best-Practices or raise the error today
+          See https://github.com/mperham/sidekiq/wiki/Best-Practices or raise the error today
           by calling `Sidekiq.strict_mode!` during Sidekiq initialization.
         EOM
       end
--- a/lib/sidekiq/job_util.rb
+++ b/lib/sidekiq/job_util.rb
@@ -12,6 +12,15 @@ module Sidekiq
       raise(ArgumentError, "Job class must be either a Class or String representation of the class name: `#{item}`") unless item["class"].is_a?(Class) || item["class"].is_a?(String)
       raise(ArgumentError, "Job 'at' must be a Numeric timestamp: `#{item}`") if item.key?("at") && !item["at"].is_a?(Numeric)
       raise(ArgumentError, "Job tags must be an Array: `#{item}`") if item["tags"] && !item["tags"].is_a?(Array)
+
+      if Sidekiq.options[:raise_on_complex_arguments]
+        klasses = [String, Numeric, TrueClass, FalseClass, NilClass]
+        item["args"].each do |arg|
+          if klasses.none? { |klass| arg.is_a?(klass)}
+            raise ArgumentError, 'Out of luck!'
+          end
+        end
+      end
     end
 
     def normalize_item(item)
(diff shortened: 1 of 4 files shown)


kellysutton:
Sounds good. We've updated the PR to address the feedback!

mperham:
I'm worried about the equality check. Does that correctly handle deep elements, e.g. an Array of Symbols?

kellysutton:
Let's find out. We'll write a few more interesting test cases.

kellysutton:
Test cases improved. It looks like the JSON.parse/dump approach works to prevent things like symbols either as values or keys be serialized, while letting JSON-friendly hashes/arrays pass through.

mperham:
Ok, now:

1. when should it be enabled by default? `if env == 'development'`?
2. should we provide an API for it? `Sidekiq.strict_mode(true/false)`?
3. should the raised error note how to enable/disable it?
4. where do we document this, if at all, in the wiki?

kellysutton:
> 1. when should it be enabled by default? if env == 'development'?

My thinking: It should be configurable in 6.x but maybe become the default in `env == 'development'` starting in 7.x. Maybe with a deprecation warning closer to the release of 7.x to smooth the upgrade path. How does that sound? Too complicated?

> 2. should we provide an API for it? Sidekiq.strict_mode(true/false)?

Seems like it could be a nice way of simplifying the API, but I can go either way. My assumption here is that calling `Sidekiq.strict_mode!` would set a value in the `options` hash. Happy to get that whipped up for whichever way you think is best.

> 3. should the raised error note how to enable/disable it?

Yep. I'll push a commit to include a note about that once a decision is made for (2).

> 4. where do we document this, if at all, in the wiki?

For sure in the Best Practices section about simple arguments. If we go with the 6.x/7.x split in (1), the Best Practice can read something like:

> If you want this best practice enforced, you can use `Sidekiq.strict_mode!` in the environments where you'd like to raise an error.

If we make it the default in 7.x, that message will need to change a bit.

mperham:
Maybe a WARN level log in 6.4+ and raise in 7+?

We're really building a hyper-specialized linter here. Is this better off as a Rubocop cop? What do you think about warning about larger argument sizes, e.g. greater than 16kb?

kellysutton:
> Maybe a WARN level log in 6.4+ and raise in 7+?

Sounds good. I'll make it so.

> We're really building a hyper-specialized linter here. Is this better off as a Rubocop cop? What do you think about warning about larger argument sizes, e.g. greater than 16kb?

The check we're doing is a bit more of a runtime check (whereas I think of linters as static checks), but I think the ideal place for this type of check is statically. I believe Sorbet/RBS definitions would be the best place to enforce this type of thing, since they could do so without actually running the code.

> What do you think about warning about larger argument sizes, e.g. greater than 16kb?

Absolutely. I was going to propose that in a future PR :) Happy to roll it into this one if that makes more sense.

For now, I'll get things tidied up based on the conversation so far.

kellysutton:
Alright, ready for another review at your leisure. Let me know if you'd like me to squish the commits on this PR, since we've accumulated quite a few.

lib/sidekiq/job_util.rb
```diff
@@ -42,5 +57,11 @@ def normalized_hash(item_class)
         Sidekiq.default_worker_options
       end
     end
+
+    private
+
+    def job_is_json_safe(item)
```
mperham, line 63: I believe a better name is `json_safe?`.

test/test_client.rb
```diff
@@ -122,6 +122,83 @@ class QueuedWorker
...
+        after do
+          Sidekiq.strict_mode!(false)
+        end
+
+        it 'raises an error' do
+          assert_raises ArgumentError do
+            InterestingWorker.perform_async(
+              :symbol,
+              Date.new(2021, 1, 1),
+              { some: 'hash', 'with' => 'different_keys' },
```
mperham, line 173: I'd prefer to see a few different tests to verify we catch symbols in Hashes, symbols in Arrays and Dates. As is, this test fails but maybe because it catches one of those cases but not the others. Test one simple case at a time.

mperham (commented):
Looking pretty good. Do we need a way for the user to disable the check completely (to restore current behavior?). Consider if they have an app which breaks the best practices but their code handles it.

kellysutton:
> Do we need a way for the user to disable the check completely (to restore current behavior?). Consider if they have an app which breaks the best practices but their code handles it.

Added a flag in d26808b to handle this case. Not sure what the best case to test this would be, since the only side effect here is logging. We'll need to document this if it gets merged.

We addressed the other bits of feedback in other commits!

gingerlime:
Any tips for migrating our rails app? I guess we can use `JSON.parse(JSON.parse(param))` for hashes that might contain symbols... but I guess that's not a proper solution... curious about some strategies for tackling this... side-note/rant: this is where I wonder again why ruby invented symbols ;-)

kellysutton:
@gingerlime Migrations will probably take a lot of "lead bullets," i.e. taking a look at each job that is/might be using symbols in your application. Hopefully the ability to turn this on/off in different environments helps!

gingerlime:
@kellysutton thank you, yes the flag can be very helpful. I'm trying to set it however in my initializer and I get an error

`/app/config/initializers/sidekiq.rb:48:in `<top (required)>': undefined method `strict_mode!' for Sidekiq:Module (NoMethodError)`

I checked and I'm using sidekiq 6.4 (it shows me the warnings)...

I tried setting it both with `Sidekiq.strict_mode!` and `Sidekiq.strict_mode!(true)`

gingerlime:
Ah, I saw that it was changed to `strict_args!` ... :)

gingerlime:
Not sure this is the right place, but I'm feeling a bit torn about this new feature (impending in Sidekiq 7).

I completely understand that we want to protect people from footguns sending stuff that gets serialized differently to JSON and therefore get altered in the workers. Not to mention the option to turn this on/off per environment right now.

On the other hand, going through our code and doing things "properly" by converting symbols to strings whenever they *might* get sent to Sidekiq feels strange and not-very ruby-friendly? symbols are such a natural part of writing ruby, that it feels strange to convert them to strings, particularly in classes or methods that don't directly interact with Sidekiq (but are being used by other code that might). The alternative, to wrap things with `JSON.parse(JSON.dump(...))`, while explicit, also feels like a hack ... 

Is there some middle ground or a better way to serialize data that maintains some of the ruby properties? we've been using Sidekiq since version 3 or so(?) and it was fine... Our jobs had to assume symbols got converted to strings and deal with it. It wasn't that bad, even though it's not purely correct... 

I don't want this to sound like a rant, and I'm sorry I'm not familiar with the Sidekiq internals to offer something concrete, but I'm just trying to describe my experience trying to convert our code to fit the RightWay of doing things...

bnorton:
I was not aware of this coming change but honestly feels like a sort of strange default to enforce -- In the many applications that I am involved with we pass arguments that are not "equal" via json dump + parse but are "equal" in all other ways.

As a path forward are we expecting 1000's of sidekiq users to change essentially all of the places that `perform_async`?

An couple examples:

```ruby
 # the id field is a BSON::ObjectId
CampaignCreated.perform_async(self.d)
```

```ruby
 # keys are symbols
CampaignUpdated.perform_async(self.id, field: changed_attr, value: changed_value)
```

```ruby
# the the model defines the `as_json` function to return the id and a few other key properties to handle the deletion now
# that the model cannot be looked up
CampaignDeleted.perform_async(self)
```

gingerlime:
Perhaps we're panicking for no reason and the option to disable it will stay available? but the [Changelog](https://github.com/mperham/sidekiq/blob/8ee13c703d569beba5b1c18484a991808b1e993a/Changes.md?plain=1#L17-L19) can be a bit confusing for me...

>  Sidekiq will now log a warning if JSON-unsafe arguments are passed to `perform_async`.
  Add `Sidekiq.strict_args!(false)` to your initializer to disable this warning.
  This warning will switch to an exception in Sidekiq 7.0.

Would `strict_args!(false)` continue to be available in Sidekiq 7.0 ? Or would the switch from warning to an exception also mean that we won't be able to turn this off?

mperham:
It will remain available for backwards compatibility reasons.

gingerlime:
Thanks @mperham. That's very reassuring. 👍

NemyaNation:
Am I missing something or am I in the wrong section:

warning in logs: 

```
app[web] info redis.pipelined do
app[web] info   redis.get("key")
app[web] info end
app[web] info
app[web] info should be replaced by
app[web] info
app[web] info redis.pipelined do |pipeline|
app[web] info   pipeline.get("key")
app[web] info end
app[web] info
app[web] info (called from /app/vendor/bundle/ruby/2.7.0/gems/sidekiq-6.4.0/lib/sidekiq/api.rb:57:in `block in fetch_stats_fast!'}
Feb 4 00:09:38 my-app app[web] info Pipelining commands on a Redis instance is deprecated and will be removed in Redis 5.0.0.
```

`initializers/sidekiq.rb`

```
	config.redis = {
		url: ENV["REDISCLOUD_URL"]
	}
end

	config.redis = {
		url: ENV["REDISCLOUD_URL"],
		size: 2
	}
end

Sidekiq.strict_args!(false)

```

Still clogging up my logs, and this is just from live polling on the dashboard.

mperham:
@NemyaNation Why are you commenting on this issue?

NemyaNation:
@mperham Sorry just realised I commented on the wrong issue.

Shall I create a new issue or is there a simple fix other than downgrading a level.

tenpaiyomi:
I'm kind of with @bnorton and @gingerlime on this.  I can understand the necessity of wanting to ensure users don't send along a whole Ruby object, but the strict enforcement on hashes feels kind of hamfisted.  As the change exists currently, I have to go through and change hundreds of lines of code, because a `perform` function like this:

`def perform(user_id, user_event, properties = {})`

 that is called like this
 
`SomeWorker.perform_async(some_id, some_string, { foo: 'string', bar: 'string' })`

is required to be set up like this

`SomeWorker.perform_async(some_id, some_string, { 'foo' => 'string', 'bar' => 'string' })`

Even though the output of the hash in both cases with `JSON.dump` is identical,

```
irb(main):001:0> JSON.dump({foo: 'string', bar: 'string'})
=> "{\"foo\":\"string\",\"bar\":\"string\"}"
irb(main):002:0> JSON.dump({'foo' => 'string', 'bar' => 'string'})
=> "{\"foo\":\"string\",\"bar\":\"string\"}"
```

But the latter format is no longer Ruby idiomatic and is largely discouraged over utilizing the `symbol: value` format.

It feels like the `json_safe?` method is too simplistic.  I can understand the check with regards to some things (Date objects, for instance), but other things should be treated a bit more logically, IMO.  Some whitelisted basic classes, and adding a recursive check, seems like the more logical approach that doesn't force potentially old style conventions.

tenpaiyomi:
Right, I'm aware of the initializer flag that can be set, but that still leaves the fact that the argument checking will probably be somewhat confusing to individuals who are passing valid JSON serializable arguments, that the wiki says are acceptable, and getting errors about them.  Likewise, that is negating the protection of potentially passing _actual_ invalid objects as well.

mperham:
If the wiki says they are acceptable, let's fix the wiki. Symbols are not valid JSON.

tenpaiyomi:
Nor are the hash rockets either.  The thing is in this case is not what is valid in JSON, the problem is what is being _expected_ to consider the object valid.

Not valid JSON, not accepted:
`{foo: 'bar'}`

Not valid JSON, accepted:
`{'foo' => 'bar'}`

Neither of these are valid JSON objects.  If we were going on _only_ accepting valid JSON objects, then only a variant such as this should pass

`"{\"foo\": \"bar\"}"`

```
irb(main):020:0> args = "{'foo': 'bar'}"
=> "{'foo': 'bar'}"
irb(main):021:0> JSON.parse(JSON.dump(args)) == args
=> true
irb(main):022:0> args = {'foo' => 'bar'}
=> {"foo"=>"bar"}
irb(main):023:0> JSON.parse(JSON.dump(args)) == args
=> true
irb(main):024:0> args = {'foo':  'bar'}
=> {:foo=>"bar"}
irb(main):025:0> JSON.parse(JSON.dump(args)) == args
=> false
```

mperham:
Let's be clear here:

`perform_async` is not a normal Ruby method call. It is an helper for constructing a background job which maps onto a Ruby method call "somewhere else". The wire format for this "remote procedure call" is JSON. We are trying to make that mapping between wire format and the Ruby method call mapping safer by catching unsafe conversions to/from Ruby/JSON. `{'foo' => 'bar'}` is perfectly legal because it maps safely to JSON.

kellysutton:
The symbolized names have bitten us a few times before, which is why we introduced this check. While hashes with keys for symbols serializes safely to JSON, it does not deserialize to the same value. Here's a code snippet to illustrate this:

```ruby
class MyWorker
  include Sidekiq::Worker

  def perform(hash)
    id = hash[:id] # will never exist, but hash['id'] will!
  end
end

MyWorker.perform_async({ id: 123 }) #=> becomes the JSON "{ \"id\": 123 }"
```

tenpaiyomi:
@kellysutton yeah, I can certainly understand that scenario.  Unfortunately that is one of the things where `symbolize_names` on the receiving would resolve that issue, but then it would have the same problem of `hash['id']` being `nil`, and that is not something I'm really advocating to change as it doesn't make sense in this case.

I think it makes sense to say "your received arguments will be always string keyed hashes", but I'm not sure it makes sense to say "you have to adhered to this explicit format for _sending_ arguments," especially when, as noted above, the format doesn't actually dictate the validity of the object serialization, instead just being an arbitrary way that just happens to match how the `JSON.dump` options are set up.  On top of that, as was noted by another commenter, ensuring that _all_ hashes match this explicit arbitrary format is far more difficult than ensuring that all usages in `perform` just make use of `hash['id']` and similar.

tenpaiyomi:
Also, I just want to note that I'm not trying to argue this point out of a sense of stubbornness or anything of the sort.  I'm advocating for this perspective because I do feel like this change will cause an undue burden and confusion on the end user, especially with the change several ruby versions back that the hash rocket was the less favored way of generating hashes over the `key: value` format, and so I feel like it's an important topic to actively discuss the different perspectives of to ensure the best user experience as possible.  I do greatly appreciate the back and forth dialog.

silverdr:
It's a year old thread but maybe somebody could shed a bit of light on this for me as this seems strictly related?
```
WARN: Job arguments to MyModule::Jobs::MyJob do not serialize to JSON safely. This will raise an error in Sidekiq 7.0. See https://github.com/mperham/sidekiq/wiki/Best-Practices or raise an error today by calling `Sidekiq.strict_args!` during Sidekiq initialization.

[2022-12-02T13:24:06.684+00:00] INFO: Enqueued MyModule::Jobs::MyJob (Job ID: 652e4346-9e09-4111-8795-10acf473040c) to Sidekiq(default) with arguments: {"object_id"=>1, "ip_address"=>"44.44.44.44"}
```
I pass a simple, two k/v pairs `Hash` with string keys. Nothing fancy, no symbols or complex objects. The above message confirms getting the hash in question. Even worse: I also get the same warning even if I positionally pass simple primitives (integer and string). What am I doing wrong here?

Info says the version is: `Sidekiq 6.5.7`

bdewater-thatch:
@silverdr I ran into something similar that looked odd initially, turns out I was serializing a BigDecimal instance as a value. Open a Rails console, set a breakpoint (ruby/debug or byebug: `break Sidekiq::JobUtil#json_safe?`), enqueue the job and inspect the actual objects classes there.

PR #5077, opened 2021-12-01
Removing delay extensions

### LINKED ISSUE #5076: Remove delayed extensions
They have been disabled by default since Sidekiq 5.0 and per #5064, Ruby 3.1 is proposing to flag the YAML load operation as unsafe. Remove them from core.

I fully expect and am fine with someone baking the code into a 3rd party gem which people can use if they wish to continue using the delayed extensions.

skatkov:
fixes #5076 

Completely removes delay extensions and all related code.


mperham:
The issue was an initial proposal. The changes need to be staged as part of a 7-x branch. We do not have a timeline for 7.0 yet.

I appreciate the energy but I don't want to break main just yet.

mperham:
Fixed in the 7-0 branch

PR #5080, opened 2021-12-06
Sidekiq 7.0 changes

mperham:
This branch holds all pending Sidekiq 7.0 changes. There is no timeline for release and we are still months away from release. Assume Railsconf in April 2022.

- [x] Removed delayed extensions [#5076]


lib/sidekiq.rb
```diff
@@ -54,7 +53,7 @@ module Sidekiq
   }
 
   def self.❨╯°□°❩╯︵┻━┻
-    puts "Calm down, yo."
+    puts "Take a deep breath and count to ten..."
```
bf4, line 56: 👀

dilraj-vidyard: How can I trigger this?

dilraj-vidyard (approved):

Gemfile.lock
```diff
@@ -215,3 +216,6 @@ DEPENDENCIES
   standard
   toxiproxy
   yalphabetize
+
+BUNDLED WITH
+   2.3.7
```
adamniedzielski, line 221: You somehow keep merging this in into this branch by mistake 🙈

mperham: This is my biggest gripe with Bundler. This insistence on using a precise version. It should have just been major version. Removed the .lock file.

mperham:
The metrics subsystem is a WIP and won't hold up 7.0. I expect lots of changes to it over the next few releases as it goes from unusable to "pretty decent".

jpcamara:
Excited to see this coming together! Totally fine to tell me to move this question to discussions or an issue (or just go away I guess 😅 ), but did the Sidekiq::Config concept get dropped for the Sidekiq 7 release? I didn't see it included in this PR or the list of features.

I was really excited to see https://github.com/mperham/sidekiq/blob/main/docs/global_to_local.md within https://github.com/mperham/sidekiq/pull/5340, and maybe wrongly assumed that the next step for embeddable sidekiq launching would happen here.

mperham:
Thanks for the reminder, I've been meaning to revisit this now that metrics and 6.5 are in a good place.

Gemfile
```diff
@@ -5,6 +5,7 @@ gemspec
 gem "rake"
 gem "rails"
 gem "redis-client"
+# gem "debug"
```
liijunwei, line 8: suggestion: as `Gemfile.lock` is not checked into git repo, maybe introduce trick like `Gemfile.local` to avoid this kind of commits
ref: https://github.com/rubocop/rubocop/blob/bd42ff683f2da12a01c829aaac95c0575de1f129/Gemfile#L30

web/assets/javascripts/base-charts.js
```diff
@@ -0,0 +1,104 @@
...
+}
+
+class Colors {
+  constructor() {
+    this.assignments = {};
+    this.success = "#006f68";
+    this.failure = "#af0014";
+    this.fallback = "#999";
+    this.primary = "#537bc4";
+    this.available = [
```
nateberkopec, line 13: I can't quite visualize what this will look like, but if I may suggest choosing colorblind-safe palettes for anything where colors must be distinguished, e.g. .https://lospec.com/palette-list/ibm-color-blind-safe

dilraj-vidyard:
Amazing job @mperham!

PR #5120, opened 2022-01-13
Add keyword arguments support in extensions

4e4c52:
This PR allows the delay extension to support keyword arguments.
I know the extensions will be deleted in 7.0 but it would be a nice addition in the meantime.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -26,6 +26,7 @@ bin/rails generate sidekiq:job ProcessOrderJob
 ```
 - Fix job retries losing CurrentAttributes [#5090]
 - Tweak shutdown to give long-running threads time to cleanup [#5095]
+- Add keyword arguments support in extensions
 
 6.3.1
 ---------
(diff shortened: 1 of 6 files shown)


mperham:
Nice, good catch.

There doesn't appear to be a test for the actual perform call, asserting that any kwargs are filled in as expected.

4e4c52:
@mperham I added more test to check the keyword arguments forwarding 😉

4e4c52:
I added a condition to check if the kwargs are empty so I do not pass them to the proxied object.
Looks like it has to do with what's described here: https://eregon.me/blog/2021/02/13/correct-delegation-in-ruby-2-27-3.html

mperham:
Thank you!

mperham:
I reverted this support. It was becoming a maintenance chore; they will be removed soon and I don't have any desire to increase their maintenance burden in the meantime.

PR #5134, opened 2022-01-23
Add keyword arguments handler in Sidekiq::JobRecord#display_args

### LINKED ISSUE #5133: Sidekiq 6.4.0 changes behavior of Sidekiq::JobRecord#display_args
Ruby version: 2.7.5
Rails version: 6.1.4.4
Sidekiq version: 6.4.0

There is a change in behavior of `Sidekiq::JobRecord#display_args` in Sidekiq 6.4.0 compared with Sidekiq 6.3.1

Sidekiq 6.3.1
Screen Shot 2022-01-22 at 3 49 52 PM

Sidekiq 6.4.0
Screen Shot 2022-01-22 at 3 50 38 PM

The codebase is the same, it is just the Sidekiq version changes between the two screenshots.

Here is the shorten code snippet of the `ZoomVideo::NotificationReminderService`
```ruby
class ZoomVideo
  class NotificationReminderService
    def initialize(zoom_video_id:)
    end

   def send_reminder_notification
   end
   
   def self.send_reminder_notification(*args)
      new(*args).send_reminder_notification
    end
  end
end
```

I guess this is related to #5120 where we add the support for keyword arguments in extensions, and we might need another method such as `Sidekiq::Job#display_kwargs` to handle this situation.

Please let me know if you think that is reasonable. I'm willing to add a PR for that.

Thanks for your great gem.

thachck:
Fix #5133

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -356,10 +356,10 @@ module Sidekiq
       @display_args ||= case klass
                 when /\ASidekiq::Extensions::Delayed/
                   safe_load(args[0], args) do |_, _, arg, kwarg|
-                    if kwarg.present?
-                      [arg, kwarg]
-                    else
+                    if !kwarg || kwarg.empty?
                       arg
+                    else
+                      [arg, kwarg]
                     end
                   end
                 when "ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper"
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -355,8 +355,12 @@ module Sidekiq
       # Unwrap known wrappers so they show up in a human-friendly manner in the Web UI
       @display_args ||= case klass
                 when /\ASidekiq::Extensions::Delayed/
-                  safe_load(args[0], args) do |_, _, arg|
-                    arg
+                  safe_load(args[0], args) do |_, _, arg, kwarg|
+                    if kwarg.present?
+                      [arg, kwarg]
+                    else
+                      arg
+                    end
                   end
                 when "ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper"
                   job_args = self["wrapped"] ? args[0]["arguments"] : []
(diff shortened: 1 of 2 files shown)


mperham:
This should be integrated into display_args.  kwargs are still job args.

thachck:
@mperham I have integrated kwargs into `display_args`. 
Please let me know if you would like another representation of the combined args/kwargs instead of the current `[["argument_a", "argument_b"], {:with=>:keywords}]` format.

lib/sidekiq/api.rb
```diff
@@ -355,8 +355,12 @@ def display_args
       # Unwrap known wrappers so they show up in a human-friendly manner in the Web UI
       @display_args ||= case klass
                 when /\ASidekiq::Extensions::Delayed/
-                  safe_load(args[0], args) do |_, _, arg|
-                    arg
+                  safe_load(args[0], args) do |_, _, arg, kwarg|
+                    if kwarg.present?
```
mperham, line 359: `present?` is a Rails method, right? Needs to work without Rails.

mperham (changes requested):

lib/sidekiq/api.rb
thachck: Thanks @mperham . I have updated it. Please have a look.

mperham:
I'm not sure if I want `[['arg1', 'arg2'], {keyword: arg}]` or `['arg1', 'arg2', {keyword: arg}]`. Anyone else have an opinion?

mperham (approved):

fractaledmind:
> I'm not sure if I want `[['arg1', 'arg2'], {keyword: arg}]` or `['arg1', 'arg2', {keyword: arg}]`. Anyone else have an opinion?

It's a tricky choice, but I ultimately come down on the side of the first option, where there are 2 sibling collection objects

thachck:
Thanks @mperham . Then it seems I do not need to update the representation of the arg/kwarg.

PR #5139, opened 2022-01-24
Fix deprecated uses of `Redis#pipelined` and `Redis#multi`

casperisfine:
Context: https://github.com/redis/redis-rb/pull/1059

The following is deprecated
```ruby
redis.pipelined do
  redis.get(key)
end
```

And should be rewritten as:
```ruby
redis.pipelined do |pipeline|
  pipeline.get(key)
end
```

Functionally it makes no difference.

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -54,14 +54,14 @@ module Sidekiq
     # O(1) redis calls
     def fetch_stats_fast!
       pipe1_res = Sidekiq.redis { |conn|
-        conn.pipelined do
-          conn.get("stat:processed")
-          conn.get("stat:failed")
-          conn.zcard("schedule")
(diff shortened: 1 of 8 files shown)


mperham:
How long has the new form been supported? 4.x? Do we need to update the gemspec version?

casperisfine:
> How long has the new form been supported?

3.1 at least: https://github.com/redis/redis-rb/blob/3.1/lib/redis.rb#L2096, possibly longer not sure. Since your gemspec is currently as `redis >= 4.2` I don't think you need any changes.

casperisfine:
Nxst build should be green.

mperham:
Big fan of this change. I've never liked that "magical" form of the API.

casperisfine:
Glad you like it. I wanted to deprecate this for years.

Issue #5131, opened 2022-01-21
Serialization interface


fractaledmind:
I maintain a newer Ruby gem for helping to make Jobs (whether ActiveJob or Sidekiq) more resilient by leaning on ACIDic database transactions: http://github.com/fractaledmind/acidic_job

At the heart of this approach is essentially tracking job runs via a database record. One essential ingredient for this to work is to be able to serialize and deserialize jobs, so that I can store a job run and then start that job with the job queue. ActiveJob provides the following interface which makes this straightforward:

```ruby
SomeJob.deserialize(some_job_hash)
SomeJob.new(arguments).serialize
```

I have patched around Sidekiq's lack of a similar interface, but it has required me to essentially copy and re-execute the code in `Sidekiq::Client#push` (for serialization) and `Sidekiq::Processor#dispatch` (for deserialization). These are both hacks that are highly liable to break if anything in Sidekiq changes. 

I am curious how amenable you are to the idea of working to bringing the same basic serialization interface to Sidekiq?

I should note that there is one major architectural difference that makes this a non-trivial feature—ActiveJob treats the object instance as the primary entity that holds all information, including `perform` arguments; Sidekiq doesn't put much emphasis on the object instance, and all of the "magic" happens via the `perform` method. There are no arguments and no `jid` until `perform_async` ends up at the Redis push. If you call `SomeWorker.new.perform()`, you will have no `jid`

I have worked around this by prepending a custom `perform` method into Jobs to allow me to capture the arguments such that I can then serialize them. However, a Sidekiq job cannot be initialized with arguments and then make a call to `perform`. I also manually add my own `jid` if the job is called synchronously. 

So, I wanted to get a sense of your interest in these kinds of features in Sidekiq core before attempting a PR, as I don't see a path to a very cosmetic PR that simply adds serialization. I imagine some re-imaging of the Sidekiq internals would likely be needed, and that is no simple thing.

mperham:
Could you take over Sidekiq::Client#raw_push and redirect the payloads into database records?

Give me time to grok and think about this over the weekend.

fractaledmind:
Perhaps, but that isn't a full solution to my needs. I need to be able to allow a job to be "staged" and not enqueued yet, through something like a `perform_transactionally` sibling method to `perform_async`. In this context I need to serialize the job but be sure to not push it to Redis. Only once the database transaction commits would I then want to take that serialized job and push it to Redis. 

[...]

PR #5146, opened 2022-01-26
Job serialization

fractaledmind:
This is a proof-of-concept for how to add job serialization into Sidekiq re: #5131 

The goals here were to be purely additive and not require any changes to any existing user-facing code, while also allowing jobs to easily fetch the serialization of themselves and users to prepare the serialization of a job. 

If you wanted, you could also walk thru the commits one-by-one as I ensured that each commit passed the tests, so each commit presents a single non-breaking advance towards the full goal.


mperham:
I'd suggest you target the 7-0 branch. That removes the need for you to support the Delayed extensions.

fractaledmind:
Ah, great point. Will update

lib/sidekiq/client.rb
```diff
@@ -70,7 +70,7 @@ def initialize(redis_pool = nil)
     #   push('queue' => 'my_queue', 'class' => MyWorker, 'args' => ['foo', 1, :bat => 'bar'])
     #
     def push(item)
-      normed = normalize_item(item)
+      normed = item.key?("normalized") ? item : normalize_item(item)
```
adamniedzielski, line 73: Can you explain the reason for adding this `normalized` key? I'm guessing that it's related to the backwards compatibility that you mention in the description, but I still don't get why it's actually needed.

lib/sidekiq/extensions/generic_proxy.rb
```diff
@@ -24,9 +24,8 @@ def method_missing(name, *args, **kwargs)
         if marshalled.size > SIZE_LIMIT
           ::Sidekiq.logger.warn { "#{@target}.#{name} job argument is #{marshalled.bytesize} bytes, you should refactor it to reduce the size" }
         end
-        @performable.client_push({"class" => @performable,
-                                  "args" => [marshalled],
-                                  "display_class" => "#{@target}.#{name}"}.merge(@opts))
+        item = @performable.serialize({"display_class" => "#{@target}.#{name}"}.merge(@opts), [marshalled])
+        @performable.client_push(item)
```
adamniedzielski, line 28: Is this related to "Delayed extensions" that @mperham mentioned? This part of the Sidekiq code looks unfamiliar to me.

lib/sidekiq/worker.rb
```diff
@@ -351,12 +347,31 @@ def sidekiq_options(opts = {})
...
+    # Creates a new job instance. Takes the arguments that will be
+    # passed to the perform method.
+    def initialize(opts_or_args = {}, *args)
+      @opts = opts_or_args.is_a?(Hash) ? opts_or_args : {}
+      @args = opts_or_args.is_a?(Hash) ? args : Array[opts_or_args]
+    end
+    ruby2_keywords(:initialize)
+
+    def serialize
+      item = @opts.transform_keys(&:to_s).merge("class" => self.class, "args" => @args)
```
adamniedzielski, line 369: This looks like a good and useful extraction 👍

lib/sidekiq/worker.rb
```diff
@@ -351,12 +347,31 @@ def sidekiq_options(opts = {})
...
+      end
     end
+
+    # Creates a new job instance. Takes the arguments that will be
+    # passed to the perform method.
+    def initialize(opts_or_args = {}, *args)
+      @opts = opts_or_args.is_a?(Hash) ? opts_or_args : {}
+      @args = opts_or_args.is_a?(Hash) ? args : Array[opts_or_args]
+    end
+    ruby2_keywords(:initialize)
```
adamniedzielski, line 366: What's the purpose of this code above? It seems like it's trying to handle many different cases, but why?

lib/sidekiq/worker.rb
```diff
@@ -351,12 +347,31 @@ def sidekiq_options(opts = {})
...
+      def client_push(stringified_item) # :nodoc:
         pool = Thread.current[:sidekiq_via_pool] || get_sidekiq_options["pool"] || Sidekiq.redis_pool
-        stringified_item = item.transform_keys(&:to_s)
-
         Sidekiq::Client.new(pool).push(stringified_item)
       end
+
+      def serialize(opts = {}, args = [])
+        new(opts, *args).serialize
+      end
```
adamniedzielski, line 357: This is an interface, but would you say that it's a public interface that Sidekiq extensions can rely on?

1. it's a method inside the module `ClassMethods`
2. `ClassMethods` gets included in each worker where `serialize` becomes a class method
3. The topic of public / private class methods is "interesting" in Ruby and you need to go through some hoops to make class methods private (main reason why I don't like them).
4. Therefore this is not [boring Ruby code](https://blog.sundaycoding.com/blog/2018/11/22/boring-ruby-code/) which I like.
5. IMO we can talk about a public interface when there's a public instance method, preferably in a separate class.

I'm not sure if creating such a public interface was your intention or not, but I thought that I could leave a few thoughts.

adamniedzielski (commented):
Thanks for the contribution @fractaledmind 💯 I left some questions from my side 🙂

lib/sidekiq/client.rb
fractaledmind: There are many places throughout the Sidekiq codebase that use `Client#push` outside of the context of a Job instance. Those places need to still work, thus they need to follow the original codepath. But, inside the context of a Job instance, I want `Client#push` to go down a new code path (not calling `normalize_item` being a chief difference). In order to allow `Client#push` to determine which code path to go down, I created this mechanism. So, whenever the `item` payload passed to `Client#push` has already called `normalize_item` _before_ calling `Client#push`, that can be known.

Of course, I could have leaned on the particulars of how `normalize_item` mutates the `item` hash to introspect inside of `Client#push`. I didn't want to create such a hard (and hard to notice) dependency between those two parts of the codebase.

lib/sidekiq/extensions/generic_proxy.rb
fractaledmind: Yes. For a while now, Sidekiq has included the functionality to take arbitrary model methods and run them asynchronously thru Sidekiq (delaying their execution). When I recreate this branch off of the 7.0 head, this file will be gone I believe and I won't need this change.

lib/sidekiq/worker.rb
fractaledmind: A core change here is to make the Job instance a meaningful entity to Sidekiq. Currently, it serves little to no purpose. A Sidekiq job has 2 key bits of information it needs to be executed: its options (which queue, how many retries, etc.) and its arguments. Both are optional. You can have a Sidekiq job in 4 flavors:

- no opts, no args
- opts, no args
- no opts, args
- opts, args

Moreover, while in the end all arguments need to be in an Array, many Jobs only have 1 argument. So, for every flavor where there is an argument, there are 2 additional flavors:

- arg
- [arg_1, arg_2]

I want to create an `initialize` method signature that works for all 6 flavors:

- `Job.new()`
- `Job.new(queue: 'some_queue')`
- `Job.new(record.id)`
- `Job.new(record.id, 'another_arg')`
- `Job.new({queue: 'some_queue'}, record.id)`
- `Job.new({queue: 'some_queue'}, record.id, 'another_arg')`

This is my current best attempt at an `initialize` method that does what I want in each of those 6 scenarios.

lib/sidekiq/worker.rb
fractaledmind: This exists because essentially all of the Sidekiq worker interface exists as class methods. Thus, to make use of this new feature from `Job.perform_async`, I needed a class method. Since I needed it, I figured it likely useful to expose (in the same way that `#serialize` is publicly exposed).

lib/sidekiq/client.rb
adamniedzielski: Makes sense, thanks for explaining 👍

adamniedzielski:
@fractaledmind in the meantime I implemented https://github.com/mperham/sidekiq/pull/5182 so you got some conflicts 💥 could you take a look at my PR and say if you think that this PR is still relevant? Or maybe my changes already help you with the implementation of your gem? To be honest, I'm still struggling a bit with understanding the real need for your changes and now I spent quite some time looking at these classes.

It seems to me that you're trying to avoid the repetition of this one line of code:

```ruby
@opts.transform_keys(&:to_s).merge("class" => self.class, "args" => @args)
```

And in order to do that you're introducing 20 lines of code, a mechanism to maintain backwards compatibility, and some Ruby code - https://github.com/mperham/sidekiq/pull/5146/files#r794380302 - where the syntax makes me uncomfortable, even though I have 10 years of experience with Ruby (or maybe _because_ I have 10 years of experience with Ruby 🤔 ).

That's just an open question, but I wanted to check how this resonates with you.

fractaledmind:
@adamniedzielski: The core need is articulated in the 1 test I added. I need `SerializableWorker.new('required positional argument').serialize` to return _exactly_ what will be pushed to Redis, and I need to be able to inspect that _at any point in time_, that is, both before and after the job has been pushed to Redis. Your PR focuses on the Client, which is what pushes the job to Redis. I like your changes, so thank you for the nice simplification of code paths. But reducing the number of code paths in the "push the job to Redis" part of the stack doesn't help me to introspect a job from the outside or from within the job. This is my need.

Also, that 1 line is only possible because of the rest of my changes, as `@opts` doesn't currently exist in Sidekiq. As stated in a comment somewhere, currently Sidekiq delays and hides the serialization of jobs in such a way that you can't get at is from the outside before it has been put into Redis (the JID being a key piece of information that I need to guarantee is the same when I inspect it in advance of Redis and when I inspect after Redis). I'm happy to take any suggestions for how to simplify my PR to achieve this result tho.

mperham:
A Sidekiq job is a Hash with several elements. I don't know that a custom object (with that unfortunate `initialize` method) is the right choice. It feels like you might be trying to put a square peg in a round hole.

That `initialize` method is not going to work. Remove it and consider other options; assume we cannot create instances of the user's job classes. That doesn't mean you can't introduce new classes, e.g. `Sidekiq::PendingJob`.

lib/sidekiq/worker.rb
fractaledmind: Do you ever come back to a problem and wonder what exact thought-path led you to the result you came to the first time around? I have no idea why I got it in my head to put the options hash first and the arguments array second in the method signature, but that is what made this method so weird. Coming back to it, I realized that I should simply reverse that order and use normal Ruby. Sorry for creating some consternation at the outset, should've always looked like what it is now.

fractaledmind:
@adamniedzielski: Fixed conflicts and fixed the `initialize` method, also added more serialization tests to cover the range of method invocations more fully. Hopefully this version of the code communicates the goal much more clearly, which is, essentially, to simply move calling `normalize_item` up in the order of operations so as to make its output available to the outside world and to the job itself.

mperham:
Please retarget this to 7-0.

lib/sidekiq/worker.rb
```diff
@@ -355,13 +351,41 @@ def sidekiq_options(opts = {})
...
     end
+
+    ##
+    # Creates a new job instance. 
+    # +args+ are the arguments, if any, that will be passed to the perform method
+    # +opts+ are any options to configure the job
+    def initialize(*args, **opts)
+      @args = args
+      @opts = opts
+    end
```
adamniedzielski, line 378: This looks nice 👌

lib/sidekiq/worker.rb
```diff
@@ -355,13 +351,41 @@ def sidekiq_options(opts = {})
 
       def client_push(item) # :nodoc:
         raise ArgumentError, "Job payloads should contain no Symbols: #{item}" if item.any? { |k, v| k.is_a?(::Symbol) }
-        build_client.push_bulk(item.merge("args" => [item["args"]])).first
+        build_client.push(item)
```
adamniedzielski, line 354: This reverts the change from #5182. Is it a conscious decision or omission?

adamniedzielski (commented):
This version looks much nicer 👏

lib/sidekiq/worker.rb
fractaledmind: A mix. The code here is literally what `push` does, and the original code used `push`. I figured a quick "cleanup" to use the higher-order function made sense

adamniedzielski: My intention was that `Sidekiq::Client#push` is no longer used in `Sidekiq::Worker` and `push_bulk` becomes the main interface. I left `push` because it's public and people depend on it.

I'm not sure if this makes sense or not. This is what I was thinking.

fractaledmind:
@mperham: Closing in favor of #5232

lib/sidekiq/worker.rb
fractaledmind: Closing this PR in favor of the 7.0 PR (which doesn't yet have your code), but I would suggest that your refactor to have `push` rely on `push_bulk` is solid, but `push` remains an important and useful abstraction that should be used even within the Sidekiq codebase

PR #4985, opened 2021-09-06
Reconnect on "UNBLOCKED force unblock" errors

franzliedke:
Thanks a lot for Sidekiq!

In our setup with a Redis cluster, we are having regular failovers,
which is fine. We are currently improving the way our application
handles these situations and stumbled across a possible improvement
in Sidekiq itself.

By far the most frequent error in our error monitoring system is the
following:

    UNBLOCKED force unblock from blocking operation, instance state changed (master -> replica?)

These errors can occur during Sidekiq's long-running job fetching
command. This uses Redis' blocking BRPOP primitive. On failover in a
cluster setup, these commands are interrupted by the server.

This error causes the worker threads to be restarted, but as they are
bubbled up to the top, they cause a lot of spam in our error logging
systems. As related errors from other commands are being handled (see
#2550 and #4495) this way, it seems senbile to also handle this one.

#### Details

<details>
<summary>Stacktrace</summary>

```
Redis::CommandError: UNBLOCKED force unblock from blocking operation, instance state changed (master -> replica?)
  from vendor/ruby/2.7.0/gems/redis-4.2.5/lib/redis/client.rb:132:in `call'
  from vendor/ruby/2.7.0/gems/redis-4.2.5/lib/redis/client.rb:226:in `block in call_with_timeout'
  from vendor/ruby/2.7.0/gems/redis-4.2.5/lib/redis/client.rb:300:in `with_socket_timeout'
  from vendor/ruby/2.7.0/gems/redis-4.2.5/lib/redis/client.rb:225:in `call_with_timeout'
  from vendor/ruby/2.7.0/gems/redis-4.2.5/lib/redis.rb:1224:in `block in _bpop'
  from vendor/ruby/2.7.0/gems/redis-4.2.5/lib/redis.rb:69:in `block in synchronize'
  from monitor.rb:202:in `synchronize'
  from monitor.rb:202:in `mon_synchronize'
  from vendor/ruby/2.7.0/gems/redis-4.2.5/lib/redis.rb:69:in `synchronize'
  from vendor/ruby/2.7.0/gems/redis-4.2.5/lib/redis.rb:1221:in `_bpop'
  from vendor/ruby/2.7.0/gems/redis-4.2.5/lib/redis.rb:1266:in `brpop'
  from vendor/ruby/2.7.0/gems/sidekiq-6.2.1/lib/sidekiq/fetch.rb:47:in `block in retrieve_work'
[...]

PR #5159, opened 2022-01-31
Reconnect on known errors after failover when pushing jobs to Redis

Dome-GER:
This PR targets the discussion in https://github.com/mperham/sidekiq/discussions/4990.

In a Redis cluster setup, failovers will happen. In these cases,
a `Redis::CommandError` can be raised for different reasons, 
for example, when the server becomes a replica, when there 
is a "Not enough replicas" error from the primary, or when a 
blocking command is force-unblocked.

<details>
  <summary>Sample stacktrace</summary>
  
```
  Redis::CommandError: READONLY You can't write against a read only replica.
  from vendor/ruby/2.7.0/gems/redis-4.2.5/lib/redis/client.rb:216:in `call_pipelined'
  from vendor/ruby/2.7.0/gems/redis-4.2.5/lib/redis/client.rb:170:in `block in call_pipeline'
  from vendor/ruby/2.7.0/gems/redis-4.2.5/lib/redis/client.rb:313:in `with_reconnect'
  from vendor/ruby/2.7.0/gems/redis-4.2.5/lib/redis/client.rb:168:in `call_pipeline'
  from vendor/ruby/2.7.0/gems/redis-4.2.5/lib/redis.rb:2490:in `block in multi'
  from vendor/ruby/2.7.0/gems/redis-4.2.5/lib/redis.rb:69:in `block in synchronize'
  from monitor.rb:202:in `synchronize'
  from monitor.rb:202:in `mon_synchronize'
  from vendor/ruby/2.7.0/gems/redis-4.2.5/lib/redis.rb:69:in `synchronize'
  from vendor/ruby/2.7.0/gems/redis-4.2.5/lib/redis.rb:2483:in `multi'
  from vendor/ruby/2.7.0/gems/sidekiq-6.2.1/lib/sidekiq/client.rb:189:in `block in raw_push'
  from vendor/ruby/2.7.0/gems/connection_pool-2.2.5/lib/connection_pool.rb:63:in `block (2 levels) in with'
  from vendor/ruby/2.7.0/gems/connection_pool-2.2.5/lib/connection_pool.rb:62:in `handle_interrupt'
  from vendor/ruby/2.7.0/gems/connection_pool-2.2.5/lib/connection_pool.rb:62:in `block in with'
  from vendor/ruby/2.7.0/gems/connection_pool-2.2.5/lib/connection_pool.rb:59:in `handle_interrupt'
  from vendor/ruby/2.7.0/gems/connection_pool-2.2.5/lib/connection_pool.rb:59:in `with'
  from vendor/ruby/2.7.0/gems/sidekiq-6.2.1/lib/sidekiq/client.rb:188:in `raw_push'
  from vendor/ruby/2.7.0/gems/sidekiq-6.2.1/lib/sidekiq/client.rb:74:in `push'
  from vendor/ruby/2.7.0/gems/sidekiq-6.2.1/lib/sidekiq/worker.rb:240:in `client_push'
  from vendor/ruby/2.7.0/gems/sidekiq-6.2.1/lib/sidekiq/worker.rb:200:in `perform_async'
  from app/models/my_model.rb:49:in `schedule_sync!'
```
</details>

These errors can occur when pushing a job to Redis, so it needs
to reconnect to the current master node and retry.

Reconnecting to Redis is [handled for `Sidekiq.redis`](https://github.com/mperham/sidekiq/blame/main/lib/sidekiq.rb#L106) and has been 
extended in #4985, but the Client's `#raw_push` method directly 
accesses the Redis connection pool, i.e. scheduling circumvents 
`Sidekiq.redis`.

Therefore, proper retry logic (similar to `Sidekiq.redis`) is added.

### DIFF
--- a/lib/sidekiq/client.rb
+++ b/lib/sidekiq/client.rb
@@ -189,8 +189,23 @@ module Sidekiq
 
     def raw_push(payloads)
       @redis_pool.with do |conn|
-        conn.pipelined do |pipeline|
-          atomic_push(pipeline, payloads)
+        retryable = true
+        begin
+          conn.pipelined do |pipeline|
+            atomic_push(pipeline, payloads)
+          end
+        rescue Redis::BaseError => ex
+          # 2550 Failover can cause the server to become a replica, need
+          # to disconnect and reopen the socket to get back to the primary.
+          # 4495 Use the same logic if we have a "Not enough replicas" error from the primary
+          # 4985 Use the same logic when a blocking command is force-unblocked
+          # The retry logic is copied from sidekiq.rb
+          if retryable && ex.message =~ /READONLY|NOREPLICAS|UNBLOCKED/
+            conn.disconnect!
+            retryable = false
+            retry
+          end
+          raise
         end
       end
       true
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -103,6 +103,7 @@ module Sidekiq
         # to disconnect and reopen the socket to get back to the primary.
         # 4495 Use the same logic if we have a "Not enough replicas" error from the primary
         # 4985 Use the same logic when a blocking command is force-unblocked
+        # The same retry logic is also used in client.rb
         if retryable && ex.message =~ /READONLY|NOREPLICAS|UNBLOCKED/
           conn.disconnect!
           retryable = false


mperham:
Put a `# this is copied from sidekiq.rb` comment in both places so we know to update both places in the future.

franzliedke:
In the discussion thread where we talked about this, we also touched on extracting something re-usable here. This may become more important when potentially adding further configuration (e.g. for number of retries, and a delay) here. What do you think?

mperham:
I think it's worthwhile to extract the block but I don't want to extract configuration knobs because there is no right default value. I think it is ok as is unless someone can show a good reason otherwise.

Dome-GER:
Thanks for the hint. I added comments in both places, so you don't miss updating both in the future when it needs changes.

lib/sidekiq/client.rb
```diff
@@ -189,8 +189,23 @@ def enqueue_in(interval, klass, *args)
 
     def raw_push(payloads)
       @redis_pool.with do |conn|
-        conn.pipelined do |pipeline|
-          atomic_push(pipeline, payloads)
+        retryable = true
+        begin
+          conn.pipelined do
```
mperham, line 194: You're missing some recent changes to `main`. With the latest redis-rb, this should be `conn.pipelined do |pipe|` and pass in the pipe.

franzliedke:
> I think it's worthwhile to extract the block but I don't want to extract configuration knobs because there is no right default value. I think it is ok as is unless someone can show a good reason otherwise.

In our app (@Dome-GER is a colleague 😁), a way to increase the retry count (and a delay in between) would already help. Our master is currently identified via DNS, which takes a bit longer to update, so the first retry often fails again.

As for defaults, one retry and no delay (just like now) should work?

mperham:
Hmm, that's fair. If the "error" is a known edge case (e.g. migration failover) requiring specific settings, I think that's ok to specialize retry. Want to submit a PR which we can discuss?

franzliedke:
Sure. I was hoping you would say so... that's a good place to introduce a re-usable util then, too.

Issue #5158, opened 2022-01-30
Remove parallel code paths caused by "push" and "push_bulk"


adamniedzielski:
1. `Sidekiq::Worker.perform_bulk` calls `Sidekiq::Client#push_bulk`
2. `Sidekiq::Worker.client_push` calls `Sidekiq::Client#push`
3. `Sidekiq::Client#push_bulk` and `Sidekiq::Client#push` both end up calling `Sidekiq::Client#raw_push`
4. I believe that the intention is that they are consistent

Would you accept a PR where I try to:
1. refactor `Sidekiq::Client#push` to call `Sidekiq::Client#push_bulk` as it's just a special case of having one job to push
2. refactor `Sidekiq::Worker` to always use `Sidekiq::Client#push_bulk` so there's one interaction point between `Worker` and `Client`
3. (optionally) introduce lightweight dependency injection of `Client` so that it's possible to provide custom client implementation for `Worker` - for example through worker options

? Or am I missing some practical considerations here?

mperham:
I’m fine with refactoring but we need to be very careful about changing any documented public APIs.

PR #5162, opened 2022-01-31
Add failing test for scheduling in bulk and enqueued_at

adamniedzielski:
While working on https://github.com/mperham/sidekiq/issues/5158 I noticed an inconsistency with how `enqueued_at` is being handled with `push_bulk`.

I'm submitting this PR with a new failing test to demonstrate this inconsistency. The reason for it is [here](https://github.com/mperham/sidekiq/blob/cd553fa14cefe789684f3cf35b0f61e4e22b6305/lib/sidekiq/client.rb#L106). `push_bulk` always adds `enqueued_at`. `push` only adds it [conditionally](https://github.com/mperham/sidekiq/blob/cd553fa14cefe789684f3cf35b0f61e4e22b6305/lib/sidekiq/client.rb#L209).

I'm not 100% sure what's expected here, but it prevents me from unifying `push` and `push_bulk` 😅

### DIFF
--- a/test/test_scheduling.rb
+++ b/test/test_scheduling.rb
@@ -54,6 +54,16 @@ describe 'job scheduling' do
       assert job['created_at']
       refute job['enqueued_at']
     end
+
+    it 'removes the enqueued_at field when scheduling in bulk' do
+      ss = Sidekiq::ScheduledSet.new
+      ss.clear
+
+      assert Sidekiq::Client.push_bulk('class' => SomeScheduledWorker, 'args' => [['mike'], ['mike']], 'at' => 600)
+      job = ss.first
+      assert job['created_at']
+      refute job['enqueued_at']
+    end
   end
 
 end


mperham:
600 is not a valid value for `at`. `at` must be epoch seconds, so the value should always be over one billion.

You are correct, `enqueued_at` should be mutually exclusive with `at` since scheduling does not enqueue (unless the schedule time is less than now!).

adamniedzielski:
@mperham I'm going down a rabbit hole and getting more confused in this hole 🐰 

> 600 is not a valid value for at. at must be epoch seconds, so the value should always be over one billion.

Understood. But I copy-pasted it from the test above 😅 

> scheduling does not enqueue (unless the schedule time is less than now!).

If you look at the [existing test above](https://github.com/mperham/sidekiq/blob/cd553fa14cefe789684f3cf35b0f61e4e22b6305/test/test_scheduling.rb#L41) it proves exactly the opposite for `push_bulk`:

1. `'at' => 600`
2. we push two jobs
3. and the scheduled set grows by 2
4. I checked and `Time.now` is not mocked in this test

Is this a bug with a test passing by mistake or am I missing something obvious in my rabbit hole?

mperham:
Ok, find and fix what you can.

mperham:
I've fixed this test.

PR #5171, opened 2022-02-02
Further improve tests related to bulk scheduling of jobs

adamniedzielski:
It would be good to fix the build on `main` at some point, but sometimes you need to take a step back to move forward 😆 There are two commits in this PR and they should be reviewed separately.

### DIFF
--- a/test/test_scheduling.rb
+++ b/test/test_scheduling.rb
@@ -38,7 +38,7 @@ describe 'job scheduling' do
       assert_equal 3, ss.size
       assert_equal qs+1, q.size
 
-      assert Sidekiq::Client.push_bulk('class' => SomeScheduledWorker, 'args' => [['mike'], ['mike']], 'at' => 600)
+      assert Sidekiq::Client.push_bulk('class' => SomeScheduledWorker, 'args' => [['mike'], ['mike']], 'at' => Time.now.to_f + 100)
       assert_equal 5, ss.size
 
       assert SomeScheduledWorker.perform_in(TimeDuck.new, 'samwise')
@@ -59,7 +59,7 @@ describe 'job scheduling' do
       ss = Sidekiq::ScheduledSet.new
       ss.clear
 
-      assert Sidekiq::Client.push_bulk('class' => SomeScheduledWorker, 'args' => [['mike'], ['mike']], 'at' => 600)
+      assert Sidekiq::Client.push_bulk('class' => SomeScheduledWorker, 'args' => [['mike'], ['mike']], 'at' => Time.now.to_f + 100)
       job = ss.first
       assert job['created_at']
       refute job['enqueued_at']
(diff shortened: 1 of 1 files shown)


mperham:
No more PRs which break the build. You can submit a failing test but please then include a commit which fixes said test. Not two separate PRs.

adamniedzielski:
No worries, I'll include the fix in this PR when I have time to implement it and if I manage to find the bug 🙂 To be honest, I was surprised that you decided to merge https://github.com/mperham/sidekiq/pull/5162 without a fix ready. I opened the PR to demonstrate the bug and to start the conversation. And it could have been a good base branch for the future PR with the actual fix.

Feel free to revert on `main` or fix yourself, if the red build is making you uncomfortable 😆

mperham:
I guess I was under the impression you were actively working on the fix. Just a miscommunication, no worries. Main is already fixed.

adamniedzielski:
Ah, you did fix it in the meantime 💯

adamniedzielski:
Thank you! "actively working on the fix" is a bit hard to define, because contributing here is just my hobby in the evening. I was planning to fix it, but I'm happy that you did it faster.

I'll update this PR in the coming days and make sure that it includes both failing tests and fixes for them.

mperham:
👍🏼

One possible niceity: allowing `at` to be a Proc so we can schedule dynamically in bulk. Might be a little tricky to implement since raw_push handles `at`.  Like:

```ruby
now = Time.now.to_f
push_bulk('class' => Foo, 'args' => 10.times.to_a.map {|idx| Array(idx) }, 'at' => ->(idx) { now + idx })
```

Just a random thought, not necessary at all.

adamniedzielski:
@mperham I investigated a bit further and added one more failing test for demonstration.

I discovered that `Sidekiq::Client.push` also doesn't enqueue the job when the schedule time is in the past. In fact, this behavior is only in `Sidekiq::Worker` - https://github.com/mperham/sidekiq/blob/72f0dddac76fa8f3302faef05ac2eae4e475cba0/lib/sidekiq/worker.rb#L260. It is [duplicated there](https://github.com/mperham/sidekiq/blob/72f0dddac76fa8f3302faef05ac2eae4e475cba0/lib/sidekiq/worker.rb#L331). It is also [here](https://github.com/mperham/sidekiq/blob/72f0dddac76fa8f3302faef05ac2eae4e475cba0/lib/sidekiq/client.rb#L172), but this code path goes back to `Worker`.

Therefore, I'd say that the intention was to handle it in `Worker` and not `Client`. I think that the only change in this PR which is worth keeping is "600 => Time.now.to_f + 100" and the two new tests I'll remove as it's not expected that `Client` handles this logic.

adamniedzielski:
@mperham this is green now. It turned out smaller than I thought, but it's still an improvement.

PR #5173, opened 2022-02-03
Fix lingering command pipelining to Redis instance

nickrivadeneira:
Related to #5139 . Redis was still throwing deprecation warnings for the
`conn.expire` call. I also changed the `conn.hset` call since my assumption is any commands called directly on the Redis instance are undesirable.


mperham:
Excellent eye!

Issue #5172, opened 2022-02-03
Getting Started docs still showing old worker generator


halloffame:
Ruby version: 3.0.3
Rails version: 7.0.1
Sidekiq / Pro / Enterprise version(s): 
* sidekiq-pro (5.3.0)
* sidekiq (6.4.0)

## Problem

When I run the generator command described in this doc: https://github.com/mperham/sidekiq/wiki/Getting-Started#rails

```bash
rails generate sidekiq:worker hard
```

I get the following error

```
Could not find generator 'sidekiq:worker'.
Did you mean?  sidekiq:job
```

It looks like the worker generator was replaced with the job generator. https://github.com/mperham/sidekiq/pull/5055.

Can the docs be updated to reflect this change?

adamniedzielski:
Thanks for reporting! I believe that you’re right with identity the source
of the problem.

You could update the docs yourself as anybody has permissions to do so.
However, I wonder if it wouldn’t be confusing for people on the older
versions of Sidekiq.

I see two options:
1. only update the docs but mention there both versions of the generator
name
2. bring back the old name which was removed in #5055, as a alias. This
would match how it was done with Worker and Job classes.

@mperham what do you think?

franzliedke:
If "job" is the new recommended technology, I also think it should be used in as many places as possible (documentation, tests, examples).

IMHO, it's fine to have the documentation (especially "Getting Started") aim at the latest and greatest release. Would a warning ("min version" / "required version") suffice? This is just due to the nature of this kind of a wiki without support for versions.

mperham:
I agree that "getting started" and similar docs should assume the latest. No one should be getting started with Sidekiq 5.x right now.

adamniedzielski:
It's definitely better to have the documentation describe the current version that the old version that doesn't work anymore 👍 It's even better to have documentation tied to a particular version, but that's not possible with GitHub Wiki, I believe.

> No one should be getting started with Sidekiq 5.x right now.

No one should be adding Sidekiq 5.x to their app 👍 I can imagine a developer who starts on an existing project using Sidekiq 5.x, but they don't know Sidekiq and want to refer to the "getting started" docs. That's why I suggested to mention both versions of the name there. That said, mentioning the new name is already good progress 🙂

PR #5175, opened 2022-02-03
Use new terminology in examples

franzliedke:
I was checking the README and the "Getting started" page in the wiki and stumbled across this one.

IMO, the "Getting started" page should probably be updated, too. But there were concerns about confusion for people using older versions, see #5172. So I held off on that for now.


franzliedke:
That test failure seems ... unrelated. 🤔

PR #5181, opened 2022-02-04
add comment to describe the ❤️ emoji (means heartbeat)

dorianmariefr:


mperham:
There is already a method called heartbeat so the comment just adds confusion.

dorianmariefr:
Then what does ❤️ means? I feel like you chose this emoji because you already had an `heartbeat` method and couldn't reuse the name.

I think maybe it needs some refactoring to have better names, maybe even a single `heartbeat` method, but I don't know the code enough

PR #5183, opened 2022-02-05
Setup Docker for development

adamniedzielski:
This PR proposes simple Docker setup for development. It does add some maintenance cost, so it's up to you @mperham if you want to merge it or not. At the very least it's a little showcase how nice Docker for development can be /cc @amrAbdelwahab

### DIFF
--- a/.github/contributing.md
+++ b/.github/contributing.md
@@ -32,6 +32,22 @@ And in order to run the tests and linter checks:
 bundle exec rake
 ```
 
+#### Docker for development
+
+Alternatively you can use Docker for the development setup. This requires Docker
+and Docker Compose installed.
+
+```
+make build
+make bundle
+```
+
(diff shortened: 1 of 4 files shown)


adamniedzielski:
>  I'm reluctant to containerize my development workflow with a VC-backed commercial product.

Valid point and very up to date, considering that my company is buying Docker subscription for our engineers just so we can continue using Docker Desktop on Mac 😅 

Sadly I'm not knowledgeable when it comes to the alternatives.

adamniedzielski:
@mperham how do you want to move forward with this? I submitted it mostly as a PoC so I'm fine if you just close it.

mperham:
I would love to see any OCI container support that uses the standard tooling included with macOS or Ubuntu. I don't want to pull in Docker.

PR #5193, opened 2022-02-09
Fix set(sync: true) (#5192)

### LINKED ISSUE #5192: set("sync": true) is not working with a symbol
Ruby version: 2.7.4
Rails version: 6.1.4.1
Sidekiq : 6.4.1 / Pro 5.3.1

The doc says:

```
SomeJob.set("sync": true).perform_async(args...) # will run via perform_inline
```

But sync as a symbol is ignored.
We have to use a string instead to perform the job inline:

```
SomeJob.set("sync" => true).perform_async(args...)
```

The code involved is:
https://github.com/mperham/sidekiq/blob/main/lib/sidekiq/worker.rb#L193

```
def perform_async(*args)
  if @opts["sync"] == true
    perform_inline(*args)
  else
    [...]
  end
end
```
Options keys are stringified in other places like `perform_inline` or `perform_bulk`
I guess it should be done too in `perform_async`.

inkstak:
Fix #5192
Tests are missing.

### DIFF
--- a/lib/sidekiq/worker.rb
+++ b/lib/sidekiq/worker.rb
@@ -190,7 +190,9 @@ module Sidekiq
       end
 
       def perform_async(*args)
-        if @opts["sync"] == true
+        hash = @opts.transform_keys(&:to_s)
+
+        if hash["sync"] == true
           perform_inline(*args)
         else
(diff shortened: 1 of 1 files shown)


mperham:
I went ahead and did a bigger refactor to fix the issue.

PR #5196, opened 2022-02-14
Exit sidekiqmon with non-zero exit code if an error occurs

### LINKED ISSUE #5195: sidekiqmon doesn't exit with non-zero exit code when there's an error
It seems that sidekiqmon doesn't support connecting to a Redis URL that has transit encryption enabled. At least that's what I'm assuming is going on:

```
/shopvox # REDIS_URL=redis://master.xxx.use1.cache.amazonaws.com:6379/0 sidekiqmon
Sidekiq 6.4.0
2022-02-11 18:16:21 UTC

---- Overview ----
Couldn't get status: Connection timed out
```

I wonder if there could be an optional flag added to sidekiqmon that would allow us to attempt the connection with SSL enabled?

Ruby version: 2.6.6
Rails version: 6.0.4
Sidekiq / Pro / Enterprise version(s): 6.4.0

stephenwiebe:
Closes #5195. Exit monitor with status code 1 if an exception is raised and sidekiq status isn't able to be displayed.

### DIFF
--- a/lib/sidekiq/monitor.rb
+++ b/lib/sidekiq/monitor.rb
@@ -17,8 +17,7 @@ class Sidekiq::Monitor
       end
       send(section)
     rescue => e
-      puts "Couldn't get status: #{e}"
-      exit 1
+      abort "Couldn't get status: #{e}"
     end
 
     def all
--- a/lib/sidekiq/monitor.rb
+++ b/lib/sidekiq/monitor.rb
@@ -18,6 +18,7 @@ class Sidekiq::Monitor
       send(section)
     rescue => e
       puts "Couldn't get status: #{e}"
+      exit 1
     end
 
     def all


mperham:
This doc seems to indicate that callers should use `exit(false)` to denote a failure.

https://apidock.com/ruby/Kernel/exit

Can you confirm that would do the same thing?

stephenwiebe:
`exit(false)` is the same as `exit(1)` on all Unix-based systems and Windows as well, from what I've read. In theory there could be an OS that treats an exit code of 0 as an error and an exit code of 1 as a success, in which case `exit(false)` would be different from `exit(1)`, but I'm not aware of that being the case for any existing OS. Just to be on the safe side though, I think https://apidock.com/ruby/Kernel/abort might be the best bet here, which is equivalent to emitting a warning message and `exit(false)`.

PR #5199, opened 2022-02-15
Add yalphabetize for yaml normalisation

samrjenkins:
This PR introduces [yalphabetize](https://github.com/samrjenkins/yalphabetize) to the project. I would love to hear your thoughts/feedback.

It provides us with the ability to quickly normalise/alphabetise YAML files such as locales and to maintain this normalisation into the future.

Advantages of alphabetising YAML include:
- easier to locate specific keys in large files
- easier to spot duplicate keys
- easier to compare equivalent locale YAMLs for different languages

Yalphabetize can also be integrated into CI to ensure that future PRs maintain the alphabetisation of locale files.


mperham:
Any reason not to target this for 7.0? That’s a lot of noise for files which have not changed in years.

Is there a way to detect missing keys in YAML files? Some of the locales are missing phrases as they are added in English.

PR #5204, opened 2022-02-19
Add `perform_sync` alias method of `perform_inline`

yujideveloper:
Since v6.4.0, Sidekiq supports `perform_sync` class method in `Sidekiq::Job`.
https://github.com/mperham/sidekiq/blob/6a6fda35fa9d07c4fe8ed3b05609a0eba6781101/Changes.md?plain=1#L22-L27

However, when I called `perform_sync`, `NoMethodError` has occurred.
``` ruby
class SomeJob
  include Sidekiq::Job

  def perform
    puts "foo"
  end
end

SomeJob.perform_sync # => undefined method `perform_sync' for SomeJob:Class (NoMethodError)
```

I added `perform_sync` alias of `perform_inline` and fix it.

### DIFF
--- a/lib/sidekiq/worker.rb
+++ b/lib/sidekiq/worker.rb
@@ -294,6 +294,7 @@ module Sidekiq
       def perform_inline(*args)
         Setter.new(self, {}).perform_inline(*args)
       end
+      alias_method :perform_sync, :perform_inline
 
       ##
       # Push a large number of jobs to Redis, while limiting the batch of
--- a/test/test_worker.rb
+++ b/test/test_worker.rb
@@ -144,5 +144,9 @@ describe Sidekiq::Worker do
         end
       end
     end
+
+    it 'has the alias #perform_sync' do
+      assert_equal MyCustomWorker.method(:perform_inline), MyCustomWorker.method(:perform_sync)
+    end
   end
 end


mperham:
Are you sure you have the latest code? This alias is in 2ed5af0a6.

yujideveloper:
@mperham 

I tried v6.4.1 and latest main branch(6a6fda35fa9d07c4fe8ed3b05609a0eba6781101).
But it has occurred `NoMethodError`.

I understand that the alias is added in 2ed5af0a68e21a0695a80d403a5f2295034e8de8. But It is not same this PR one. It added to `Sidekiq::Worker::Setter` but not to `Sidekiq::Worker::ClassMethods`.
So, a job class that included `Sidekiq::Job` have not `perform_sync` class method.

Could you review again?

----

reproduction code:
``` ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"

  # Those don't work
  gem "sidekiq", "6.4.1"
  # gem "sidekiq", github: "mperham/sidekiq", ref: "6a6fda35fa9d07c4fe8ed3b05609a0eba6781101" # ref of latest main branch 

  # This works
  # gem "sidekiq", github: "yujideveloper/sidekiq", ref: "feature/perform-sync-alias"

  gem "activesupport"
end

require "sidekiq"
require "active_support"
require "active_support/core_ext/string/inflections"

class SomeJob
  include Sidekiq::Job

  def perform
    puts "foo"
  end
end

SomeJob.perform_sync
```

adamniedzielski:
I can confirm that something doesn't work here 🤔 I added a clear high-level failing test in https://github.com/mperham/sidekiq/pull/5208.

mperham:
Fixed

PR #5208, opened 2022-02-22
Add failing test for perform_sync

adamniedzielski:
Related to #5204

Please don't merge this PR without having a fix ready.

### DIFF
--- a/test/test_worker.rb
+++ b/test/test_worker.rb
@@ -145,4 +145,17 @@ describe Sidekiq::Worker do
       end
     end
   end
+
+  describe '#perform_sync' do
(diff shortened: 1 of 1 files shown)


test/test_worker.rb
```diff
@@ -145,4 +145,17 @@ def call(*args)
       end
     end
   end
+
+  describe '#perform_sync' do
+    class MyCustomWorker2
```
adamniedzielski, line 150: Side note: I believe that something is wrong with the test setup above, because when I call it `MyCustomWorker` it fails for a strange reason. I didn't investigate it further.

PR #5223, opened 2022-02-27
Ignore Docker config in git

adamniedzielski:
I can't let go of #5183. It takes 10 minutes of my time to compile new Ruby version when I use rbenv and 10 seconds when it's done via Docker. Could we add these to `.gitignore` so that I can use it for my development without having dirty directory?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ tmp/
 pkg/*.gem
 .byebug_history
 development.log
+/Dockerfile
+/Makefile
+/docker-compose.yml


adamniedzielski:
Thank you! 💛

adamniedzielski:
Merging this also proved that https://github.com/mperham/sidekiq/runs/5345358897?check_suite_focus=true was a flaky test, because the main is green again.

adamniedzielski:
No idea what caused the flaky. The test got introduced in https://github.com/mperham/sidekiq/commit/237c70fe374142509e031dbbf136d57f43dd4913. The logic looks the same as in the case above. I don't know Redis well enough to be able to tell what might have caused this one more `total_connections_received` that failed the tests.`counts.size` was still 2 which proves that there was only one "problem" with the connection.

mperham:
🤷🏻‍♂️ I’m stumped too. If it continues to be a problem, we can remove the test. It’s a sketchy, “mutable global state” type of test.

PR #5230, opened 2022-03-04
Add event listeners after reload

### LINKED ISSUE #5184: Web UI event listeners disappear after page update when live polling is enabled
Ruby version: 2.7.5
Rails version: 5.2.5
Sidekiq / Pro / Enterprise version(s): 6.3.1 / 5.2.0

## Problem

When live poll is enabled in the web UI with an interval of `n` seconds, the following occur after `n` seconds have passed:

1. Confirmation to delete a queue is non-existent
2. Select all functionality is broken
3. Search functionality in retry queue is broken

Unfortunately, due to (1), a large job queue was deleted after someone accidentally clicked the delete button next to the queue. The confirmation message never appeared, as the web UI was in "live poll" mode, and the polling interval had passed.

I don't really care about (2) or (3), as they work fine when not in "live poll" mode, but I think (1) is a pretty serious issue, especially because the delete and pause buttons are so close together.

## Cause

When live polling is enabled, the following callback is triggered:

```javascript
function livePollCallback() {
  clearTimeout(livePollTimer);

  fetch(window.location.href).then(resp => resp.text()).then(replacePage).finally(scheduleLivePoll)
}
```

`replacePage` is called, causing all event listeners created in `ready(() => {` to be wiped out, and I don't see them being reinitialized.

---

P.S. thank you for making this software, it has enabled a very small team able to process billions of jobs each year!

excid3:
This re-initializes the event listeners after the DOM is updated from polling.

Fixes #5184

### DIFF
--- a/web/assets/javascripts/application.js
+++ b/web/assets/javascripts/application.js
@@ -9,7 +9,9 @@ var ready = (callback) => {
   else document.addEventListener("DOMContentLoaded", callback);
 }
 
-ready(() => {
+ready(addListeners)
+
+function addListeners() {
   document.querySelectorAll(".check_all").forEach(node => {
     node.addEventListener("click", event => {
       node.closest('table').querySelectorAll('input[type=checkbox]').forEach(inp => { inp.checked = !!node.checked; });
@@ -61,7 +63,7 @@ ready(() => {
       scheduleLivePoll();
     }
   }
-})
+}
 
 function updateFuzzyTimes() {
   var locale = document.body.getAttribute("data-locale");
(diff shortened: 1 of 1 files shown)


web/assets/javascripts/application.js
```diff
@@ -108,4 +110,5 @@ function replacePage(text) {
   document.querySelector('.status').replaceWith(header_status)
 
   updateFuzzyTimes();
```
ezekg, line 112: It looks like `addListensers` already calls `updateFuzzyTimes`, so this is probably now redundant.

excid3: Fixed. 👍

mperham:
Thank you, guys!

Issue #5225, opened 2022-03-01
Introduce `Sidekiq::Config`


adamniedzielski:
First step towards https://github.com/mperham/sidekiq/issues/4758, see https://github.com/mperham/sidekiq/issues/4758#issuecomment-1046208730 for the explanation on why I think it's the most reasonable first step.

Regardless of the topic of running Sidekiq client and server next to each other I see the following benefits:
1. easy to find - you search for config and it has its own file / module / class / concept. You want to add new config option, you know where it should go
2. clean the code up a bit internally and make it visible which config options come from the command line, YAML, initializer
3. make it explicit which options apply to server-only, to client-only, or to both
4. differentiate between "scalar" config values (boolean flag, array of queues) and "vector" values (default Redis connection pool). The labels "scalar" and "vector" help me, because I studied some physics and this metaphor resonates with me, but maybe somebody has a better naming suggestion for this kind of distinction?
5. making the God of Immutable Objects happy. I could write more here, but I'm going to save my time by turning this point into a joke and I can pull it through

@mperham before I move forward with the implementation I wanted to check in with you:

1. Are you in general onboard with the idea of introducing `Sidekiq::Config`?
2. Do you have any expectations ("I need X to be satisfied so I can accept the PR")?
3. Do you have any personal wishes ("It's a good opportunity to also do Y")?
4. Do you have any blessings for me ("Z is going to be dragons, but good luck with that!")?

mperham:
I like where you’re going here. I dislike the global Sidekiq.options blob and seeing people muck about with it. Really grinds my gears.

mperham:
As long as we continue to support the existing documented config knobs, I’m happy with any refactoring that might clean up configuration, esp if it better enables embedding Sidekiq. Today Sidekiq really doesn’t have a supported way to start a Sidekiq instance within another process (e.g. starting Sidekiq within a puma process).

Issue #4758, opened 2020-12-07
Add API for running a Sidekiq server in-process


mperham:
Today, Sidekiq will only execute jobs if started at the command line via `sidekiq` or `sidekiqswarm`. Rails 6.1 adds a new railtie API (#4747) for starting development tools inside the Rails server process. Apps like Discourse have created their own unofficial solutions for running Sidekiq in-process.

Research and prototype an API for booting a Sidekiq server in-process. Some things to account for:

0. Is there a big demand for this?
1. Should this Sidekiq instance be a singleton? Can I start two or N Sidekiq instances in the same process?
2. How does this mesh with the new Ractor API? Can we reduce the amount of global data and APIs we touch?
3. How does this affect Sidekiq's "client" and "server" model when one process is both?
4. All configuration today is done with global APIs, how can we refactor to configure per-instance?

The closest thing to this API we have today is `Sidekiq::Launcher`.

ioquatix:
I can give you some ideas about this since all the applications (falcon, rubydns) are all built in an isolated way.

0. I would say it no, but that it does encourage good design/isolation.
1. My opinion about this is yes, instances should be isolated so starting one or more should not be a problem.
2. Your code should not need to know about Ractor. IF someone decides to make multiple ractors, your code should still work (by being isolated).
3. If the code is isolated, it should not matter.
4. Configuration should be captured in an isolated, per-instance way.

We can elaborate on any of the above.

sedubois:
0. I would first make this work in dev for the reasons mentioned [here](https://github.com/mperham/sidekiq/issues/4747#issuecomment-739908038). This new API could additionally allow to run test and prod the same way, offering more predictable deployments.

adamniedzielski:
I read through the code in `Sidekiq::CLI` and this seems difficult to achieve, especially in a long-running project like Sidekiq where we need to move in small steps.

> Configuration should be captured in an isolated, per-instance way.

Agreeded 👌 Currently configuration is split between:
- options passed to the server process
- YAML config file
- initializer

and it's all global.

> How does this affect Sidekiq's "client" and "server" model when one process is both?

> If the code is isolated, it should not matter.

Currently it does matter. The simplest example is https://github.com/mperham/sidekiq/blob/6a6fda35fa9d07c4fe8ed3b05609a0eba6781101/lib/sidekiq.rb#L68-L93

with `redis` being the conflicting option.

[...]

PR #5231, opened 2022-03-06
Introduce Config class

adamniedzielski:
This is a WIP. It doesn't bring that much at this moment.


mperham:
Do you want to keep this open still?

adamniedzielski:
I'd like to continue this work at some point. I know that https://github.com/mperham/sidekiq/issues/5225 is needed for https://github.com/mperham/sidekiq/issues/4758 and it's a good improvement in general. I'm mostly just lazy to do the coding part. There's no need for this PR to stay open 👍

PR #5232, opened 2022-03-07
Job serialization (7.0)

fractaledmind:
This add both a public class and instance `serialize` method to Sidekiq::Worker. Tests cover all intended permutations of method signatures.


fractaledmind:
@mperham: Will close the other PR in favor of this one.

mperham:
Apologies, I hadn't recently merged main into 7-0. I'm also working on a Sidekiq::Client refactor too. I will push a branch if I get something I am happy with.

fractaledmind:
No worries, now that I've found the core of the pattern clearly, I'm confident I can spin up new PRs against new branches quickly. And no rush from my side. I'm just here to help.

fractaledmind:
Resolved the conflicts.

@mperham: Would you like to see what it would look like to add `Job.deserialize` and `Job#enqueue` to this PR? As stated in my recent comment in the issue, the only things I think Sidekiq would benefit from are these 4 methods:

```ruby
Job.serialize(*args, **opts) # => Hash
Job.new(*args, **opts).serialize # => Hash
Job.deserialize(hash) # => job instance
job.enqueue # => String (jid)
```

fractaledmind:
I won't push a commit yet, but just to show that `Job.deserialize` and `job.enqueue` can be very simple to add on top of this branch:

```ruby
class Worker
  module ClassMethods
    # ...

    def deserialize(serialized_job_hash)
      klass = serialized_job_hash["class"].constantize
      klass.new(*serialized_job_hash["args"], **serialized_job_hash.except("class", "args"))
    end
  end

  # ...

  def enqueue
    self.class.client_push(serialize)
  end
end
```

mperham:
This seems to have lost inertia and I never did see an API that I was really happy with. Do you want to rework this on top of the latest 7.0, give it another shot after 7.0, or something else?

fractaledmind:
I would like to rework on top of the current head of 7.0, but I'd love to better match an API that you prefer. I personally have little preference in the API itself, so long as I can call some method to get the serialized representation of the job from within the `perform` method (that is, from within the job instance).

If you have any particular preferences for the API, I'm happy to take this challenge on again and build toward that API.

Will close this PR for now and open a new one on the current head of 7.0 once we have some direction for the API

Issue #5228, opened 2022-03-04
Strict argument warning doesn't log the correct class name


rahul342:
Ruby version: 2.7.5
Rails version: 6.0.4.6
Sidekiq / Pro / Enterprise version(s): 6.4.0

If you are using an old version, have you checked the changelogs to see if your issue has [been fixed in a later version?
> yes

### What is happening?

[This line](https://github.com/mperham/sidekiq/blob/5776c7ae8711e14959d880dd907c9e79709d5b41/lib/sidekiq/job_util.rb#L24) logs `item['class']` which prints 
``` Job arguments to ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper do not serialize to JSON safely ```.  It's the same log irrespective of the job class. The class name is not useful.

### Expectation?
Looking at the `item` param, should we log `item['wrapped']` in the warning for ActiveJob case? Apologies if I am missing something in our config and this is not a bug.

mperham:
💯 Send a PR?

rahul342:
@mperham sure thing. Will create one.

rahul342:
@mperham https://github.com/mperham/sidekiq/pull/5235, I couldn't figure how to get the activejob test working.

PR #5235, opened 2022-03-08
Surface exact job class name in complex argument error message

rahul342:
Follow up to https://github.com/mperham/sidekiq/issues/5228

I couldn't get the ActiveJob test working, but here's a draft for fixing the message.

### DIFF
--- a/lib/sidekiq/job_util.rb
+++ b/lib/sidekiq/job_util.rb
@@ -13,15 +13,18 @@ module Sidekiq
       raise(ArgumentError, "Job 'at' must be a Numeric timestamp: `#{item}`") if item.key?("at") && !item["at"].is_a?(Numeric)
       raise(ArgumentError, "Job tags must be an Array: `#{item}`") if item["tags"] && !item["tags"].is_a?(Array)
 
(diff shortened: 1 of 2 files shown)


mperham:
My understanding is that ActiveJob marshals Symbols in a JSON-safe way that doesn't trigger the warning, note `_aj_symbol_keys` below. Sidekiq chooses not to add this additional complexity to its argument handling.

How are you getting AJ to trigger the warning message?

```
      "job_id"=>"a89f4d3e-ea57-40b7-8eb0-cabb20077f92",
      "provider_job_id"=>nil,
      "queue_name"=>"default",
      "priority"=>nil,
      "arguments"=>[{"foo"=>1, "_aj_symbol_keys"=>["foo"]}],
      "executions"=>0,
      "exception_executions"=>{},
      "locale"=>"en",
      "timezone"=>"UTC",
      "enqueued_at"=>"2022-03-08T20:51:06Z"}
```

rahul342:
@mperham it wasn't symbol. Seems like `BigDecimal` values that's creating issues.

mperham:
Thank you!

PR #5238, opened 2022-03-10
Improve compatibility with Semantic Logger

ojab:
Semantic Logger returns Symbol for `#level`:
```
[1] pry(main)> SemanticLogger[Sidekiq].level
=> :debug
```
so current code fails with
```
[2] pry(main)> Sidekiq.logger = SemanticLogger[Sidekiq]
=> #<SemanticLogger::Logger:0x00007f1d27984a18 @filter=nil, @level=nil, @level_index=nil, @name="Sidekiq">
[3] pry(main)> Sidekiq.logger.debug?
ArgumentError: comparison of Symbol with 0 failed
from /home/ojab/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/sidekiq-6.4.1/lib/sidekiq/logger.rb:39:in
`<='
```
Calling `#{level}?` methods on the logger directly instead of
comparing `#level` with integer fixes that.


ojab:
`1 workflow awaiting approval ` and `bundle exec rake test` is green locally, sending to review.

mperham:
Thanks!

PR #5248, opened 2022-03-18
Prototype transaction-aware client

adamniedzielski:


mperham:
Thanks, this is helpful. I'm looking at sidekiq-postpone to see what it does. I assume we'll have two modes of operation:

1. Classic
2. Everything in after_commit

Anything more fine-grained is up to the user to develop (by pulling in some other gem). I'm a little shy about adding a new dependency but at least it's optional and the codebase is small.

adamniedzielski:
Thank you for following up with your PR!

> I'm looking at sidekiq-postpone to see what it does.

I scrolled thought the codebase. I didn't find anything interesting there. I found some "interesting code" there, which is a bad sign for me.

Issue #5239, opened 2022-03-10
Delay job push within DB transaction


mperham:
Pushing jobs to Redis before the DB transaction has been a thorn in our side since Day One. I'd like to solve this if possible for Sidekiq 7.0. The simplest solution I can see is:

0. Prepend an AR transaction-aware module to override `Sidekiq::Client#raw_push`.
2. Collect the payloads created during a transaction into a thread-local array.
3. Use an after_commit callback which will be run by Rails' Executor to actually push the payloads.

ni3t:
What about jobs that would need to be queued even if the transaction is rolled back?

mperham:
@ni3t I've never heard of such a thing.

ianhall:
> @ni3t I've never heard of such a thing.

I have - when the transaction involves an update to another external resource (like a SaaS or similar), and a compensating action must be enqueued if the transaction fails.

Under those circumstances, it can be useful to push a job to Sidekiq that will run even if the transaction fails.

ni3t:
Sure, sorry I could have been more descriptive in my initial comment. 

We have several operations we perform that are designed to be composed *and* used individually. Because AR transactions can be nested, each of the composable operations is wrapped in an AR transaction on its own, and when composed, the outer composer will wrap them *all* in a separate, _outer_ AR transaction in order to avoid partial completion issues. 

A simplified code example might look like this:

```rb
class CreateOrder
	def self.call(params:)
		ActiveRecord::Base.transaction do
			order = Order.create(params)
			charge_result = ChargeCustomer.call(amount: params[:amount])
			order.transactions << charge_result
			# maybe a dozen other steps, each one of which could fail and roll back the order creation
			
			OrderGeocoder.perform_async(order: order)
		end
 	end
end

class ChargeCustomer
	def self.call(amount:)
		ActiveRecord::Base.transaction do
			transaction = Transaction.create(amount: amount)
			stripe_charge = StripeApi.create_charge(amount: amount)
			unless stripe_charge.success?
				ChargeFailureNotifier.perform_async(stripe_charge_id: stripe_charge.id)
				raise ActiveRecord::Rollback
			end
			transaction.update(stripe_charge_id: stripe_charge.id)
			transaction
		end
	end
end
```

In the example above, we would want the ChargeFailureNotifier to *always* perform, even if the transaction is rolled back, whether the failure rolls it back, or a later step rolls it back. However, we would need the OrderGeocoder to *not* perform if the transaction is rolled back.

[...]

PR #5249, opened 2022-03-19
Prototype transaction-aware Sidekiq::Client extension, #5239

mperham:
This is another stab at #5239 in the same vein as @adamniedzielski's #5248 and extends it further to support rollback-specific jobs, an activator API and testcases. NB: "XA" is industry jargon for "transaction".

```ruby
# enable the feature in your initializer
Sidekiq.transactional_push! # this makes "xa" default to "commit"

# usage
MyJob.set("xa" => nil).perform_async # enqueue immediately
MyJob.set("xa" => "commit").perform_async # delay enqueue after commit
MyJob.set("xa" => "rollback").perform_async # enqueue ONLY after rollback
```

I have not done much thinking about how this experimental feature might interact with other features. XA push delay could very well break features based around client-side push: things like the atomic push guarantee for `Sidekiq::Batch#jobs`, Pro's reliable push or Ent's unique jobs.


lib/sidekiq/xaclient.rb
```diff
@@ -0,0 +1,52 @@
...
+        after_rollback { super }
+      else # enqueue immediately
+        super
+      end
+    end
+
+    ##
+    # We don't provide transactionality for push_bulk because we don't want
+    # to hold potentially hundreds of thousands of job records in memory due to
+    # a long running enqueue process. TODO: wdyt?
```
adamniedzielski, line 38: As long as it's clear in the docs I think that it's a good decision. As you reverted most of the changes from #5158 I understand `push_bulk` as a special case. Combining two special cases increases the complexity and may therefore lead to unexpected results like the memory usage that you mentioned.

lib/sidekiq/xaclient.rb
```diff
@@ -0,0 +1,52 @@
...
+
+module Sidekiq
+  module XAClient
+    include AfterCommitEverywhere
+
+    ##
+    # Control job push within ActiveRecord transactions. Jobs can specify an
+    # "xa" attribute to define the policy to use:
+    #   * true or "commit" means enqueue the job after committing the current transaction.
+    #   * "rollback" means enqueue this job only if the current transaction rolls back
```
adamniedzielski, line 18: Does anybody really need this behaviour? I read https://github.com/mperham/sidekiq/issues/5239#issuecomment-1064356789. I'm not really convinced by it. Moreover, I understand

> Under those circumstances, it can be useful to push a job to Sidekiq that will run even if the transaction fails.

as "enqueue this job after the current transaction finishes - either commits or rolls back".

What did you mean @ianhall?

lib/sidekiq/xaclient.rb
```diff
@@ -0,0 +1,52 @@
...
+
+begin
+  require "after_commit_everywhere"
+rescue LoadError
+  Sidekiq.logger.error("You need to add after_commit_everywhere to your Gemfile for this to work")
+  exit(-127)
+end
+
+module Sidekiq
+  module XAClient
```
adamniedzielski, line 11: What's the advantage of introducing "industry jargon" to the codebase if it needs to be explained in the PR description?

lib/sidekiq/xaclient.rb
```diff
@@ -0,0 +1,52 @@
...
+      super
+    end
+  end
+end
+
+##
+# Use `Sidekiq.transactional_push!` in your sidekiq.rb initializer
+module Sidekiq
+  def self.transactional_push!(policy: "commit") # TODO: is this knob really necessary?
+    Sidekiq::Client.prepend(Sidekiq::XAClient)
```
adamniedzielski, line 49: What's the advantage of using multiple inheritance (including modules in Ruby is multiple inheritance) over composition from #5248?

sidekiq.gemspec
```diff
@@ -25,4 +25,5 @@ Gem::Specification.new do |gem|
   gem.add_dependency "redis", ">= 4.2.0"
   gem.add_dependency "connection_pool", ">= 2.2.2"
   gem.add_dependency "rack", "~> 2.0"
+  gem.add_dependency "after_commit_everywhere"
```
adamniedzielski, line 28: I think that it should be in `Gemfile`

adamniedzielski (commented):
> I have not done much thinking about how this experimental feature might interact with other features. XA push delay could very well break features based around client-side push: things like the atomic push guarantee for Sidekiq::Batch#jobs, Pro's reliable push or Ent's unique jobs.

Good point, me neither. Some things are probably going to stop working with this mode.

The failing test is strange.

```
Failure:
Sidekiq::Client::client#test_0003_enqueues [/home/runner/work/sidekiq/sidekiq/test/test_client.rb:121]:
--- expected
+++ actual
@@ -1 +1 @@
-{"retry"=>true, "queue"=>"default", "xa"=>nil}
+{"retry"=>true, "queue"=>"default", "xa"=>"commit"}
```

1. It only fails on 2.7, so the version in the middle.
2. You didn't change this test file.

🤔

lib/sidekiq/xaclient.rb
mperham: The advantage is that we can mix and match with other Modules. With composition, I have to instantiate some known type with known behavior. With `prepend`, everyone instantiates the same concrete type with the Modules dynamically mixed in. Sidekiq Pro has two separate Modules which mix into `raw_push` to provide reliability and atomic push.

sidekiq.gemspec
mperham: Ah, right, if we want it to be completely optional.

lib/sidekiq/xaclient.rb
mperham: Yeah, I want to keep push_bulk as simple as possible with efficient semantics so I think this is the best approach.

mperham:
I can't seem to figure out why the two tests are failing. Blocked...

adamniedzielski:
@mperham I tried figuring out why these tests are failing and couldn't either. I took a step back, picked up my original approach and opened https://github.com/mperham/sidekiq/pull/5291. Your tests are passing there. I took some code from this PR. I removed the code that I wasn't convinced is necessary.

PR #5253, opened 2022-03-22
Sidekiq 7: Replace `redis` by `redis-client`

casperisfine:
As discussed in https://github.com/redis/redis-rb/issues/1070#issuecomment-1074094773

Current shortcomings (https://github.com/redis-rb/redis-client/issues/6):

  - No `namespace` support. This one is debatable.
  - No connection `url` support (this will be added very soon)
  - No `reconnect_attempts` support. I'd like to add it, but in a smarter way.

There is 2 failing tests left, not too sure what's happening with them, but I'm sure they could be fixed. The objective here was to find missing features in `redis-client`.

cc @mperham

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ development.log
 /Dockerfile
 /Makefile
 /docker-compose.yml
+Gemfile.lock
+
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ source "https://rubygems.org"
 gemspec
 
 gem "rake"
-gem "redis-namespace"
+gem "redis-client", github: "redis-rb/redis-client"
 gem "rails", "~> 7.0"
 gem "sqlite3", platforms: :ruby
 gem "activerecord-jdbcsqlite3-adapter", platforms: :jruby
--- a/bin/sidekiqload
+++ b/bin/sidekiqload
@@ -55,7 +55,7 @@ rescue ArgumentError
   puts "Signal #{sig} not supported"
 end
 
-Sidekiq.redis { |c| c.flushdb }
+Sidekiq.redis { |c| c.call("FLUSHDB") }
 def handle_signal(launcher, sig)
   Sidekiq.logger.debug "Got #{sig} signal"
   case sig
@@ -100,7 +100,7 @@ Monitoring = Thread.new do
   while true
     sleep 0.2
     qsize = Sidekiq.redis do |conn|
-      conn.llen "queue:default"
+      conn.call("LLEN", "queue:default")
     end
     total = qsize
     # Sidekiq.logger.error("RSS: #{Process.rss} Pending: #{total}")
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -63,7 +63,7 @@ module Sidekiq
   # Configuration for Sidekiq server, use like:
   #
   #   Sidekiq.configure_server do |config|
-  #     config.redis = { :namespace => 'myapp', :size => 25, :url => 'redis://myhost:8877/0' }
+  #     config.redis = { :size => 25, :url => 'redis://myhost:8877/0' }
   #     config.server_middleware do |chain|
   #       chain.add MyServerHook
   #     end
@@ -76,7 +76,7 @@ module Sidekiq
   # Configuration for Sidekiq client, use like:
   #
   #   Sidekiq.configure_client do |config|
-  #     config.redis = { :namespace => 'myapp', :size => 1, :url => 'redis://myhost:8877/0' }
+  #     config.redis = { ::size => 1, :url => 'redis://myhost:8877/0' }
   #   end
   def self.configure_client
     yield self unless server?
@@ -92,14 +92,14 @@ module Sidekiq
       retryable = true
       begin
         yield conn
-      rescue Redis::BaseError => ex
+      rescue RedisClient::Error => ex
         # 2550 Failover can cause the server to become a replica, need
         # to disconnect and reopen the socket to get back to the primary.
         # 4495 Use the same logic if we have a "Not enough replicas" error from the primary
         # 4985 Use the same logic when a blocking command is force-unblocked
         # The same retry logic is also used in client.rb
         if retryable && ex.message =~ /READONLY|NOREPLICAS|UNBLOCKED/
-          conn.disconnect!
+          conn.close
           retryable = false
           retry
         end
@@ -110,14 +110,8 @@ module Sidekiq
 
   def self.redis_info
     redis do |conn|
-      # admin commands can't go through redis-namespace starting
-      # in redis-namespace 2.0
-      if conn.respond_to?(:namespace)
-        conn.redis.info
-      else
-        conn.info
-      end
-    rescue Redis::CommandError => ex
+      conn.call("INFO").lines(chomp: true).map { |l| l.split(":", 2) }.select { |l| l.size == 2 }.to_h
+    rescue RedisClient::CommandError => ex
       # 2850 return fake version when INFO command has (probably) been renamed
       raise unless /unknown command/.match?(ex.message)
       FAKE_INFO
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -55,13 +55,13 @@ module Sidekiq
     def fetch_stats_fast!
       pipe1_res = Sidekiq.redis { |conn|
         conn.pipelined do |pipeline|
-          pipeline.get("stat:processed")
-          pipeline.get("stat:failed")
-          pipeline.zcard("schedule")
-          pipeline.zcard("retry")
-          pipeline.zcard("dead")
-          pipeline.scard("processes")
-          pipeline.lrange("queue:default", -1, -1)
+          pipeline.call("GET", "stat:processed")
+          pipeline.call("GET", "stat:failed")
+          pipeline.call("ZCARD", "schedule")
+          pipeline.call("ZCARD", "retry")
+          pipeline.call("ZCARD", "dead")
+          pipeline.call("SCARD", "processes")
+          pipeline.call("LRANGE", "queue:default", -1, -1)
         end
       }
 
@@ -93,17 +93,17 @@ module Sidekiq
     # O(number of processes + number of queues) redis calls
     def fetch_stats_slow!
       processes = Sidekiq.redis { |conn|
-        conn.sscan_each("processes").to_a
+        conn.sscan("processes").to_a
       }
 
       queues = Sidekiq.redis { |conn|
-        conn.sscan_each("queues").to_a
+        conn.sscan("queues").to_a
       }
 
       pipe2_res = Sidekiq.redis { |conn|
         conn.pipelined do |pipeline|
-          processes.each { |key| pipeline.hget(key, "busy") }
-          queues.each { |queue| pipeline.llen("queue:#{queue}") }
+          processes.each { |key| pipeline.call("HGET", key, "busy") }
+          queues.each { |queue| pipeline.call("LLEN", "queue:#{queue}") }
         end
       }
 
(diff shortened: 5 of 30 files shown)


mperham:
The `c.call("CMD", args...)` pattern vs `c.cmd(args...)` is an interesting choice. I suppose you didn't want to have to add APIs for every new command but I like having explicit commands with arguments. I assume that RESP3 is self-documenting on how to parse the response so everything can go through one `call` method?

mperham:
Can you run `bin/sidekiqload` before and after and see if there is any performance difference? It would be fantastic if there's a statistically significant improvement due to the simpler client.

casperisfine:
> I suppose you didn't want to have to add APIs for every new command

That's the main reason yes, even more so because Redis tend to add new arguments to existing commands which sometimes makes for hard API changes in the gem, e.g. `EXISTS` that used to take a single key, so was returning a boolean, then became varargs.

> I assume that RESP3 is self-documenting on how to parse the response so everything can go through one `call` method?

Yes, contrary to RESP2, RESP3 has integers, floats, booleans etc, which allows for "dump" clients like this.

> Can you run bin/sidekiqload before and after and see if there is any performance difference?

`7-0` branch:
```
2022-03-22T15:44:37.165Z pid=63175 tid=1e5n INFO: Booting Sidekiq 7.0.0 with redis options {:db=>13, :port=>6380}
2022-03-22T15:44:40.876Z pid=63175 tid=1e5n ERROR: Created 500000 jobs
2022-03-22T15:44:40.876Z pid=63175 tid=1e5n ERROR: Simulating 1ms of latency between Sidekiq and redis
2022-03-22T15:45:42.360Z pid=63175 tid=1dqj ERROR: Done, 500000 jobs in 61.484105 sec
2022-03-22T15:45:42.360Z pid=63175 tid=1dqj ERROR: Now here's the latency for three jobs
0.0019919872283935547
0.0020089149475097656
0.002010822296142578
```

This branch:

```
2022-03-22T15:43:34.125Z pid=62499 tid=1egn INFO: Booting Sidekiq 7.0.0 with redis options {:db=>13, :port=>6380}
2022-03-22T15:43:38.040Z pid=62499 tid=1egn ERROR: Created 500000 jobs
2022-03-22T15:43:38.040Z pid=62499 tid=1egn ERROR: Simulating 1ms of latency between Sidekiq and redis
2022-03-22T15:44:17.342Z pid=62499 tid=1d7b ERROR: Done, 500000 jobs in 39.302515 sec
2022-03-22T15:44:17.342Z pid=62499 tid=1d7b ERROR: Now here's the latency for three jobs
0.0007328987121582031
0.0005278587341308594
0.00044798851013183594
```

Which is in line with the micro benchmarks in redis-client repo: https://github.com/redis-rb/redis-client/blob/master/benchmark/single.md, so I suppose the `redis` gem with `hiredis` driver would probably be a bit faster than that. But I do want to include an hiredis binding by default for MRI installs.

casperisfine:
> I suppose you didn't want to have to add APIs for every new command

That being said, a `method_missing` API could be tried, as long as we don't do any argument parsing, I wouldn't mind.

mperham:
60 -> 40 sec is a **incredible** speedup. Wow.

ziggythehamster:
This isn't really a Sidekiq suggestion but I really like what this gem is doing and I see a few places which you could get even better performance by leveraging a C extension. A couple that stand out to me:

### hash/map parsing

```
    def parse_map(io)
      Hash[*parse_sequence(io, parse_integer(io) * 2)]
    end
```

This could use `rb_hash_new_with_size` to create a perfectly sized Hash, and then fill it in the same loop you read the IO. Currently, `parse_sequence` does a loop and allocates an array it doesn't need, then `Hash[]` does a loop over that array. It can also be frozen (which you can do in Ruby and probably should regardless), which allows Ruby to pass around references to the same object where it would normally pass copies (though this could be nothing in the runtime right now, I haven't been following Ruby 3 as much as I should :)).

### numeric parsing

When you read the bytes from the IO, you could use something like `rb_str_new_cstr` and avoid retaining the string you're parsing. You'd have to dig some more but you can probably also use a simpler C standard library function to parse the number instead of `Integer()`/`rb_Integer` and friends, since those support radixes that you don't need (e.g., if your string read in `"0xFF"` then you'd have 255 there...but should instead have an error)

### sets

This probably gets into diminishing returns territory, but since `Set` is written in Ruby, it's easy to see that the `parse_sequence` improvement I mentioned above for `Hash` would apply there too. You'd either have to create a new instance of `Set` without calling the initializer and then set the `@hash` ivar or create a subclass (potentially in Ruby) which gets passed the backing Hash. This could potentially break if this changes upstream, so probably not worth doing given the maintenance overhead :).

Also all of the above could be done where there is a Ruby and a C implementation, so it works on JRuby.

igorwwwwwwwwwwwwwwwwwwww:
It'd be interesting to see a comparison of CPU profiles (e.g. stackprof) to understand where the time is being spent currently.

casperisfine:
> by leveraging a C extension

That's planned, see https://github.com/redis-rb/redis-client/issues/5. I'd like to first look at binding `hiredis` (that's work in progress), but depending on how it goes I might just scale down to just some small C helpers to speed up the IO and parsing.

casperisfine:
@mperham do you know how essential `redis-namespace` is? Because I looked at what it would take to support it, but it's a major pain.

You basically have to implement the whole `keyspec` thing https://redis.io/docs/reference/key-specs/, and even then there's still a bunch of edge cases, and it definitely won't be fast.

How would you feel about either dropping `namespace` support or doing it in Sidekiq itself (by calling `redis_client` with already namespaced keys rather than to rely on a middleware of sorts).

mperham:
I’m definitely interested in dropping support as I documented it as a hack for years now.  

https://www.mikeperham.com/2017/04/10/migrating-from-redis-namespace/

This is an excellent reason to do so.

casperisfine:
Sounds good, so I can just remove the feature?

mperham:
Yes

casperisfine:
Ok, I removed the namespacing support.

### Features still missing in the client

I noticed there's a test about Redis sentinel, that's something I need to implement.

I also still need to add retries, but that one is simple.

### Important note about the previous benchmark

Not 100% sure what happened, but I assume I hit a bug that caused some of the 500k jobs to silently not be inserted, because after digging more into the perf, I can't reproduce the results anymore, and all clients (`redis-rb`, `hiredis-rb`, `redis-client`, `redis-client hiredis bindings`) all pretty much give similar results (~60s), which kinda makes sense because with 500k jobs, 10 threads and 1ms latency, you get an incompressible 50s of IOs, no matter what, so the 40s results just isn't possible.

So apologies for the faulty results and for not running the benchmark more than once 😞 .

casperisfine:
Ok, all the missing features have been implemented in redis-client.

I expect to release `redis-client 0.1.0` soon, one thing I'm considering before is to provide built-in connection pooling.

I also need to have a deeper look into a few SSL options (such as hostname verification).

But overall not much should change from the Sidekiq point of view, so if you have feedback on the current state of the PR, please share :)

mperham:
How do the benchmarks look now?

casperisfine:
> How do the benchmarks look now?

No visible difference, but as said before, that benchmark has an incompressible 50s spent on IOs, so in the end the Ruby code doesn't account for much.

However If I skip toxiproxy, we can see a small improvement:

```bash
byroot@whipstaff:sidekiq:(7-0)$ bin/sidekiqload 
2022-04-13T07:41:06.249Z pid=16089 tid=8hd INFO: Booting Sidekiq 7.0.0 with redis options {:db=>13, :port=>6379}
2022-04-13T07:41:11.330Z pid=16089 tid=8hd ERROR: Created 500000 jobs
2022-04-13T07:41:11.330Z pid=16089 tid=8hd ERROR: Simulating 1ms of latency between Sidekiq and redis
2022-04-13T07:41:34.274Z pid=16089 tid=829 ERROR: Done, 500000 jobs in 22.944555 sec
2022-04-13T07:41:34.274Z pid=16089 tid=829 ERROR: Now here's the latency for three jobs
0.0008571147918701172
0.000408172607421875
0.00041103363037109375
byroot@whipstaff:sidekiq:(7-0)$ bin/sidekiqload 
2022-04-13T07:41:37.439Z pid=16149 tid=abd INFO: Booting Sidekiq 7.0.0 with redis options {:db=>13, :port=>6379}
2022-04-13T07:41:42.320Z pid=16149 tid=abd ERROR: Created 500000 jobs
2022-04-13T07:41:42.320Z pid=16149 tid=abd ERROR: Simulating 1ms of latency between Sidekiq and redis
2022-04-13T07:42:04.406Z pid=16149 tid=9w9 ERROR: Done, 500000 jobs in 22.085094 sec
2022-04-13T07:42:04.406Z pid=16149 tid=9w9 ERROR: Now here's the latency for three jobs
0.00047779083251953125
0.0001289844512939453
0.0001537799835205078
byroot@whipstaff:sidekiq:(7-0-redis-client)$ bin/sidekiqload 
2022-04-13T07:39:43.433Z pid=15409 tid=avx INFO: Booting Sidekiq 7.0.0 with redis options {:db=>13, :port=>6379}
2022-04-13T07:39:48.326Z pid=15409 tid=avx ERROR: Created 500000 jobs
2022-04-13T07:39:48.326Z pid=15409 tid=avx ERROR: Simulating 1ms of latency between Sidekiq and redis
2022-04-13T07:40:08.180Z pid=15409 tid=9q5 ERROR: Done, 500000 jobs in 19.854274 sec
2022-04-13T07:40:08.181Z pid=15409 tid=9q5 ERROR: Now here's the latency for three jobs
0.00033211708068847656
0.00014925003051757812
0.00015091896057128906
byroot@whipstaff:sidekiq:(7-0-redis-client)$ bin/sidekiqload 
2022-04-13T07:40:11.254Z pid=15468 tid=aww INFO: Booting Sidekiq 7.0.0 with redis options {:db=>13, :port=>6379}
2022-04-13T07:40:16.164Z pid=15468 tid=aww ERROR: Created 500000 jobs
2022-04-13T07:40:16.164Z pid=15468 tid=aww ERROR: Simulating 1ms of latency between Sidekiq and redis
2022-04-13T07:40:36.182Z pid=15468 tid=9r4 ERROR: Done, 500000 jobs in 20.017992 sec
2022-04-13T07:40:36.182Z pid=15468 tid=9r4 ERROR: Now here's the latency for three jobs
0.00027108192443847656
0.00018215179443359375
0.00011014938354492188

```


That said, please benchmark on your end to confirm (lesson learned :))

mperham:
That looks like a 10% improvement. Sounds great to me! When do you plan to release a v1.0?

casperisfine:
> When do you plan to release a v1.0?

1.0 I don't know, probably not before quite a while. But I can't think of anything else needed for a `0.1.0`. So I'll sleep on it just in case and likely release it tomorrow morning.

mperham:
I need to be careful here. I can't have a major, major project like Sidekiq using a v0.1.0 that's been out for days or even weeks. Too experimental. I look forward to the release and hopefully a few months will tell us how stable it is. Do we need a better migration path, maybe support both old and new client gem with an adapter/monkeypatch for redis-rb? Bake in `call` and that should handle 90% of the call sites, i.e.

```ruby
def call(meth, *args)
  send(meth, *args)
end
```

casperisfine:
> an adapter isn't possible...

`redis-rb` has a `call` method as well, so it should be possible.

casperisfine:
Hum, actually nevermind, using `Redis#call` means we're using `RESP2`, so lots of types wouldn't be properly casted.

We can probably have a small wrapper to translate the API, but then we'd need to handle the different exception types. I'll try my hand at it, but I'm not super hopeful.

casperisfine:
> 'll try my hand at it

Opened as a new PR to avoid too much confusion: https://github.com/mperham/sidekiq/pull/5287

In the end it wasn't very hard at all.

PR #5275, opened 2022-04-03
Enable live poll from url params

Benratelade:
This is related to this discussion: 
https://github.com/mperham/sidekiq/discussions/5197#discussioncomment-2491400

When viewing the queued jobs in the web view, one can pass a `?poll=true` parameter in the URL which will enable polling.


myapp/db/schema.rb
```diff
@@ -10,13 +10,12 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2012_01_23_214055) do
```
mperham, line 13: This shouldn't be part of this PR.

Benratelade:
noted @mperham, I undid those changes to schema.rb.

PR #5287, opened 2022-04-15
Sidekiq 7: Add a compatibility layer for `redis-client`

casperisfine:
As discussed in https://github.com/mperham/sidekiq/pull/5253

Switching entirely to `redis-client` is deemed risky, so instead we can support both.

All is needed is a small translation layer, and some very minimal adjustments in the callers.

cc @mperham

PR #5298, opened 2022-04-26
Add a compatibility layer for `redis-client`

casperisfine:
As discussed in https://github.com/mperham/sidekiq/pull/5253
This is https://github.com/mperham/sidekiq/pull/5287 but targeting the `main` branch. The only ajustement was to replace `...` argument forwarding with `ruby2_keywords` for compatibility with Ruby 2.6 and older.

Switching entirely to redis-client is deemed risky, so instead we can support both.

All is needed is a small translation layer, and some very minimal adjustments in the callers.

cc @mperham


mperham:
As I had mentioned earlier, I think I'd prefer not to see connection stuff go into `Sidekiq` namespace. What do you think about using `Sidekiq::RedisConnection.adapter = :redis_client` so S::RC remains the public API for anything Redis-related?

casperisfine:
Done.

mperham:
Could you avoid overriding the entire `create` method and instead only override the build_client method? I think this would make things a lot cleaner. This way the public API is always `Sidekiq::RedisConnection.create(...)` and we don't need to duplicate the sizing and pool logic.

casperisfine:
> Could you avoid overriding the entire create method and instead only override the build_client method? 

As is, no because the `config` has to be created outside the `ConnectionPool` proc, so that all instances are built from it.

But I'll try to reduce the duplication.

casperisfine:
Done.

casperisfine:
> That doesn't seem right. Shouldn't they be mutually exclusive?

No because both `RedisConnection` and `RedisClientConnection` are unit tested independently, see: `test/test_redis_client_connection.rb`, so yeah I guess these tests are ran twice for no benefit. If it's a blocker I can find a way to only run them in their respective suite.

mperham:
> the `config` has to be created outside the `ConnectionPool` proc, so that all instances are built from it.

Is it possible to retrofit a config builder method onto the redis gem, even if monkeypatched, so both adapters can use `create`? I'd prefer to see `RedisConnection` remain the only class with adapters for "redis" and "redis_client". A process should only ever load one driver. I'd prefer to keep the mental model simple and ensure mutual exclusion so debugging is simpler.

I'm sorry if this is overly picky, I appreciate your patience with my concerns. I will have more time next week to dig into this and contribute changes rather than just critique.

casperisfine:
> Is it possible to retrofit a config builder method onto the redis gem, even if monkeypatched, so both adapters can use `create`?

I pushed a new commit which refactor all this, hopefully it satisfies your requirement.

> I'm sorry if this is overly picky, I appreciate your patience with my concerns.

No worries at all.

mperham:
MUCH better. Each testsuite run now only loads the appropriate adapter. Well done!

mperham:
* [x] Sidekiq test suite passes and Sidekiq boots up
* [ ] Sidekiq Pro test suite passes and boots up
* [ ] Sidekiq Enterprise test suite passes and boots up

Sidekiq Pro does not work due to missing the PUBLISH and PSUBSCRIBE commands. Also, this error when calling `hmset`:

```ruby
          conn.multi do |m|
              m.hmset(@key, *data)
```

```
Unsupported command argument type: NilClass
    /Users/mperham/.gem/ruby/3.0.3/gems/redis-client-0.2.1/lib/redis_client/resp3.rb:80:in `block in coerce_command!'
    /Users/mperham/.gem/ruby/3.0.3/gems/redis-client-0.2.1/lib/redis_client/resp3.rb:73:in `map!'
    /Users/mperham/.gem/ruby/3.0.3/gems/redis-client-0.2.1/lib/redis_client/resp3.rb:73:in `coerce_command!'
    /Users/mperham/.gem/ruby/3.0.3/gems/redis-client-0.2.1/lib/redis_client.rb:283:in `call'
    /Users/mperham/src/sidekiq/lib/sidekiq/redis_connection/redis_client_adapter.rb:38:in `simple_call'
    /Users/mperham/src/pro/lib/sidekiq/batch.rb:217:in `block (2 levels) in jobs'
    /Users/mperham/src/sidekiq/lib/sidekiq/redis_connection/redis_client_adapter.rb:93:in `block in multi'
    /Users/mperham/.gem/ruby/3.0.3/gems/redis-client-0.2.1/lib/redis_client.rb:341:in `build_transaction'
    /Users/mperham/.gem/ruby/3.0.3/gems/redis-client-0.2.1/lib/redis_client.rb:230:in `multi'
    /Users/mperham/src/sidekiq/lib/sidekiq/redis_connection/redis_client_adapter.rb:93:in `multi'
    /Users/mperham/src/pro/lib/sidekiq/batch.rb:216:in `block in jobs'
```

Can I make commits on this pull also? There's a few other fixes that I've done locally but I can't seem to push them.

mperham:
The issue with the NilClass appears to be this:
```
["b-2st9kbWgCFEncQ", "created_at", 1651274249.82103, "callbacks", "{}", "description", nil]
```

The batch `description` is an attribute whose value can be nil. Is there some way to support nil attributes?

mperham:
I have most of the Pro test suite passing now (now I conditionally add description if not nil) except for the pub/sub stuff, which I will wait for your advise.

One minor hangup seems to be `sismember`. redis-client returns 0/1. redis-rb returns true/false. I guess I should add a specialized call in CompatClient.

mperham:
I'm also struggling with a `set` call. How can I make this call in a way which is compatible with both gems?

```ruby
        conn.set("super_fetch_orphan_check", Time.now.to_f, ex: delay, nx: true)
```

byroot:
> Sidekiq Pro does not work due to missing the PUBLISH and PSUBSCRIBE commands.

What do you mean by missing? You mean in the "compat client" class? Yeah it will need to be exposed, the redis-client API is described here: https://github.com/redis-rb/redis-client#publish--subscribe

No idea how sidekiq-pro uses the subscriptions, but if you share a snippet I can look at how best to transcribe it.

> Can I make commits on this pull also?

Not too certain how to do that, I suppose I can add you as collaborator on my fork? I'll try that in a bit.

> The batch description is an attribute whose value can be nil. Is there some way to support nil values?

Redis hashes values are always string. `redis-rb` does coerce `nil` to `""`, I chose not to replicate this behavior as I believe it can lead to bugs. I'd suggest explictly casting all the hash values to string.

> How can I make this call in a way which is compatible with both gems?

```ruby
def set(key, value, ex: nil, nx: false)
  command = ["SET", key, value]
  command << ex if ex
  command << "NX" if nx
  @client.call(*command)
end
```

I also need to have a quick look at how the default argument "converter" works, I think I could make some small changes to have a default behavior closer to redis-rb.

It would be nice if `{foo: true, bar: 42}` was automatically flattened as `["foo", "bar", "42"]`, but I'm afraid it might cause some unexpected behaviors when passing a hash for something like `HMSET`. So maybe it could only apply if passed as keyword arguments, but again I need to dig a bit because it might not be possible to have a good backward compatibility on such behavior.

mperham:
Here's the basic pattern:

```ruby
Thread.new {
    Sidekiq.redis do |conn|
      conn.psubscribe(CHANNEL) do |on|
        on.pmessage do |pattern, channel, msg|
          $messages << msg
        end
      end
    rescue Sidekiq::Shutdown
    end
}
```

mperham:
The new pubsub API is so different from the previous, I'm not sure we can adapt it without changing the code. What do you think, something like this...?

```ruby
Thread.new {
    Sidekiq.redis do |conn|
      pubsub = conn.pubsub
      pubsub.call("subscribe", CHANNEL)
      loop do
        evt = pubusb.next_event
        (_, _, msg) = evt
        $messages << msg
      end
    end
  end
}

**Comment by byroot** [comment]
> What do you think, something like this...?

Yes, that's how it's meant to be used. Looks good to me. Is `$messages` a `Queue`?

File: lib/sidekiq/redis_connection/redis_client_adapter.rb [hunk 1]
```diff
@@ -32,6 +32,24 @@ def call(*args)
...
+            @client.call("PUBLISH", channel, msg)
+          end
+
+          def set(key, value, ex: nil, nx: false, px: nil)
+            command = ["SET", key, value]
+            command << "NX" if nx
+            command << "EX" if ex
+            command << ex if ex
+            command << "PX" if px
+            command << px if px
```
byroot, line 45: You can simplify this a bit:

```ruby
            command << "EX" << ex if ex
            command << "PX" << px if px
```

mperham:
Does pubsub have good test coverage? I'm seeing weird behavior in testing publish/subscribe:

```
[:publish, :test, "\x04\bS:\x1ASidekiq::Pro::Message\a:\tverb:\ttest:\fpayload{\x06I\"\bfoo\x06:\x06ETI\"\bbar\x06;\tT"]
["subscribe", "sidekiq:config", 1]
```

I'm publishing that marshalled payload but getting "1" back from `next_event`.

byroot:
You are getting your `subscribe` acknowledgement back. The events you want start with `["message"`, see: https://github.com/redis-rb/redis-client/blob/561058c9ff6b72f36bd7e4930499e3592f0e064d/test/redis_client/subscriptions_test.rb#L28-L33

mperham:
Yep, thanks for explaining. Another oddity, I'm seeing cases where result is sometimes 1 and sometimes true:

```ruby
      result, _ = Sidekiq.redis { |conn|
        conn.multi do |m|
          m.sadd("paused", name)
          Sidekiq::Pro::Config.publish(m, :pause, name)
        end
      }
      p [:pause, result]
```

Does `multi` always return the underlying command results so I can expand the returned array like that?

byroot:
> I'm seeing cases where result is sometimes 1 and sometimes true:

That's weird, `SAAD` should always return an Integer, if you have a repro I'm interested.

> Does multi always return the underlying command results 

Yes, it always return the return of the underlying `EXEC` command, which should always be an Array.

byroot:
> which should always be an Array.

Assuming the transaction succeeded.

By the way, I'm not sure why you'd need a `multi` here. At a glance a regular `pipelined` would do.

mperham:
One thing I'm doing which I'm not sure is ok: calling "publish" on a normal, non-pubsub connection. Is it ok to mix with normal commands?

byroot:
> Is it ok to mix with normal commands?

Yes, only `SUBSCRIBE` / `PSUBSCRIBE` shouldn't be mixed.

mperham:
This appears to be the source of my confusion. I think this is a questionable API. I'd suggest it only return Integer.

```
     def srem(key, member)
        send_command([:srem, key, member]) do |reply|
          if member.is_a? Array
            # Variadic: return integer
            reply
          else
            # Single argument: return boolean
            Boolify.call(reply)
          end
        end
      end
```

byroot:
> I think this is a questionable API.

Ah you meant with the `redis` gem? Yes, that's among the kind of problems I want to solve by pushing redis-client.

`SREM` used to take a single member, and then became variadic, that's how `redis-rb` was painted into a corner having to keep returning false when no member is deleted.

I'd love to change this in `redis-rb`, but I got so much flak when I made a similar change for `exists` that I don't know if I'll ever do it :/

mperham:
Wait, so I'm somehow using the redis-rb gem in the redis-client... Ok, that's the problem.

mperham:
I think this is at 100% except for release notes which I'm working on.

* Sidekiq 6.5 will allow people to enable usage of `redis-client`. This allows plugin maintainers to run their test suite against the redis-client gem using the CompatClient we provide.
* Sidekiq 7.0 will remove support for the old `redis` gem. We'll refactor the codebase but keep the CompatClient facade. The idea being that we'll use the command-based method aliases for the foreseeable future, e.g. `client.command(args)`. I don't envision migrating to using `client.call("COMMAND", args)` ever.

mperham:
https://github.com/mperham/sidekiq/wiki/Using-redis-client

casperisfine:
> I think this is at 100%

🎉 

> I don't envision migrating to using client.call("COMMAND", args) ever.

Interesting. If you have time summarizing your thoughts on this, I'd be very interested.

casperisfine:
> https://github.com/mperham/sidekiq/wiki/Using-redis-client

I took the liberty to make a minor edit: https://github.com/mperham/sidekiq/wiki/Using-redis-client/_compare/515c583f6ab888fc89a737c164df091b4c458396...309eef7f6c9f908ab5fdcfc4603f9983c7c5b4f3

Also from the wiki:

```ruby
Sidekiq.redis {|c| c.call("SADD", "myset", 1) } # good
```

`c` here would be the `CompatClient` right? We might need to delegate the `call` family of methods. I'm also not sure the `CompatClient` is clean enough to be exposed.

mperham:
Yes, the idea is to expose the CompatClient as the connection so we retain the more idiomatic API form of `conn.smembers(*args)`, rather than the lower level `conn.call("smembers", *args)`. If you want to pull something like CompatClient into redis-client itself, I would love that. I think it would make for a much easier migration path for people going from `redis` to `redis-client`.

casperisfine:
> If you want to pull something like CompatClient into redis-client itself, I would love that.

Yeah, I'm toying with this idea. I definitely don't want it to have command knowledge though, so it would be a single `method_missing` based decorator like described in https://github.com/redis-rb/redis-client/pull/18.

What worries me though is that It also means methods such as `sismember` would always return an Integer. So code such as `if redis.sismember(...)` would silently be bugged.

And also since I plan `redis 5.0` to be essentially a thin wrapper around `redis-client`, it begs the question of wether or not to just use that future `redis 5.0`.

mperham:
Yeah, when people are migrating apps to using `redis-client`, I think it's fair that they need to adjust their code to the minor API differences. Compatibility helpers like `method_missing` can really help minimize that code churn.

But Sidekiq isn't an app, I can't adjust the code until I can drop support for using `redis` in Sidekiq 7 so libraries using `redis` are the harder problem. Yeah, it sounds like Redis 5 is going to be something very similar to CompatClient.

Maybe you can provide a single arg method `sismember?` which always returns a boolean in 4.x and provide a compatibility helper in 5.x, so `sismember` can return only integer in 5.x?

casperisfine:
Ok, so after exploring several ideas, I think I settled on providing a `Decorator` class.

Here's what it looks like on the `redis-client` side: https://github.com/redis-rb/redis-client/commit/ec07a79a4494ad49e23fd3602d8c8e9654f0e807

And on the sidekiq side: https://github.com/casperisfine/sidekiq/commit/7e936725815c4d146cc77af53fc25ccaf4b90c3e

It allows to define methods that will be delegated on the client but also the pipeline object.
So:

```ruby
def exists?(key)
  @client.call("EXISTS", key) { |c| c > 0 }
end
```
means `redis.pipelined { |p| p.exists?("key")} # => [true]`

That would be one major footgun out of the way.

It also allow to have a blind `method_missing`, meaning any third party code using `Sidekiq.redis` may still work even if it use commands not explicitly defined by `CompatClient` (assuming there's no special argument or return value handling in `redis-rb`). NB: I remove the explicit commands definition to showcase the effectiveness, but we could keep them for performance purposes.

This also happens to be a very good building block to implement `redis-rb 5.0`, so I'm totally fine with embeding this, even more so since it's not required by default, totally opt-in.

mperham:
I'll need you to retrofit your Decorator into this branch once you are happy with it. I'd like to test it against the Pro and Ent test suites too before you cut a new release.

casperisfine:
@mperham I'm happy with the decorator now. Let me know what you think, and if it's good with you I'll cut a `0.3.0`.

lib/sidekiq.rb
```diff
@@ -209,7 +209,7 @@ def self.logger
 
   def self.logger=(logger)
     if logger.nil?
-      self.logger.fatal!
```
mperham, line 212: Inappropriate change, something's wrong here.

lib/sidekiq/manager.rb
```diff
@@ -107,7 +107,7 @@ def hard_shutdown
         jobs = cleanup.map { |p| p.job }.compact
 
         logger.warn { "Terminating #{cleanup.size} busy threads" }
-        logger.warn { "Jobs still in progress #{jobs.inspect}" }
```
mperham, line 110: Another unrelated change from main. Were some commits merged or rebased wrong? I believe the change is correct, this log was recently changed to debug on main.

casperisfine: Hum, that's weird, I must have messed something up, but no idea what, my last few commits look clean.

casperisfine:
Ok, I cleaned it up by properly rebasing. I had to force push the branch though.

mperham:
I just discovered a major concern: [redis-client is currently a native gem](https://github.com/redis-rb/redis-client/blob/ed42f162cd2d4f3b6f505a4acd313014cd928402/redis-client.gemspec#L31). As I mentioned before, I don't want Sidekiq to depend on any native gems by default. Would you be willing to separate the hiredis extension somehow?

casperisfine:
> As I mentioned before, I don't want Sidekiq to depend on any native gems by default.

Sorry, I must have missed this. Is there a writeup of the whys somewhere?

> Would you be willing to separate the hiredis extension somehow?

It could be done, but I'd prefer to understand what's the blocker first. The reason I chose to bundle the extension was to offer the best performance out of the box, and ensure it's properly tested.

The old `hiredis-rb` gem wasn't really maintained and had limitations, but that's not the case of the one in `redis-client`. It's also very conservative, if you are on Windows or JRuby / Truffle, the extension isn't compiled at all.

I can split it into a `hiredis-client` gem or something if there's really no other way, but I'd prefer not too.

byroot:
Ok, you make some good point. I'll look at making the split tomorrow.

casperisfine:
> I'll look at making the split tomorrow.

Done. `redis-client 0.4.0` is pure ruby, to enable hiredis, one can simply add `hiredis-client` to their Gemfile.

mperham:
It looks methods matching `?scan_each` seem to work fine but I'm having trouble with the Enterprise code which scans and displays a paginated list of existing rate limiters within Redis:

```ruby
     Sidekiq::Limiter.redis do |conn|
        while results.size < 25 && nxt != "0"
          (nxt, keys) = conn.scan(cursor, match: pattern, count: 200)
          keys.each do |key|
```

With redis-client, `keys` is nil. Any tips?

casperisfine:
Right, `redis-client`'s `scan` returns an `Enumerator` when called without block, that's what you're running into.

I'll push a change to improve the compat client.

casperisfine:
Here you go.

Before:

```
>> Sidekiq::RedisConnection.adapter = :redis_client
=> :redis_client
>> Sidekiq.redis { |r| p r.scan(0, match: "foo*") }
#<Enumerator: #<RedisClient:0x00000001093a1a10  ....
```
after:
```ruby
>> Sidekiq::RedisConnection.adapter = :redis_client
=> :redis_client
>> Sidekiq.redis { |r| p r.scan(0, match: "foo*") }
["0", ["foo"]]
=> ["0", ["foo"]]
```

mperham:
💥 🎉

casperisfine:
😻

PR #5299, opened 2022-04-26
Restoring logger compatibility upon logger being set to nil

### LINKED ISSUE #5297: Sidekiq 6.4.2 introduces a breaking change by using logger.fatal!
Ruby version: 2.6.6
Rails version: 6.0.0
Sidekiq / Pro / Enterprise version(s): 6.4.2

We use the `semantic_logger` and `rails_semantic_logger` gems (v4.10.0) in our project, next to Sidekiq.
By default, RailsSemanticLogger will replace the default Sidekiq logger with its own.
In our initializer, we set Sidekiq's logger to `nil` as we use ActiveJob and use its logger instead.

Doing so will call the [following code](https://github.com/mperham/sidekiq/blob/20cdf68a7362c9f7cc40508e294b96026754772b/lib/sidekiq.rb#L212) in Sidekiq:

```ruby
def self.logger=(logger)
  if logger.nil?
    self.logger.fatal!
    return self.logger
  end

  logger.extend(Sidekiq::LoggingUtils)

  @logger = logger
end
```

The problem is that at this point, RailsSemanticLogger has already replaced Sidekiq's logger and it does not implement the `fatal!` method, resulting in an undefined method error . Prior to Sidekiq 6.4.2, the code was:

```ruby
def self.logger=(logger)
  if logger.nil?
    self.logger.level = Logger::FATAL
    return self.logger
  end

  logger.extend(Sidekiq::LoggingUtils)
  @logger = logger
end
```

This makes it a breaking change and that was probably not the intention in a patch version 🙂

simonc:
As part of Sidekiq v.4.6.2, the way the log level is set when Sidekiq's logger is explicitly set to `nil` broke compatibility with gems like RailsSemanticLogger.

## Problematic case:

If a Rails app uses RailsSemanticLogger and sets `Sidekiq.logger = nil` in an initializer the app would not be able to load due to an undefined method error. This was caused by the fact that

* RailsSemanticLogger sets Sidekiq's logger to a SemanticLogger instance
* Setting Sidekiq's logger later on would trigger a call to logger.fatal!

SemanticLogger does not implement `fatal!` and this would be the cause of the error.

This PR fixes the isssue by reverting the way the log level is set.

(fixes #5297)

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -209,7 +209,7 @@ module Sidekiq
 
   def self.logger=(logger)
     if logger.nil?
-      self.logger.fatal!
+      self.logger.level = Logger::FATAL
       return self.logger
     end
 


mperham:
Thanks!

It'll probably be a few more weeks before 6.4.3 so I'd suggest you monkeypatch the method or use 6.4.1 until then.

Issue #5311, opened 2022-05-04
Sidekiq v6.4.2: Breaking change for client middleware


matt-taylor:
Ruby version:
2.7
Rails version:
5.2
Sidekiq / Pro / Enterprise version(s):
6.4.2 / 5.2.2 /2.2.2

## Problem:

Sidekiq v6.4.1 and below pass the Worker class object as a parameter to the middleware ([process_single](https://github.com/mperham/sidekiq/blob/v6.4.1/lib/sidekiq/client.rb#L74) -> [middleware](https://github.com/mperham/sidekiq/blob/v6.4.1/lib/sidekiq/client.rb#L235-L237)).

Sidekiq v6.4.2 changes the way arguments are passed to the middleware to the middleware on the client. v6.4.2 passes the [normalized parameters](https://github.com/mperham/sidekiq/blob/v6.4.2/lib/sidekiq/client.rb#L73-L76) to the middleware which are all `string` type.

This causes a breaking failure for custom middleware's that expect the `worker_class` variable in the middleware to be the class  that pushes the data rather than a string representing the name of the class

## Our mitigation Plan
Our plan is to downgrade to v6.4.1 until we can change our middleware accordingly. 

## Question
 Was this an intended change and will this remain going foreward or should we hold out for a patch?

mperham:
This was not intentional. I’d welcome a PR to fix.

matt-taylor:
@mperham Pushed up https://github.com/mperham/sidekiq/pull/5312

matt-taylor:
@mperham Thanks for merging in #5312.

When can we expect for the next release to be cut? And since v6.4.2 had a breaking change will it be yanked?

mperham:
I’m working on the Redis PR changes right now. This will go in 6.5.0,
hopefully by Railsconf.

PR #5312, opened 2022-05-04
Middleware Argument type

matt-taylor:
https://github.com/mperham/sidekiq/issues/5311

---

v6.4.2 introduced breaking changes to the argument types of the middleware client.

### DIFF
--- a/lib/sidekiq/client.rb
+++ b/lib/sidekiq/client.rb
@@ -71,7 +71,7 @@ module Sidekiq
     #
     def push(item)
       normed = normalize_item(item)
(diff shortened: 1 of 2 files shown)


danielvdao:
I could be wrong, but I think we need to add `items["class"` to `push_bulk` - otherwise we're missing one spot where we're not passing the class name. Here's a code example: https://github.com/mperham/sidekiq/compare/main...danielvdao:main#diff-ee2515b430eefbe153698ed9aaa4d6850e7404c2648ddef85586fcfd096062e1R113.

mperham:
Yes, you're missing a similar change in `push_bulk`.

PR #5320, opened 2022-05-10
constantize consolidation

dpep:
Thought it might be useful to consolidate the three different versions of constantize implemented within Sidekiq, and to avoid using the Rails one implicitly since it causes a bug if rails is not loaded.

```ruby
> class MockWorker
  include Sidekiq::Worker

  def perform; end
end
=> :perform
> MockWorker.perform_inline
        5: from /Users/dpepper/.rvm/rubies/ruby-2.7.6/bin/irb:23:in `load'
        4: from /Users/dpepper/.rvm/rubies/ruby-2.7.6/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
        3: from (irb):6
        2: from /Users/dpepper/.rvm/gems/ruby-2.7.6/gems/sidekiq-6.4.2/lib/sidekiq/worker.rb:294:in `perform_inline'
        1: from /Users/dpepper/.rvm/gems/ruby-2.7.6/gems/sidekiq-6.4.2/lib/sidekiq/worker.rb:221:in `perform_inline'
NoMethodError (undefined method `constantize' for "MockWorker":String)
> require 'rails'
=> true
> MockWorker.perform_inline
=> true
```

### DIFF
--- a/lib/sidekiq/worker.rb
+++ b/lib/sidekiq/worker.rb
@@ -218,7 +218,7 @@ module Sidekiq
         msg = Sidekiq.load_json(Sidekiq.dump_json(item))
 
         # prepare the job instance
-        klass = msg["class"].constantize
+        klass = Sidekiq.constantize(msg["class"])
         job = klass.new
         job.jid = msg["jid"]
         job.bid = msg["bid"] if job.respond_to?(:bid)
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -3,6 +3,7 @@
 require "sidekiq/version"
 fail "Sidekiq #{Sidekiq::VERSION} does not support Ruby versions below 2.5.0." if RUBY_PLATFORM != "java" && Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.5.0")
 
+require "sidekiq/constantize"
 require "sidekiq/logger"
 require "sidekiq/client"
 require "sidekiq/worker"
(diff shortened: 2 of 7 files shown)


lib/sidekiq/worker.rb
```diff
@@ -218,7 +218,7 @@ def perform_inline(*args)
         msg = Sidekiq.load_json(Sidekiq.dump_json(item))
 
         # prepare the job instance
-        klass = msg["class"].constantize
```
dpep, line 221: isn't this using the Rails `String#constantize` method?

mperham: Yep, `perform_inline` is a relatively new addition to the API. Looks like few are using it so far.

dpep:
moved logic to lib/sidekiq/constantize.rb and added test coverage - is this what you had in mind?  we could also just put it directly into lib/sidekiq.rb?

mperham:
Yes, that's closer. But thinking about it more, should we instead patch String? Should we defer to ActiveSupport if it exists in the load path?

I just did some research and found that AS's constantize is this:

```ruby
    def constantize(camel_cased_word)
      Object.const_get(camel_cased_word)
    end
```

That tells me that we could be a little aggressive and simply drop constantize completely. This requires Ruby 2.7 so we'll drop constantize in Sidekiq 7.0. In the meantime, I don't want to touch what's there. It's going away in the medium term so there's no need for further investment.

dpep:
yeah, it looks like ActiveSupport's implementation simplified dramatically around [v7](https://github.com/rails/rails/commit/7057ccf6565c1cb5354c1906880119276a9d15c0) after Ruby 2.5 support was dropped.  Dropping `.constantize` out of Sidekiq sounds like a good path forward...I'll find a workaround and will stay excited for Sidekiq v7!

PR #5331, opened 2022-05-16
code doc for "dead" sidekiq_option

brandonc:
I was testing our workers to ensure that only valid options were present, and found that 'dead' was a valid option, yet I could not find documentation about it.

Another unsurprising finding was that many authors thought the "retry" option was actually "retries", probably because it is commonly accompanied by an integer.


mperham:
I thought this was covered by the Error Handling wiki page. Does it need improvement?

brandonc:
I now see the option in the error handling wiki page! I suppose what I was searching for was a definitive list of options which is how I happened upon this code doc.

PR #5341, opened 2022-05-24
Using full page width, disabled truncate and increased  display width…

vizo:
… of arguments


mperham:
Please show before and after screenshots.

PR #5352, opened 2022-05-31
Fix drain_all infinite loop after clearing worker with symbolized queue

dsander:
`jobs_by_queue` saves the queue name of enqueued jobs as strings. When
`clear` is called for a worker that has it's queue name defined as a
symbol `clear` calls `Queues.clear_for` using a symbol. As a result
`jobs_by_queue` is not cleared, while `jobs_by_class` is. Calling
`drain_all` then runs into an infinite loop because of the mismatch of
the two state hashes.

### DIFF
--- a/lib/sidekiq/testing.rb
+++ b/lib/sidekiq/testing.rb
@@ -188,7 +188,7 @@ module Sidekiq
       end
 
       def clear_for(queue, klass)
-        jobs_by_queue[queue].clear
+        jobs_by_queue[queue.to_s].clear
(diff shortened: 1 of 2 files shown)


dsander:
Apart from the issue described in the commit message we ran into the same problem caused by a third party gem which is calling `Sidekiq::Queues[:queue_name].clear` . Are you open for a second PR which converts all queue names passed to the `Queue` testing API to strings or should this be fixed by the third party?

PR #5361, opened 2022-06-02
Hide tagline on md viewports

### LINKED ISSUE #5351: Make header bar always visible
On https://sidekiq.org, the header bar only appears once you start to scroll down. I don’t like this magical behavior which hides UI options and can confuse people. Make the header bar always visible.

Sidekiq.org is on the gh-pages branch.

omegahm:
Hiding the tagline on `md` yields the following result:

https://user-images.githubusercontent.com/178448/171727899-03bec94b-8c5c-4f9a-945d-5c7f0ca1f5d7.mp4

Fixes #5351


mperham:
Thank you so much!

PR #5369, opened 2022-06-05
Rdoc

SamArdrey:
This has the basic running of rdoc in here, it needs to be updated a bit to include more explanations and maybe formatted better, as well. Going to speak with you on it on Monday.


mperham:
I don't see a need to commit the generated content at all. The user or sites like rdoc.info will do it.

PR #5384, opened 2022-06-10
Per-job execution metrics

mperham:
We store time and counts per-queue, per-class and totals, daily. Data expires after 90 days. I'm not sure how much of this we will surface in the Web UI but I'd love to have help in building graphs and visualizations on top of this.

WDYT? Comments, suggestions, or other feedback?

- [x] Store per-job and per-queue count and execution time
- [x] Provide API to mark deployments so metric changes can be correlated to deploys
- [x] Provide a Metrics API to query existing metric data
- [x] Provide a Metrics tab in the Web UI to visualize metrics and marks


mperham:
Sidekiq has had the same stats format since 1.0. Mostly for Resque compatibility but also because the Web UI has been difficult to change, given my general ignorance of JavaScript.

I think it would be very useful for users to see execution data based on queue and job class (sidekiq-statistic did this years ago), especially data which can be used to discover or diagnose performance issues.

The data must power the Dashboard as it is today. I want the data to naturally expire over time so the keys should be based on the date/time. I propose a 90 day TTL for the data.

# Old Format

* `stat:processed` - integer of total job executions
* `stat:failed` - integer of total job failures
* `stat:processed:DATE` - integer of total job executions on DATE
* `stat:failed:DATE` - integer of total job executions on DATE

This was mostly copied from Resque. At that time, Redis 1.x had few operations for metrics beyond INCR.

# New format

Redis now has the Hash structure with operations like [HINCRYBY](https://redis.io/commands/hincrby/). I propose we move the daily stats into a Hash.

Hash `exec:DATE` with each hash having the following fields:

* `total|p` - integer of total job executions
* `total|f` - integer of total job failures
* `total|ms` - integer milliseconds spent executing jobs
* `q:QUEUE|p` - integer of total job executions from QUEUE
* `q:QUEUE|f` - integer of total job failures from QUEUE
* `q:QUEUE|ms` - integer milliseconds spent executing jobs from QUEUE
* `CLASS|p` - integer of total job executions for CLASS
* `CLASS|f` - integer of total job failures for CLASS
* `CLASS|ms` - integer milliseconds spent executing CLASS jobs

We special-case the name `total`. The `q:` prefix identifies queue-specific data. I can imagine a drop-down in the Web UI which allows the user to focus on a specific CLASS or QUEUE based on the data.

# No Change

No reason to drop these totals. I plan to retain them.

* `stat:processed` - integer of total job executions
* `stat:failed` - integer of total job failures

## Open Questions

* Is tracking millisecond execution per day useful? We can only get average execution time, not a histogram, and only over an entire day so there's not much to drill into for diagnosis. It would be useful more for long-term trends. Tracking hourly would be far more useful for finding performance issues in a new deploy.
* Should we track milliseconds for **failed** jobs? Failures may be due to timeouts and cause spikes in average execution time.

mperham:
Need to add support for arbitrary markers in time, e.g. deploys.

mperham:
I've removed per-queue metrics as I don't believe their value is greater than their complexity.

mperham:
Further discussion and prototyping leads me to these changes:

- [ ] Metrics home should show two tables, top N jobs by processed count and execution time
- [ ] Each table row should link to the job-specific metrics page showing the last day of data
- [ ] Initial spike should focus on high-resolution data (histogram for every 10 minutes) expiring quickly (24 hours?)
- [ ] Reach goal of hourly histograms stored for 7 or 14 days (maybe an Enterprise feature which uses a periodic job to aggregate the 10min 'grams?)

mperham:
Super relevant Redis op: https://redis.io/commands/bitfield/

natematykiewicz (commented):
I think you want `multi` instead of `pipelined`. The MULTI command is an atomic transaction in 1 command. A pipeline is multiple commands that can independently succeed or fail in a single network connection. For example, with a pipeline, it’s possible for the HSETNX to fail but the EXPIRE to succeed.

Though maybe that’s what you’re going for.

If the hash key already exists, do you want to reset the TTL?

https://redis.io/commands/multi/

mperham:
This PR has reached the point where it really needs a JS wizard to implement some amazing dynamic graphs. The data is all on the page in text format. Anyone interested in collaborating on it, maybe with Tuple?

hjiangwen:
How about we track the latency of job as well
https://github.com/mperham/sidekiq/pull/5452

mperham:
Latency has little to do with the type of job and far more variable based on queue config and backlog. Sidekiq Enterprise’s historic metrics will automatically track queue latencies for you.

Issue #5381, opened 2022-06-10
`Sidekiq::Client.push_bulk` with `at` broken in 6.5.0


christos:
Ruby version: `3.1.2`
Rails version: - 
Sidekiq / Pro / Enterprise version(s): `Sidekiq 6.5.0`

Since upgrading to version `6.5.0` the following code fails:

```ruby
#!/usr/bin/env ruby

require "sidekiq"

class MyJob
  include Sidekiq::Worker
end

Sidekiq::Client.push_bulk(
  "class" => MyJob,
  "args" => [[1], [2], [3]],
  "at" => [Time.now.to_i, Time.now.to_i, Time.now.to_i]
)
```

with 

```
/Users/christos/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/redis-4.6.0/lib/redis/commands/sorted_sets.rb:69:in `zadd': wrong number of arguments (ArgumentError)
from /Users/christos/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/sidekiq-6.5.0/lib/sidekiq/client.rb:223:in `atomic_push'
```

The `at` argument for `push_bulk` is not documented clearly but it is indeed supported, right? 

https://github.com/mperham/sidekiq/blob/f38cf3050cf62e34a38ef47cd5d812108e7308cf/lib/sidekiq/client.rb#L102-L104

I believe the root cause to be the addition of the splat here: 

https://github.com/mperham/sidekiq/commit/aaac999c6de54bfa21f09cf29c5473dadf9355fd#diff-ee2515b430eefbe153698ed9aaa4d6850e7404c2648ddef85586fcfd096062e1R223

Weirdly enough, with only two items `args`/`at` arrays  the error doesn't happen. It needs to have 3 or more to manifest itself.

byroot:
> I believe the root cause to be the addition of the splat here

Yes you are right. In `redis-rb` the argument handling is a bit finicky:

https://github.com/redis/redis-rb/blob/4e76b035a98ce2c7bd7666a1553fc4394279ca83/lib/redis/commands/sorted_sets.rb#L62-L69

I think we can just drop the splat. 

cc @casperisfine (assigning myself as first think to look at when I get back to work on Monday).

PR #5387, opened 2022-06-13
redis-client: Fix `ZADD` compatibility

casperisfine:
Fix: https://github.com/mperham/sidekiq/issues/5381

`redis-rb`'s `zadd` argument parsing is particularly strict and doesn't make it easy to have transparent compatibility.

The bug only happens if there is more than two elements being added, hence why we missed it.

I'll be improving the `zadd` signature in `redis-rb` as well to make the transition easier.

cc @mperham @christos

### DIFF
--- a/lib/sidekiq/client.rb
+++ b/lib/sidekiq/client.rb
@@ -220,7 +220,7 @@ module Sidekiq
 
     def atomic_push(conn, payloads)
       if payloads.first.key?("at")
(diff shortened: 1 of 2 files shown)


casperisfine:
Ok, I found the better way to be compatible with both: we can pass a single flat array.

MartynasM:
When will this be released?

mperham:
Tomorrow.

Issue #5391, opened 2022-06-15
v6.5 "breaking" change?


codebender:
Ruby version: 3.1
Rails version: 6.1.x
Sidekiq / Pro / Enterprise version(s): 6.5

Hey Sidekiqers!

Previously in order to remove Sidekiq's default exception handling logger, some projects would do something like this:
```ruby
Sidekiq.error_handlers << MyCustom::ExceptionHandler::Logger.new
Sidekiq.error_handlers.reject! { |handler| handler.is_a?(Sidekiq::ExceptionHandler::Logger) }
```
Here is an example from gitlabs: https://github.com/gitlabhq/gitlabhq/blob/7eca3f56625526ffa7f263c1fef0fcea34de8ca6/config/initializers/sidekiq.rb#L33-L40

It looks like PR https://github.com/mperham/sidekiq/pull/5340 might have created an unintended breaking change while progressing forward to v7. Does it make sense to note this as a breaking change in the Changes.md?

It looks like the fix would be to update the `reject!` from the example line to something like: 
```ruby
Sidekiq.error_handlers.reject! { |handler| handler == Sidekiq.method(:default_error_handler) }
```

Perhaps we should add documentation around this in the wiki https://github.com/mperham/sidekiq/wiki/Error-Handling

Cheers

mperham:
Use `clear` instead of depending on an impl detail.

codebender:
Well I didn't want to clear what was already added by other initializers/gems. For example here is the list of error_handlers when the sidekiq.rb file is loaded:

```
#<Method: Sidekiq.default_error_handler(ex, ctx) /Users/matthewbender/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/sidekiq-6.5.1/lib/sidekiq.rb:81
#<Proc:0x0000000117c51ee0 /Users/matthewbender/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/newrelic_rpm-8.8.0/lib/new_relic/agent/instrumentation/sidekiq.rb:96
#<Proc:0x00000001171ab3d8 /Users/matthewbender/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/rollbar-3.3.0/lib/rollbar/plugins/sidekiq.rb:15
```

 If I `.clear`, I lose my rollbar and newrelic error handlers, so I sort of have to depend on that implementation detail 😿

mperham:
[...]

PR #5395, opened 2022-06-17
define DEFAULT_ERROR_HANDLER

codebender:
This will allow the default error handler to be easily removed if needed via `error_handlers.delete(DEFAULT_ERROR_HANDLER)` without having to depend on an internal implementation detail

I'm not sure where you might want this constant to be defined as it has to be after the method `self.default_error_handler` is defined.  So I thought right after the method might be okay or maybe at the end of the module?

Closes: https://github.com/mperham/sidekiq/issues/5391

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -84,6 +84,11 @@ module Sidekiq
     logger.warn(ex.backtrace.join("\n")) unless ex.backtrace.nil?
   end
 
+  # DEFAULT_ERROR_HANDLER is a constant that allows the default error handler to
+  # be referenced. It must be defined here, after the default_error_handler
+  # method is defined.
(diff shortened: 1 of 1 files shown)


mperham:
I’m on vacation so this will need to wait for a while.

codebender:
No rush, enjoy your vacation! 🏝️

codebender:
@mperham Hey Mr Perham, are you back from vacation?

mperham:
I'm on it, Mr. @codebender!

rparjun:
with Sidekiq 7, it should be  `Sidekiq.error_handlers.delete(Sidekiq::Config::ERROR_HANDLER)`. @mperham should this be mentioned in the [upgrade](https://github.com/sidekiq/sidekiq/blob/main/docs/7.0-Upgrade.md) notes?

PR #5398, opened 2022-06-20
Sort processes in UI

KevinBongart:
Hi @mperham and thanks for this essential tool!

I noticed that the UI sorts processes by comparing hostnames as strings and, because Heroku just adds a number after the process name (`worker.1`, `worker.2`…), it results in processes named like `worker.10` ranking above `worker.1`:

Screen Shot 2022-06-20 at 4 34 47 PM

This PR tries to address it without adding dependencies or too much unmaintainable cruft, so I thought a helper would contain this logic toward the frontend code.

The new helper `sorted_processes` sorts processes by hostname following the **natural sort order** so that

```rb
'worker.1' < 'worker.2' < 'worker.10' < 'worker.20'
'2.1.1.1' < '192.168.0.2' < '192.168.0.10'
```

Another approach would be to add or modify a `<=>` method in the `ProcessSet` class, whichever is preferable.

What do you think?

### DIFF
--- a/lib/sidekiq/web/helpers.rb
+++ b/lib/sidekiq/web/helpers.rb
@@ -153,6 +153,8 @@ module Sidekiq
     # '2.1.1.1' < '192.168.0.2' < '192.168.0.10'
     def sorted_processes
       @sorted_processes ||= begin
+        return processes unless processes.all? { |p| p['hostname'].present? }
+
         split_characters = /[\._-]/
 
         padding = processes.flat_map { |p| p['hostname'].split(split_characters) }.map(&:size).max
--- a/lib/sidekiq/web/helpers.rb
+++ b/lib/sidekiq/web/helpers.rb
@@ -148,6 +148,17 @@ module Sidekiq
       @processes ||= Sidekiq::ProcessSet.new
     end
 
+    # Sorts processes by hostname following the natural sort order so that
+    # 'worker.1' < 'worker.2' < 'worker.10' < 'worker.20'
+    def sorted_processes
+      @sorted_processes ||= processes.to_a.sort_by do |process|
+        words = process['hostname'].split('.')
+        padding = words.map(&:size).max
+
+        words.map { |word| word.rjust(padding, '0') }
+      end
+    end
+
     def stats
       @stats ||= Sidekiq::Stats.new
     end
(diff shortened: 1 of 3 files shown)


mperham:
I’m gonna say no as is. Too specialized to Heroku. Its too bad they didn’t zero pad the names.

KevinBongart:
@mperham I've fixed the algorithm and added some examples showing that, outside Heroku, the natural sorting can also be useful. For instance, hostnames named with IP addresses are currently sorted like this:

```rb
['182.168.2.1', '182.168.10.1', '19.0.0.1', '2.1.1.1'].sort
=> ["182.168.10.1", "182.168.2.1", "19.0.0.1", "2.1.1.1"]
```

But one would expect them to be sorted like this:

```rb
=> ["2.1.1.1", "19.0.0.1", "182.168.2.1", "182.168.10.1"]
```

I've also added `-` and `_` as splitting characters so that hostnames like `sidekiq-worker-1` and `sidekiq_worker_1` can be sorted naturally as well.

All in all, the implementation is getting a bit more complex than my first attempt and I also agree that this minor annoyance is, in my years of experience with Sidekiq, usually limited to Heroku process names. Feel free to close this PR, and maybe I'll extract this into a tiny extension for my own peace of mind 😅

mperham:
Well done, this looks really nice. The clean code really makes me want to merge.

mperham:
CI very very broken.

KevinBongart:
Thanks! I'll take a look at the failing tests

KevinBongart:
@mperham I added a guard to prevent sorting processes unless all processes have a hostname set: https://github.com/mperham/sidekiq/pull/5398/commits/d489e706f22951022324abe4aeb3096cdf600186

Locally, the test suite is green (or, I guess, rainbow-colored 😄):

```
Fabulous run in 5.334341s, 80.2348 runs/s, 244.6413 assertions/s.
428 runs, 1305 assertions, 0 failures, 0 errors, 0 skips
```

Let me know if that guard makes sense or if I should instead fix tests so that all mock processes have a hostname, whichever you prefer!

lib/sidekiq/web/helpers.rb
```diff
@@ -148,6 +148,29 @@ def processes
       @processes ||= Sidekiq::ProcessSet.new
     end
 
+    # Sorts processes by hostname following the natural sort order so that
+    # 'worker.1' < 'worker.2' < 'worker.10' < 'worker.20'
+    # '2.1.1.1' < '192.168.0.2' < '192.168.0.10'
+    def sorted_processes
+      @sorted_processes ||= begin
+        return processes unless processes.all? { |p| p['hostname'].present? }
```
mperham, line 156: Is `present?` an ActiveSupport thing? Can you verify your method runs in a bare IRB session?

PR #5412, opened 2022-07-03
Allow hiding web actions

uri-shmueli:
Default
image

Adding 
`Sidekiq.options[:should_hide_web_action]=["delete_queue"]` to the user's application level, will prevent the Delete action where needed.
This is most useful where we need local env with the Delete queue ability, but we want to prevent this from happening in production level.
image


mperham:
You should open an issue to discuss before opening functional PRs. This is
already an Enterprise feature, see Web UI Authorization wiki page.

rinslow:
LGTM :)

PR #5423, opened 2022-07-10
Allow to kill job if the exception is raised

### LINKED ISSUE #5406: retry count (not retry_in) based on error?
Ruby version:  3.1.2
Rails version: 6.1.5
Sidekiq / Pro / Enterprise version(s):  Sidekiq 6.4.1

Is there a way to customize the retry count within the worker class based on the error class that was raised? I know you can customize `sidekiq_retry_in`, but can I return false or something so it stops retrying and send directly to Deadset? The [help doc](https://github.com/mperham/sidekiq/wiki/Error-Handling) doesn't describe it.

Something like this:

```
  sidekiq_retry_in do |count, exception|
    case exception
    when SpecialException
      false
    end
  end
```

fatkodima:
Fixes #5406.

`discard` is used in rails - it means the job will be just like finished/succeeded - https://github.com/rails/rails/blob/81d6012f4ff24d071eb6831c0e9a310e779e98b4/activejob/lib/active_job/exceptions.rb#L94-L101

`stop` - sounds like it will be just stopped and resumed(?) later
`kill` - sounds the most appropriate to me, because the job is moved to dead set (basically it needs to be killed to become dead 😄)

cc @netwire88

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -2,6 +2,19 @@
 
 [Sidekiq Changes](https://github.com/mperham/sidekiq/blob/main/Changes.md) | [Sidekiq Pro Changes](https://github.com/mperham/sidekiq/blob/main/Pro-Changes.md) | [Sidekiq Enterprise Changes](https://github.com/mperham/sidekiq/blob/main/Ent-Changes.md)
 
+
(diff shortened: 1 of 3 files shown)


mperham:
I implemented it a bit differently in #5432, thank you anyways.

PR #5432, opened 2022-07-14
Enable `sidekiq_retry_in` to signal :kill or :discard dynamically

### LINKED ISSUE #5406: retry count (not retry_in) based on error?

mperham:
Allow `sidekiq_retry_in` to return either an Integer of seconds OR :kill OR :discard. The latter two symbols tell Sidekiq to send it directly to the morgue or throw the job away. Any other returned value will result in the default retry processing.

Fixes #5406

### DIFF
--- a/lib/sidekiq/job_retry.rb
+++ b/lib/sidekiq/job_retry.rb
@@ -91,7 +91,7 @@ module Sidekiq
 
       msg = Sidekiq.load_json(jobstr)
       if msg["retry"]
-        attempt_retry(nil, msg, queue, e)
+        process_retry(nil, msg, queue, e)
       else
         Sidekiq.death_handlers.each do |handler|
           handler.call(msg, e)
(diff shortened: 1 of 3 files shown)


filipesperandio:
Was this documented somewhere?

mperham:
Only in the changelog. It's quite easy for me to fail to update the wiki since I can't change the wiki until the commit is released, which can be days, weeks or even months after merging. Thanks for the reminder.

https://github.com/sidekiq/sidekiq/blob/main/Changes.md#652

filipesperandio:
@mperham I see it [here](https://github.com/sidekiq/sidekiq/wiki/Error-Handling#configuration) now! Thanks!

PR #5433, opened 2022-07-15
Make ProcessSet cleaning flag injectable.

fotinakis:
`clean_plz` doesn't seem to be used anywhere in the codebase. This provides a semi-private hook (similar to internal LINGER const) which can be injected if needed in special cases (described in support email).


fotinakis:
Closing this in preference for https://github.com/mperham/sidekiq/issues/5435

Issue #4850, opened 2021-03-24
Upgrading sidekiq: 6.1.3 → 6.2.0 causes `Sidekiq::Web needs a valid Rack session for CSRF protection.` in a rails app even though mounting sidekiq inside the routes


johannesengl:
Upgrading sidekiq: 6.1.3 → 6.2.0 (minor) #3599 causes `Sidekiq::Web needs a valid Rack session for CSRF protection.` in a rails app even though we are mounting sidekiq inside the routes like so: 

```
Rails.application.routes.draw do
mount Sidekiq::Web => "/sidekiq"
....
end 
```

Ruby version: '2.7.2'
Rails version: `6.1`
Sidekiq / Pro / Enterprise version(s): 6.2.0

Initializer:

```
require 'sidekiq'
require 'sidekiq/scheduler'
require 'sidekiq/web'
require 'sidekiq-scheduler/web'
require 'sidekiq/enqueuer'

Sidekiq::Web.use(Rack::Auth::Basic) do |user, password|
  [user, password] == [...]
end

if Rails.env.production?
  Sidekiq.configure_server do |config|
    config.redis = { ... }
  end

  Sidekiq.configure_client do |config|
    ......
  end
end

if ENV.fetch('RUN_SIDEKIQ_CRON').to_bool && Sidekiq.server?
  Sidekiq.configure_server do |config|
    config.on(:startup) do
      scheduler_file_content = File.read(Rails.root.join('config/sidekiq_scheduler.yml.erb'))

      Sidekiq.schedule = YAML.safe_load(
        ERB.new(scheduler_file_content).result
      )

      Sidekiq::Scheduler.reload_schedule!
    end
  end
else
  Sidekiq::Scheduler.enabled = false
end
```

Error: 

```
RuntimeError - Sidekiq::Web needs a valid Rack session for CSRF protection. If this is a Rails app,
make sure you mount Sidekiq::Web *inside* your application routes:
Rails.application.routes.draw do
  mount Sidekiq::Web => "/sidekiq"
  ....
end
If this is a bare Rack app, use a session middleware before Sidekiq::Web:
# first, use IRB to create a shared secret key for sessions and commit it
require 'securerandom'; File.open(".session.key", "w") {|f| f.write(SecureRandom.hex(32)) }
# now use the secret with a session cookie middleware
use Rack::Session::Cookie, secret: File.read(".session.key"), same_site: true, max_age: 86400
run Sidekiq::Web:
```

[...]

PR #5451, opened 2022-07-29
Add example how to configure Rails cookies store only for sidekiq web

artofhuman:
This PR adds example how to configure cookies store for sidekiq web if rails in api mode (for rack app we already have an example) It may reduce wasting time to find solution

### DIFF
--- a/lib/sidekiq/web/csrf_protection.rb
+++ b/lib/sidekiq/web/csrf_protection.rb
@@ -81,6 +81,11 @@ module Sidekiq
 
             https://guides.rubyonrails.org/api_app.html#using-session-middlewares
 
+            # example config sidekiq specific session
+            # https://github.com/mperham/sidekiq/wiki/Monitoring#rails-api-application-session-configuration
+            Sidekiq::Web.use ActionDispatch::Cookies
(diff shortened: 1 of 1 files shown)


lib/sidekiq/web/csrf_protection.rb
```diff
@@ -81,6 +81,11 @@ def session(env)
 
             https://guides.rubyonrails.org/api_app.html#using-session-middlewares
 
+            # example config sidekiq specific session
+            # https://github.com/mperham/sidekiq/wiki/Monitoring#rails-api-application-session-configuration
+            Sidekiq::Web.use ActionDispatch::Cookies
+            Sidekiq::Web.use ActionDispatch::Session::CookieStore, key: "_interslice_session"
```
mperham, line 87: What is `_interslice_session`? Where does this code go?

artofhuman: It's from link above https://guides.rubyonrails.org/api_app.html#using-session-middlewares and also i found this answer https://github.com/mperham/sidekiq/issues/4850#issuecomment-810880012

mperham:
I don't want to add anything more than the link to the rails guide. The instructions might be Rails version-specific.

Issue #5458, opened 2022-08-05
Improve metrics UI


mperham:
This is a general placeholder for any upcoming work on improving and polishing the Metrics UI.

adamlogic:
Hey @mperham, I'd love to help with this! I've spent a bit of time digging into the current implementation, and I want to summarize my current understanding.

**Data**

From what I can tell, the current metrics implementation gives us access to the following data:

- Processed count, failed count, and total exec time
	- For all jobs, and each individual job class
	- By the day, hour, and minute (retained for 90 days, 7 days, and 8 hours respectively)
	- Stored as hashes at each point in time
	- Processed count includes successful and failed jobs
	- Exec time does not include failed jobs
- Histogram exec times
	- For each individual job class
	- By the minute (retained for 24 hours)
	- Stored as BITFIELD for each combination of job class and point in time
	- Histogram is a collection of job counts broken up over 26 exec time buckets

**Tracking & collection**

The `ExecutionTracker` is injected as server middleware. For each job, it tracks the execution time and records the histogram value. This data is kept in memory.

On each "beat" (which I think runs every 5 seconds—I'm still a bit confused by the heartbeat), these in-memory metrics are flushed to Redis.

**JS lbraries**

Here are the tools/libraries I see currently in use by the Sidekiq UI:

- Charts: [D3](https://d3js.org) [3.5.16](https://github.com/d3/d3/releases/tag/v3.5.16) & [Rickshaw](https://tech.shutterstock.com/rickshaw/) [1.6.0](https://github.com/shutterstock/rickshaw/releases/tag/v1.6.0)
- JS behavior utilities: None

**Technical questions**

- Are you open to tracking other metrics? Personally I think latency per queue over time would be very useful.
- Are you open to other charting libraries? I use D3 on Rails Autoscale, and while it's super powerful, I'm not sure I'd recommended for open source. It's got a significant learning curve, and it's very easy to get out of date (v3.5.16 is over six years old). I don't have a ton of experience with other libraries, but I'd probably recommend [Chart.js](https://www.chartjs.org) for its popularity, ecosystem, docs, and beautiful charts.
- Are you open to lightweight JS behavior utilities like [Alpine](https://alpinejs.dev) or [Stimulus](https://stimulus.hotwired.dev)? Both of these operate on server-generated HTML—no client-side rendering. I'm a huge Alpine fan, but Stimulus is more common in Rubyland.

**Functional questions**

- Are there specific charts you have in mind?
[...]

PR #5467, opened 2022-08-10
[WIP] Improve metrics UI with charts

adamlogic:
This is a work-in-progress to address #5458. It introduces [Chart.js](https://www.chartjs.org/docs/latest/) to the new metrics page, plotting the total execution time of the 10 most time-consuming jobs on a line chart.

CleanShot 2022-08-10 at 16 37 06@2x


lib/sidekiq/metrics/query.rb
```diff
@@ -35,48 +35,32 @@ def initialize(pool: Sidekiq.redis_pool, now: Time.now)
...
-      # into top processed count and execution time based on class.
-      def top_jobs
-        resultset = {}
-        resultset[:date] = @time.to_date
-        resultset[:period] = :hour
-        resultset[:ends_at] = @time
-        time = @time
+      # Get metric data for all jobs from the last hour
+      def top_jobs(minutes: 60)
+        result = Result.new
```
adamlogic, line 40: I had a lot of trouble wrapping my head around the previous implementation and shaping the data the way I needed it for the chart. It helped me to extract the `Query::Result` and `Query::JobResult` classes here, but I'm sure we could also get it working with the nested Hashes if you don't like this approach.

lib/sidekiq/metrics/query.rb
```diff
@@ -35,48 +35,32 @@ def initialize(pool: Sidekiq.redis_pool, now: Time.now)
...
-        results = @pool.with do |conn|
+        time = @time
+        redis_results = @pool.with do |conn|
           conn.pipelined do |pipe|
-            resultset[:size] = 60
-            60.times do |idx|
+            minutes.times do |idx|
               key = "j|#{time.strftime("%Y%m%d")}|#{time.hour}:#{time.min}"
               pipe.hgetall key
+              result.prepend_bucket time
```
adamlogic, line 48: We need this running tally of all the "buckets" so we can provide the chart a complete x-axis, even for times when there's no job data.

test/test_metrics.rb
```diff
@@ -104,16 +99,19 @@ def create_known_metrics(time = fixed_time)
...
+      assert_equal fixed_time, result.ends_at
+
+      assert_equal 60, result.buckets.size
+      assert_equal "21:04", result.buckets.first
+      assert_equal "22:03", result.buckets.last
+
+      assert_equal %w[App::SomeJob App::FooJob].sort, result.job_results.keys.sort
+      some_job_result = result.job_results["App::SomeJob"]
+      refute_nil some_job_result
+      assert_equal 1, some_job_result.series.dig("ms", "22:03")
```
adamlogic, line 118: This test is randomly failing with different actual values, and I'm not sure why, since we're using a fixed time. It passes most of the time.

```
Failure:
Sidekiq::Metrics::querying#test_0002_fetches top job data [test/test_metrics.rb:118]:
Expected: 1
  Actual: 8

Failure:
Sidekiq::Metrics::querying#test_0002_fetches top job data [test/test_metrics.rb:118]:
Expected: 1
  Actual: 2
```

web/assets/javascripts/metrics.js
```diff
@@ -0,0 +1,41 @@
...
+  });
+  const options = {
+    aspectRatio: 4,
+    scales: {
+      y: {
+        beginAtZero: true,
+      },
+    },
+    interaction: {
+      mode: "x",
```
adamlogic, line 25: This configures the tooltip to show all points at the current spot on the x-axis, rather than just a single datapoint.

CleanShot 2022-08-10 at 16 47 53@2x

versus...

CleanShot 2022-08-10 at 16 48 25@2x

web/assets/javascripts/metrics.js
```diff
@@ -0,0 +1,41 @@
...
+    },
+    plugins: {
+      legend: {
+        display: false,
```
adamlogic, line 29: I opted the hide the default legend because I thought it was ugly. I think the swatches in the data table are just as effective and look a lot nicer.

CleanShot 2022-08-10 at 16 49 21@2x

We could also potentially make the data table swatches interactive if, for example, we only wanted to initially show the top 5 job classes in the chart. You could use the swatch to toggle a job class' visibility in the chart. 🤷‍♂️

web/assets/javascripts/metrics.js
```diff
@@ -0,0 +1,41 @@
...
+  };
+
+  const chart = new Chart(ctx, {
+    type: "line",
+    data: { labels: labels, datasets: datasets },
+    options: options,
+  });
+}
+
+document.addEventListener("DOMContentLoaded", init);
```
adamlogic, line 41: The dashboard uses this [custom `ready` function](https://github.com/adamlogic/sidekiq/blob/3ab19ed514c10510e4e403067fcd61e5f2b48164/web/assets/javascripts/dashboard.js#L224-L227), but I don't know if it's necessary.

web/views/metrics.erb
```diff
@@ -1,50 +1,58 @@
+<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/3.9.1/chart.min.js"></script>
+<script type="text/javascript" src="<%= root_path %>javascripts/metrics.js"></script>
 
-<h1><%= t('Metrics') %></h1>
+<h3><%= t('Metrics') %></h3>
```
adamlogic, line 4: Changed this heading level to match other pages in the web UI.

web/views/metrics.erb
```diff
@@ -1,50 +1,58 @@
...
-</div>
+<canvas
+  id="metrics-chart"
+  data-labels="<%= h Sidekiq.dump_json(@query_result.buckets) %>"
+  data-series="<%= h Sidekiq.dump_json(job_results.map { |(kls, jr)| [kls, jr.dig("series", "ms")] }.to_h) %>"
+  data-colors="<%= h Sidekiq.dump_json(colors) %>"
+></canvas>
 
-<h3>Top Jobs by Execution Time</h3>
+<h3>Most Time-Consuming Jobs</h3>
```
adamlogic, line 31: Renamed this in case we want to make this table sortable. It will also list the top X most time-consuming jobs, but they might not be _sorted by_ execution time (although they are right now).

adamlogic (commented):
Consider this a first draft. Happy to iterate on it as much as we need.

Some things I'm still thinking about...

- [x] Add deploy markers to the chart.
- [ ] Add the ability to change the metric shown in the chart (processed jobs, failed jobs, avg exec time)
- [ ] Show *more* job classes in the data table and *fewer* job classes in the chart. Right now the chart feels a bit too busy and the table a bit too anemic.

mperham:
🤩 Holy cow.

Maybe we could have checkboxes in the table rows which show/hide each job type within the graph? And have the graph only show the top 5 by default?

web/assets/javascripts/metrics.js
mperham: I don't know either. If it's not needed, it's a bug.

davidwessman:
Is it possible to add labels for the x- and y-axis? `Total execution time (s)` vs `Time (hh:mm)`

test/test_metrics.rb
davidwessman: Could it help to debug it by adding the whole hash as the message?

```suggestion
      assert_equal 1, some_job_result.series.dig("ms", "22:03"), some_job_result
```

adamlogic: Fixed in https://github.com/mperham/sidekiq/pull/5467/commits/d0b1616ad82250204cf1a270aac17a64000de37b.

adamlogic:
> Maybe we could have checkboxes in the table rows which show/hide each job type within the graph? And have the graph only show the top 5 by default?

https://github.com/mperham/sidekiq/pull/5467/commits/da81f58146fe4d58abe2fa25c47db9c68550fcdc converted the color swatches to checkboxes, and https://github.com/mperham/sidekiq/pull/5467/commits/da81f58146fe4d58abe2fa25c47db9c68550fcdc wired them up to toggle each job type within the graph.

The top 20 most time-consuming job types will be shown in the data table, but only 5 will initially be shown on the chart.

https://user-images.githubusercontent.com/372/184196628-da4e08a9-1562-45aa-b2e8-de2a029ab880.mp4

We don't have enough data here to trigger the `fallbackColor`, so for the demo below, I deleted all but three colors.


https://user-images.githubusercontent.com/372/184196765-53170c41-a020-4db6-b5b0-515128a1269c.mp4

adamlogic:
After posting those demos above, I realized the unchecked checkboxes looked disabled. That's fixed now.

CleanShot 2022-08-11 at 13 35 00@2x

web/views/metrics.erb
```diff
@@ -1,50 +1,60 @@
+<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/3.9.1/chart.min.js"></script>
```
mperham, line 1: We will need to vendor this. The Web UI should be totally self-contained.

mperham (commented):
Fantastic progress. Any luck on the vertical deploy marks?

web/views/metrics.erb
mperham: Make sure our copy of chart.js.min includes the MIT license in their repo.

mperham: Oh, looks like it already does include a small license header.

adamlogic:
UX improvements in https://github.com/mperham/sidekiq/pull/5467/commits/9f5a658f0e83f68d0cededfc09b9380dd80e37ff:

- The chart now displays seconds instead of milliseconds (matches the data table)
- Added a label for the x-axis
- Added rounding and labels to the tooltip

CleanShot 2022-08-11 at 14 04 04@2x

Note that I did not add a label for the x-axis because I didn't think the added noise provided enough value. The x-axis is obviously time, but not obviously in UTC, which is why I added "UTC" to the tooltip.

adamlogic:
> Any luck on the vertical deploy marks?

Up next!

adamlogic:
Deploy marks!

CleanShot 2022-08-11 at 16 34 28@2x

Those labels show the first "word" of the mark label, but the full label is shown in the tooltip.

CleanShot 2022-08-11 at 16 34 47@2x

Perhaps it's too much to assume the first word of the label is the deploy SHA. I'm wondering if we should lose the label and just keep the tooltip.

mperham:
That's a reasonable question. We could either add more smarts to the deploy mark by supporting an explicit `scm_label` which could be a git sha, git tag, svn id, etc. Or we could just create a convention where if the first word in the marker label is surrounded by `[word]` then we display `word`. Hard code "smarts" or rely on a convention?

adamlogic:
Bug: With the deploy annotations, the chart y-axis scale is no longer updating based on the selected job types. I'll figure it out tomorrow.

https://user-images.githubusercontent.com/372/184244551-7b06c3b6-b505-4e3e-8a72-1a1a3fdb96ce.mp4

web/views/metrics.erb
adamlogic: Scripts are now vendored.

adamlogic:
> Hard code "smarts" or rely on a convention?

I don't have a strong opinion here, but if it were up to me I'd probably say "neither", and remove the label annotation altogether. It feels like we're trying to assume too much about how this feature is going to be used.

Also, the label annotation required an [ugly hack](https://github.com/mperham/sidekiq/pull/5467/commits/8a8fe3f51aed047e1767c9581c8eab74f365575b) to get the positioning working correctly. The vertical bar didn't require the same hack, so without the label I could remove that hack.

adamlogic:
Here's a quick demo of the UX *without* the label.

CleanShot 2022-08-12 at 08 47 42

mperham:
Well, an unlabeled line is harder to answer "what is this? why is it here?" Maybe we could have a separate legend or table for deploy marks.

adamlogic:
Great! I've removed the label annotations, which really cleaned things up.

How are you feeling about the scope of this PR? I'm not sure there's more I want to do on the metrics overview page, for now at least. I'd rather move on to the job metrics page, and I can either continue with this PR or start a new one.

mperham:
I love it as is, I'll test it out and merge if all is well. Open up a new PR for the job specific page.

mperham:
If you want extra credit, test it with dark mode and using Hebrew as your browser language (to get RTL UI flow).

mperham:
Lots of room for polish but infinitely better than what was there before.
Screen Shot 2022-08-12 at 9 51 06 AM

PR #5469, opened 2022-08-11
Remove Sidekiq Logger#add patch

the-spectator:
ruby 2.7.0 now honors Logger#level overrides, and since sidekiq 7 will require ruby version >= 2.7, we can remove the patch.

Ref:
https://github.com/ruby/ruby/commit/eb18cb3e476db3bc44d489e090e1535237c4c6c9
https://github.com/rails/rails/pull/44247


mperham:
Thank you!

the-spectator:
Thank you for the merge! 🙏 🎉

PR #5473, opened 2022-08-12
Histogram chart for job-specific metrics

adamlogic:
Work-in-progress while I explore some possibilities.

### DIFF
--- a/lib/sidekiq/metrics/query.rb
+++ b/lib/sidekiq/metrics/query.rb
@@ -72,45 +72,42 @@ module Sidekiq
         result
       end
 
-      def for_job(klass)
-        resultset = {}
-        resultset[:date] = @time.to_date
-        resultset[:period] = :hour
-        resultset[:ends_at] = @time
-        marks = @pool.with { |c| c.hgetall("#{@time.strftime("%Y%m%d")}-marks") }
+      def for_job(klass, minutes: 60)
+        result = Result.new
 
         time = @time
-        initial = @pool.with do |conn|
+        redis_results = @pool.with do |conn|
           conn.pipelined do |pipe|
-            resultset[:size] = 60
-            60.times do |idx|
-              key = "j|#{time.strftime("%Y%m%d|%-H:%-M")}"
+            minutes.times do |idx|
+              key = "j|#{time.strftime("%Y%m%d")}|#{time.hour}:#{time.min}"
               pipe.hmget key, "#{klass}|ms", "#{klass}|p", "#{klass}|f"
+              result.prepend_bucket time
               time -= 60
             end
           end
         end
 
         time = @time
-        hist = Histogram.new(klass)
-        results = @pool.with do |conn|
-          initial.map do |(ms, p, f)|
-            tm = Time.utc(time.year, time.month, time.mday, time.hour, time.min, 0)
-            {
-              time: tm.iso8601,
-              epoch: tm.to_i,
-              ms: ms.to_i, p: p.to_i, f: f.to_i, hist: hist.fetch(conn, time)
-            }.tap { |x|
-              x[:mark] = marks[x[:time]] if marks[x[:time]]
-              time -= 60
-            }
+        @pool.with do |conn|
+          redis_results.each do |(ms, p, f)|
+            result.job_results[klass].add_metric "ms", time, ms.to_i if ms
+            result.job_results[klass].add_metric "p", time, p.to_i if p
+            result.job_results[klass].add_metric "f", time, f.to_i if f
+            result.job_results[klass].add_hist time, Histogram.new(klass).fetch(conn, time)
+            time -= 60
           end
         end
 
-        resultset[:marks] = marks
-        resultset[:starts_at] = time
-        resultset[:data] = results
-        resultset
+        marks = @pool.with { |c| c.hgetall("#{@time.strftime("%Y%m%d")}-marks") }
+        result_range = result.starts_at..result.ends_at
+        marks.each do |timestamp, label|
+          time = Time.parse(timestamp)
+          if result_range.cover? time
+            result.marks << MarkResult.new(time, label)
+          end
+        end
+
+        result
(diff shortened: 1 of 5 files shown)


adamlogic:
Here's my first exploration here - using a bubble chart to visualize the histogram data for a job.

CleanShot 2022-08-12 at 15 36 55@2x

CleanShot 2022-08-12 at 15 36 49@2x

Curious for feedback before I move forward.

adamlogic:
The other thing I was going to check out was this [box plot plugin](https://github.com/sgratzl/chartjs-chart-boxplot) for Chart.js. LMK if you think I should go down that route.

mperham:
I’m not sure about the y axis. I’m not clear how to identify it as logarithmic, not linear.

the bubbles have a lot of visual noise, maybe the box plot will be cleaner? 🤷🏻‍♂️

adamlogic:
Here's an attempt at a box plot chart. Y-axis is milliseconds.

CleanShot 2022-08-13 at 08 39 30@2x

CleanShot 2022-08-13 at 08 39 18@2x

And here's another version of the bubble chart with a linear y-axis (same y-axis as the box plot above).

CleanShot 2022-08-13 at 08 40 42@2x

CleanShot 2022-08-13 at 08 40 30@2x

I personally find the logarithmic axis more readable and a better visual aesthetic, and I think the bubbles convey more information than the box plot. My favorite amongst these is the first version (bubbles with logarithmic axis), but I'll roll with the one you prefer. Or we can keep exploring if you have more ideas.

adamlogic:
I deployed this branch to our staging environment for Rails Autoscale to get some screenshots of real data.

CleanShot 2022-08-13 at 08 54 59@2x

CleanShot 2022-08-13 at 08 55 15@2x

CleanShot 2022-08-13 at 08 55 32@2x

adamlogic:
And just for fun, here's the metrics overview page.

CleanShot 2022-08-13 at 08 59 00@2x

mperham:
The box plot is too visually “heavy” with the box filled in. And if the box is just outlined, that would make the dot in the middle pop much better. I do like the bubbles though.

adamlogic:
Here it is with the box outlined. Note that these examples look a lot different because we have some extreme outliers now—these happened by accident when I closed my laptop for a bit while jobs were running.

CleanShot 2022-08-13 at 16 12 16@2x

CleanShot 2022-08-13 at 16 12 37@2x

I will say that the bubble chart with the linear scale does a much better job of showing just how far these outliers are from the rest, which I think is helpful. The bubbles at the bottom are all muddied together, but that's not the interesting data in the chart.

The linear scale also aligns with the scale we're using on the metrics overview chart (especially once I change the metric from milliseconds to seconds and add the y-axis label). I think there's benefit to having consistency between the two pages.

I'm definitely still pro-bubbles, but I'm starting to lean toward the linear scale vs. logarithmic.

mperham:
I'm ok with letting you drive the design. If you think bubbles work best, ship it. Others can comment or propose improvements over time.

adamlogic:
> I'm ok with letting you drive the design.

Famous last words. 😂 

I'm happy with the current state of this PR, so I'm marking it ready for review. Here's where I landed:

CleanShot 2022-08-15 at 08 21 16@2x

CleanShot 2022-08-15 at 08 23 12@2x

CleanShot 2022-08-15 at 08 22 30@2x

mperham:
Looks nice so far, I'll test it soon. Be sure to check dark mode.

mperham:
Total and Average Execution Time should probably have three digits, i.e. "421 sec" or "4.21 sec", so the math is clear, otherwise you get strange calculations like Processed = 7, TET = 2, Average = 0.42 (???)

adamlogic:
I just changed the rounding so all of the seconds measurements on these pages are rounded to two decimals for consistency. Also, here's dark mode:

CleanShot 2022-08-15 at 15 10 33@2x

mperham:
Nice. Can you verify the page looks good with no data and I'll merge?

mperham:
What do you think about two level headers:

h1 Metrics / OrderJunkJob

h2 Histogram Summary

h2 Performance over Time

Where the h1 is providing breadcrumbs and Metrics can link to the previous page (i.e. metrics home).

mperham:
What do we do with the data table? Doesn't seem super useful, especially if it shows rows with all zeros.

adamlogic:
> What do you think about two level headers

👍 I agree. I don't think there's any precedent for subheaders in the UI, so I'll do my best to make it work visually.

> What do we do with the data table? Doesn't seem super useful, especially if it shows rows with all zeros.

I vote to keep it. The data in the table is not reflected in the charts above, so it _could_ be helpful to dig into specifics at a given point in time.

Not a strong opinion, though. Happy to drop it if that's your preference.

CleanShot 2022-08-16 at 12 36 17@2x

adamlogic:
Here are the new headers for the metrics pages. Showing the dashboard as well to see that the header style matches (even though all the other pages use an `h3` instead of `h1`).

https://user-images.githubusercontent.com/372/184937807-0b93d945-dca2-4fc8-86a0-4f2de050f63b.mp4

And here's how it looks with no data at all.

CleanShot 2022-08-16 at 12 54 36@2x

mperham:
I don't think you need the Metrics header on the Metrics home page. It's useful on the Metrics detail page to link back to the Metrics home. It's not a bad idea for all detail pages: Queues > default, Batches > $BID, Retries > $JID, Limits > $NAME, etc. Let's avoid scope creep and make that a TODO in the future for now.

mperham:
If the data is completely empty, I think you should display a separate "No data found" message like on the other existing pages. For instance, retries or scheduled data.

adamlogic:
Removed the metrics header:

CleanShot 2022-08-16 at 13 53 05@2x

Added "no job metrics" message:

CleanShot 2022-08-16 at 13 53 22@2x

PR #5478, opened 2022-08-17
Re-introduce JobLogger#job_hash_context

pcreux:
What
----

Re-introduce `JobLogger#job_hash_context` that was [added in `5.0.1`](https://github.com/mperham/sidekiq/blob/main/Changes.md#501), then silently removed in `6.4.1`: https://github.com/mperham/sidekiq/commit/01686c5e25dd83ab89c6d85b36feec26e203e81b.

Why
----

It's much easier to add extra context to logs by overwriting `job_hash_context` than [`prepare`](https://github.com/mperham/sidekiq/blob/0b3751bf29088cb703d40787bb52f0cb6aa6e74f/lib/sidekiq/job_logger.rb#L24).

Example:

```ruby
  config[:job_logger] = Sidekiq::AddedContextLogger
end

class Sidekiq::AddedContextLogger < Sidekiq::JobLogger
  def job_hash_context(job_hash)
    context = super
    
    context[:organization] = job_hash["tenant"]

    context
  end
end
```

### DIFF
--- a/lib/sidekiq/job_logger.rb
+++ b/lib/sidekiq/job_logger.rb
@@ -24,14 +24,7 @@ module Sidekiq
     def prepare(job_hash, &block)
       # If we're using a wrapper class, like ActiveJob, use the "wrapped"
       # attribute to expose the underlying thing.
-      h = {
-        class: job_hash["display_class"] || job_hash["wrapped"] || job_hash["class"],
-        jid: job_hash["jid"]
-      }
-      h[:bid] = job_hash["bid"] if job_hash.has_key?("bid")
-      h[:tags] = job_hash["tags"] if job_hash.has_key?("tags")
-
-      Thread.current[:sidekiq_context] = h
+      Thread.current[:sidekiq_context] = job_hash_context(job_hash)
       level = job_hash["log_level"]
       if level
         @logger.log_at(level, &block)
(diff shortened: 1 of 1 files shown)


mperham:
Please open an issue/discussion to ask how to do something. You need to create a server middleware which adds the context:

```ruby
class LogTenant
  include Sidekiq::ServerMiddleware
  def call(instance, job_hash, queue, &block)
    Sidekiq::Context.with(organization: job["tenant"], &block)
  end
end

  config.server_middleware do |chain|
    chain.add LogTenant
  end
end
```

pcreux:
Thank you for your reply @mperham.

I tried what you're suggesting above with `6.5.5` and `main` without any success.

Looking at the code, it seems that middlewares run within the logger, not around, so the context is updated after the logger logs the "start" message and it's being reset before the logger logs the "done" message.

It looks something like this:

```
Processor#dispatch
  -> job_logger.prepare # set `Sidekiq::Context.current` to job context `{ class: "...Job", jid: "59c..." }`
    -> job_logger.call
      # log "start" with job context `{ class: "...Job", jid: "59c..." }`
      -> Middleware::chain.invoke
        -> LogTenant#call
          -> Sidekiq::Context.with({tenant: "customer1"})
            -> perform
      # reset `Sidekiq::Context.current`
      # log "done" with job context `{ class: "...Job", jid: "59c..." }`
```

Am I missing something?

mperham:
My initial thought is that's a bug. start/done should be logged around Job#perform and include any context set up by middleware.

makaroni4:
@pcreux I had the same problem, here's the working solution from Sidekiq 6.5.8 without a middleware or `job_hash_context`:

```ruby
module LoggingWithArguments
  def call(item, queue)
    Sidekiq::Context.add("queue", queue)
    Sidekiq::Context.add("args", item["args"])

    super
  end
end

  Sidekiq::JobLogger.prepend(LoggingWithArguments)

  config.logger.formatter = Sidekiq::Logger::Formatters::JSON.new
end
```

Hope that helps 🤞

PR #5482, opened 2022-08-19
Toggle which metric is shown in the job metrics overview

adamlogic:
This PR wires up the column headers so users can choose which metric to view in the charts. Clicking a column header will also sort the table (descending) based on the selected metric.

https://user-images.githubusercontent.com/372/185694543-7a82ffe3-5945-4134-8a16-fabfbc723203.mp4

Dark mode:

CleanShot 2022-08-19 at 15 37 31@2x

### DIFF
--- a/web/views/metrics.erb
+++ b/web/views/metrics.erb
@@ -37,7 +37,7 @@
 
 <div class="table_container">
   <table class="table table-bordered table-striped table-hover">
-    <tbody>
+    <thead>
       <tr>
         <th><%= t('Name') %></th>
         <th>
@@ -58,6 +58,8 @@
           }
         </script>
       </tr>
+    </thead>
+    <tbody>
       <% if job_results.any? %>
         <% job_results.each_with_index do |(kls, jr), i| %>
           <tr>
--- a/web/views/metrics.erb
+++ b/web/views/metrics.erb
@@ -23,6 +23,18 @@
         f: job_results.map { |(kls, jr)| [kls, jr.dig("series", "f")] }.to_h,
         avg: job_results.map { |(kls, jr)| [kls, jr.series_avg("s") ] }.to_h,
       },
+      metricLabels: {
+        s: "Total execution time (sec)",
+        p: "Processed jobs",
+        f: "Failed jobs",
+        avg: "Avg execution time (sec)",
+      },
+      metricUnits: {
+        s: "seconds",
+        p: "jobs",
+        f: "jobs",
+        avg: "seconds",
+      },
       marks: @query_result.marks.map { |m| [m.bucket, m.label] },
       visible: visible_kls,
       labels: @query_result.buckets,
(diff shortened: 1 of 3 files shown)


web/views/metrics.erb
```diff
@@ -2,7 +2,7 @@
 <script type="text/javascript" src="<%= root_path %>javascripts/chartjs-plugin-annotation.min.js"></script>
 <script type="text/javascript" src="<%= root_path %>javascripts/metrics.js"></script>
 
-<h2>Total execution time</h2>
+<h2>Metrics</h2>
```
adamlogic, line 5: Note that I changed the header to something generic, since the actual metric can change. I didn't think it was worth the extra complexity to wire this up via JS.

adamlogic:
Still a draft. Just realized the y-axis units/label don't change. 🤔

adamlogic:
Labels/units are fixed. Still need to fix deploy marks.

mperham:
As soon as this is ready, I'll merge and ship 6.5.5.

adamlogic:
This one's ready. 🚀 

https://user-images.githubusercontent.com/372/185746260-7c182f82-3439-4e49-8b2f-637bf6581ca2.mp4

mperham:
I'm a little confused. Typically clicking on a header will sort by that column. But here there's no sorting but instead we are selecting which data we see in the graph? That's not immediately intuitive. What about the second table below the first? Is it becoming redundant?

mperham:
Ah, the second table has disappeared.

adamlogic:
> But here there's no sorting but instead we are selecting which data we see in the graph?

It's doing both—clicking a header sorts the table by that column and shows it in the graph.

mperham:
Ok.

Another thing I noticed is that colors can jump between rows. I wonder if a color should always be associated with a given job type, no matter how we sort or which rows we de-select. Otherwise changing colors can be confusing.

adamlogic:
Can you show me an example? Sorting should _not_ change the colors associated to job types, but the colors _will_ shift rows because the rows themselves are moving.

adamlogic:
@mperham I spent a bit of time this weekend trying to rebase this PR onto 7-0 so I can make the changes we discussed, but I can't get metrics tracking working for `myapp` on 7-0. From what I can tell the `ExecutionTracker` middleware isn't being run.

I tried debugging for a while, but the new Capsule stuff is hard for me to follow—specifically with middleware. `Config` and `Capsule` both have their own middleware chains, and I'm not sure when each of them are being used.

There also appears to be an extra (duplicate?) [root-level `config.rb`](https://github.com/mperham/sidekiq/blob/7-0/config.rb) that I don't think is used at all.

LMK if you're able to get metrics tracking working for `myapp` in the 7-0 branch, and I can dive back in.

mperham:
Thanks for pointing that out. I will get it fixed first thing Monday.

mperham:
Ok, it should work now.

PR #5487, opened 2022-08-23
Introduce Sidekiq::Capsule

mperham:
`Sidekiq::Capsule` is a new abstraction in Sidekiq 7.0 which allows different queues to be processed with different concurrencies. See `docs/capsule.md` for an overview.

This is a major, major internal refactoring and will likely break lots of things. Feedback and bug reports very welcome.


mperham:
And yeah, there's a test suite refactoring in the middle which makes this harder to read than necessary. Apologies.

PR #5489, opened 2022-08-24
Redis 5.0.0.beta4 compatibility

casperisfine:
Fix: https://github.com/redis/redis-rb/issues/1135

All but one changes are purely test things.

The only actual change is:

```ruby
opts[:driver] ||= Redis::Connection.drivers.last || "ruby"
```

Which is unclear to me why it's even there since it has been the redis-rb default behavior for over a decade:
https://github.com/redis/redis-rb/blame/4.x/lib/redis/client.rb#L511

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         ruby: ["2.5", "2.6", "2.7", "3.0", "3.1"]
+        redis-rb: ["< 5", "edge"]
     services:
       redis:
         image: redis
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -18,6 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         ruby: ["2.5", "2.6", "2.7", "3.0", "3.1"]
+        redis-rb: ["< 5", "edge"]
```
casperisfine, line 21: NB: I added this to make sure it doesn't only work on my machine, but it's up to you wether or not you want this in your CI.

mperham:
I don't like the added version logic complexity. I think I will lock Sidekiq 6.x to Redis-rb < 5 and move forward with redis-client in Sidekiq 7. That ok with you?

casperisfine:
I mean, I added the version check just because, but honestly IMHO these tests should just be deleted because they are asserting private redis-rb APIs.

> That ok with you?

Up to you. I only submitted this as courtesy. The only annoying part is that it may hinder Redis 5.0 adoption, but assuming people upgrade to sidekiq 7.x, it's probably fine.

PR #5499, opened 2022-08-30
Get max retry job count from job class

rwstauner:
The [job class has a `retry` option](https://github.com/rwstauner/sidekiq/blob/rwstauner/job-retry-count/lib/sidekiq/worker.rb#L343
) that can be an integer,
but the [default job options](https://github.com/rwstauner/sidekiq/blob/rwstauner/job-retry-count/lib/sidekiq.rb#L237,L239
) set `retry` to `true`
and the retry logic is currently only checking the message,
so when a worker specifies a `retry` integer, it is being ignored and the max (25) is used.

### DIFF
--- a/lib/sidekiq/job_retry.rb
+++ b/lib/sidekiq/job_retry.rb
@@ -138,7 +138,12 @@ module Sidekiq
     # instantiate the job instance.  All access must be guarded and
     # best effort.
     def process_retry(jobinst, msg, queue, exception)
-      max_retry_attempts = retry_attempts_from(msg["retry"], @max_retries)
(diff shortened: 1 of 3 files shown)


rwstauner:
thanks for the quick reply!
i have a job right now that is retrying a lot because the job has `retry: true` when the job class says `retry: 1`.
I'll have to dig deeper tomorrow.

mperham:
Whatever is creating the job is not seeing the latest version of your code.

rwstauner:
:facepalm: Thank you, you're right, I restarted the server but not the console I was using.
It's been years since I dug through the Sidekiq code.  I appreciate the explanation.
Sorry for the noise.

PR #5501, opened 2022-08-30
Enable to define at argument while pushing in bulk with #perform_bulk

aovertus:
Provide an iteration on `#perform_bulk` implementation to allow to use `at` argument as a param

Our current project end up using `Sidekiq::Client.push_bulk` directly as we need to be able to define `at`

```
  def self.perform_bulk_in(interval, items)
    at = interval.seconds.from_now.to_f
    Sidekiq::Client.push_bulk("class" => self, "args" => items, "at" => at)
  end
```

### DIFF
--- a/lib/sidekiq/worker.rb
+++ b/lib/sidekiq/worker.rb
@@ -236,10 +236,12 @@ module Sidekiq
       end
       alias_method :perform_sync, :perform_inline
 
-      def perform_bulk(args, batch_size: 1_000)
+      def perform_bulk(args, batch_size: 1_000, at: nil)
         client = @klass.build_client
         result = args.each_slice(batch_size).flat_map do |slice|
-          client.push_bulk(@opts.merge("class" => @klass, "args" => slice))
+          items = @opts.merge("class" => @klass, "args" => slice)
+          items = items.merge("at" => at) if at.present?
+          client.push_bulk(items)
         end
 
         result.is_a?(Enumerator::Lazy) ? result.force : result
--- a/test/test_client.rb
+++ b/test/test_client.rb
@@ -331,6 +331,13 @@ describe Sidekiq::Client do
         assert_equal 1_001, jids.size
       end
 
+      it "pushes a large set of jobs with specified 'at' argument" do
+        time = Time.new(2019, 1, 1)
+        jids = MyWorker.perform_bulk((1..1_001).to_a.map { |x| Array(x) }, at: time.to_i)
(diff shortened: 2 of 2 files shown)


mperham:
I don't want to enable this usecase as it leads to large clumps of jobs all scheduled at the exact same time. Needs jitter.

nateberkopec:
I tend not to use scheduled jobs in the first place, but I too think there are probably better solutions for your usecase than bulk scheduling in the future.

nateberkopec:
One easy idea - write a new job class that just runs:

```
Sidekiq::Client.push_bulk("class" => self, "args" => items)
```

... and schedule _that_ job in the future.

Sidekiq's scheduled set is expensive to work with, you don't want to be adding massive amounts of things to it and making it really slow. So just add one job to it and let that job enqueue all the other jobs when you need them.

aovertus:
Hi there! Thanks for the feedback

@mperham 

our complete implementation is the following:

```
  def self.perform_bulk_with_jitter(items, batch_size: 1000, min_wait: 0.seconds, max_wait: 5.minutes)
    items.each_slice(batch_size).flat_map do |slice|
      random_wait = rand(min_wait...max_wait)
      perform_bulk_in(random_wait, slice)
    end
  end

  def self.perform_bulk_in(interval, items)
    at = interval.seconds.from_now.to_f
    Sidekiq::Client.push_bulk("class" => self, "args" => items, "at" => at)
  end
```

My assumption was that we would keep the sidekiq layer as abstract as possible. Should the jitter strategy be included in the sidekiq method itself ?

aovertus:
> One easy idea - write a new job class that just runs:
> 
> ```
> Sidekiq::Client.push_bulk("class" => self, "args" => items)
> ```
> 
> ... and schedule _that_ job in the future.
> 
> Sidekiq's scheduled set is expensive to work with, you don't want to be adding massive amounts of things to it and making it really slow. So just add one job to it and let that job enqueue all the other jobs when you need them.

@nateberkopec so the more the scheduled set is cluttered the slower is fetching/writing to it ? The ideal solution would be a recurring job fetching in batches then ? 

Our use case is that we have thousands of api calls to make on separate objects, the idea was to bulk jitter to respect the api rate limit.

mperham:
Unfortunately the bulk APIs really are turning into a bikeshed. I'm going to decline. Build your own APIs, create your own public gem if you'd like, but I don't want a dozen minor variations in Sidekiq core.

PR #5513, opened 2022-09-07
Fix poll interval not changing based on process count

engwan:
We should not memoize the value of `scaled_poll_interval` because this changes based on the number of Sidekiq processes.

Previously, at every poll, we check the process count to determine the formula for computing the random interval. But the `poll_interval_average` was still using a value computed from the first poll.

### DIFF
--- a/lib/sidekiq/scheduled.rb
+++ b/lib/sidekiq/scheduled.rb
@@ -147,13 +147,16 @@ module Sidekiq
         # As we run more processes, the scheduling interval average will approach an even spread
         # between 0 and poll interval so we don't need this artifical boost.
         #
-        if process_count < 10
+        count = process_count
+        interval = poll_interval_average(count)
+
+        if count < 10
           # For small clusters, calculate a random interval that is ±50% the desired average.
-          poll_interval_average * rand + poll_interval_average.to_f / 2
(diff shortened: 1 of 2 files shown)


lib/sidekiq/scheduled.rb
```diff
@@ -147,13 +147,16 @@ def random_poll_interval
         # As we run more processes, the scheduling interval average will approach an even spread
         # between 0 and poll interval so we don't need this artifical boost.
         #
-        if process_count < 10
+        count = process_count
+        interval = poll_interval_average(count)
```
engwan, line 151: I added the process count argument so we don't have to do `scard` multiple times in a single poll.

mperham:
Thanks, that's an excellent point about cleanup. Maybe we should call `Sidekiq::ProcessSet.new.size` on startup to keep it clean?

engwan:
> Thanks, that's an excellent point about cleanup. Maybe we should call `Sidekiq::ProcessSet.new.size` on startup to keep it clean?

Yes, I think running once on startup should be good enough.

mperham:
done https://github.com/mperham/sidekiq/commit/f3e0640a79fec2ea9a87bdfab12e7c45b4153778

PR #5483, opened 2022-08-20
Rebuild dashboard charts using Chart.js

adamlogic:
This PR is built upon #5482. I'll clean this up once that's merged. I just wanted to get a draft going so you know what I'm up to.

So far only the history graph is using Chart.js. Here's what I plan to tackle before this is ready for review:

- [x] Rebuild history chart using Chart.js
- [ ] Rebuild realtime chart using Chart.js
- [ ] Remove Rickshaw/D3
- [ ] Test dark mode
- [ ] Test responsiveness
- [ ] Test RTL

## New history chart

https://user-images.githubusercontent.com/372/185748219-70b3dc79-baef-4793-bc36-c352214ff1e6.mp4

## Old history chart

https://user-images.githubusercontent.com/372/185748319-ec798cca-64d3-490f-8c9c-d044c8319f27.mp4

PR #5523, opened 2022-09-13
Rebuild dashboard charts using Chart.js

adamlogic:
Second attempt at this PR, replacing #5483.

The new dashboard charts work and look mostly the same as the old dashboard charts, but they're rewritten to use Chart.js since that's the library we're using for the new metrics pages.

Things I still want to do in a follow-up PR before 7.0 drops:

- [ ] Show a "sticky" vertical line on the real-time chart that corresponds to the legend/tooltip below the chart.
- [ ] Internationalize chart axis labels (currently hardcoded in JS).

I'm not so sure about the tooltip behavior being different in the real-time chart vs. the history and metrics charts. The latter all use the built-in Chart.js tooltip, but that didn't work well for the real-time chart since the data under the cursor is moving.

See demos below.

### Live updating and tooltips

https://user-images.githubusercontent.com/372/190145998-345d9cef-2403-4e8e-8d3b-5cfbb23189d7.mp4

### History charts

https://user-images.githubusercontent.com/372/190146242-da04dba6-9722-4a6d-a844-50e68e6cf35e.mp4

### Dark mode

CleanShot 2022-09-14 at 07 51 32@2x

### RTL

CleanShot 2022-09-14 at 07 52 11@2x

### Responsive

https://user-images.githubusercontent.com/372/190147004-d8811917-99ca-46d0-8e1c-7cfb9ef10502.mp4


davidwessman (commented):
Looks very clean to inherit from the BaseChart!

adamlogic:
@mperham @davidwessman I have a couple things I still want to implement (see the issue description), but I don't want to delay this PR anymore. It'll be a few days at least before I'm able to dive back in.

PR #5530, opened 2022-09-17
render a "cursor" that follows the tooltip

adamlogic:
This is not yet working.

The intention is to have a vertical line on the chart corresponding to the currently shown "tooltip" info in the legend below the chart. This is the behavior of Sidekiq 6.

### DIFF
--- a/web/assets/javascripts/base-charts.js
+++ b/web/assets/javascripts/base-charts.js
@@ -65,7 +65,9 @@ class BaseChart {
   get chartOptions() {
     let chartOptions = {
       interaction: {
-        mode: "x",
+        mode: "nearest",
+        axis: "x",
+        intersect: false,
       },
       scales: {
         x: {
--- a/web/assets/javascripts/dashboard-charts.js
+++ b/web/assets/javascripts/dashboard-charts.js
@@ -114,7 +114,7 @@ class RealtimeChart extends DashboardChart {
         <span class="swatch" style="background-color: ${dp[1].dataset.borderColor};"></span>
         <span>${dp[1].dataset.label}: ${dp[1].formattedValue}</span>
       </span>
-      <span>${dp[0].label}</span>
+      <span class="time">${dp[0].label}</span>
     `;
   }
 
--- a/web/assets/javascripts/dashboard-charts.js
+++ b/web/assets/javascripts/dashboard-charts.js
@@ -119,8 +119,10 @@ class RealtimeChart extends DashboardChart {
   }
 
   renderCursor(dp) {
-    const cursor = this.chart.options.plugins.annotation.annotations.cursor
-    cursor.xMin = cursor.xMax = dp[0].label;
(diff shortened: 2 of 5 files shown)


adamlogic:
I wanted to post my work here, but I'm ready to throw in the towel. The [chartjs-annotation-plugin](https://www.chartjs.org/chartjs-plugin-annotation/master/) isn't great to work with, and I can't figure out how to update the position of the vertical line as the tooltip updates.

If anybody wants to try to pick this up and get it working, have at it.

mperham:
Would it be useful or reasonable to contact the maintainers and ask their advice? Maybe it’s a feature they would be interested in supporting.

adamlogic:
Just needed to sleep on it, apparently. It's helpful to actually _include_ the chartjs-plugin-annotation library on the dashboard page. 🙄 

Anyway, this is working now. Here's a demo.


https://user-images.githubusercontent.com/372/190903872-347339a4-ed2d-4fec-9787-6a5dd789a3e6.mp4

mperham:
Looks pretty good to me. I wonder if the bottom legend should be right aligned because the cursor will usually be on the right side of the graph. Otherwise your eyes jump from the right side cursor to the left side text. Confirm/deny?

adamlogic:
I think you're right. It feels much more natural to have the legend under the right side of the chart.

CleanShot 2022-09-20 at 12 00 17

Note that I also moved the time to the right of the metrics so that it doesn't jump around as the cursor moves. Here's the "jumping" before I made that change:

CleanShot 2022-09-20 at 11 58 26

mperham:
The values are still jumping and quite distracting. It’d be nice to use min-width or similar to minimize jumping. On Sep 20, 2022, at 09:05, Adam McCrea ***@***.***> wrote:﻿
I think you're right. It feels much more natural to have the legend under the right side of the chart.

Note that I also moved the time to the right of the metrics so that it doesn't jump around as the cursor moves. Here's the "jumping" before I made that change:

adamlogic:
Since the width of the actual content changes as the values change, I think we have to choose:

1. Items shifting as widths change.
2. Inconsistent margins between items.

I tried applying a min-width to the items in the legend, and the result is an inconsistent margin that IMO looks worse than the jumping. The demo below was a local experiment, and is not part of this PR.

https://user-images.githubusercontent.com/372/191807082-25173025-a474-4a04-bcf0-087d74e91a45.mp4

Open to more ideas, but I think I'd call it good enough to ship. 🤷‍♂️

mperham:
That demo video ^^^^^ looks perfect. You don't like it?

adamlogic:
I didn't love the uneven spacing between the elements of the legend, but I didn't feel strongly either way. I tweaked it a bit, and I'm happier with this:


https://user-images.githubusercontent.com/372/191820716-20e2a07e-1517-4e94-93a6-168a2872e749.mp4


I pushed that change to the PR.

PR #5538, opened 2022-09-21
WIP: Refreshing metrics

davidwessman:
- First draft of refreshing Metrics graph via JSON
- Remaining
  - Update the table data
  - Update graph data instead of replacing whole graph
  - Remove duplication of data-preparation
  - Add refreshing for other metric pages

### DIFF
--- a/web/assets/javascripts/metrics.js
+++ b/web/assets/javascripts/metrics.js
@@ -8,6 +8,7 @@ class JobMetricsOverviewChart extends BaseChart {
   }
 
   get datasets() {
+    console.log(this.options)
     return Object.entries(this.options.series)
       .filter(([kls, _]) => this.visibleKls.includes(kls))
       .map(([kls, _]) => this.buildDataset(kls));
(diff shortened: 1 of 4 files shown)


web/assets/javascripts/metrics.js
```diff
@@ -60,6 +61,17 @@ class JobMetricsOverviewChart extends BaseChart {
     };
   }
 
+  update(options) {
```
davidwessman, line 64: @adamlogic I tried to write an update method based on the documentation for Chart.js but kind of got stuck.
How would you implement it? Would like to be able to update the datasets of existing chart.

adamlogic: You should be able to mutate `this.chart.data.datasets` then call `this.chart.update()`.

Take a look at `RealtimeChart` for how I'm live-updating the chart on the dashboard. You might be able to extract a lot of that into something that can be used by both `dashboard-charts.js` and `metrics.js`.

https://github.com/adamlogic/sidekiq/blob/0039b849c061ab43f29019d2dd2aeb0310705799/web/assets/javascripts/dashboard-charts.js#L76-L82

davidwessman: Thanks, will check it out! 🙂

PR #5547, opened 2022-09-28
Add ability to parse capsules from a config file

BobbyMcWho:
I figured I would throw this together really quick, as it would be beneficial for cases like ours where we use multiple sidekiq config files and would like to configure a capsule for a single one of them.


mperham:
I declare this PR totally sick.

PR #5556, opened 2022-10-01
Add pagination to "Busy" page

### LINKED ISSUE #5555: Add pagination to running jobs on "Busy" page
We have lots of sidekiq's running and they are processing thousands of jobs at the same time. Opening the "Busy" page just hangs. My mac just hung and rebooted because of this 10 minutes ago.
Would you be open for a PR adding a pagination?

fatkodima:
Closes #5555.

The pagination is slower (compared to `/retries` etc pages), because we need to get all of the entries to sort them (by `run_at`) before returning. But now this loads with no problems.

Seems like it is also problematic to add mass actions like "Kill"/"Delete", because we will need to inefficiently iterate over the large list `list` (compared to `/retries` page where we can efficiently remove from the `zset`).

You mentioned filtering in the issue. This is Pro/Ent feature, right? So I have not access to it. Better to be implemented by you or someone else who has access.

### DIFF
--- a/lib/sidekiq/paginator.rb
+++ b/lib/sidekiq/paginator.rb
@@ -43,5 +43,13 @@ module Sidekiq
         end
       end
     end
+
+    def page_items(items, pageidx = 1, page_size = 25)
+      current_page = pageidx.to_i < 1 ? 1 : pageidx.to_i
+      pageidx = current_page - 1
+      starting = pageidx * page_size
+      items = items.to_a
+      [current_page, items.size, items[starting, page_size]]
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


fatkodima:
Thanks, @mperham. Agreed with the suggestions.

ngan:
We're seeing an error when visiting the Busy page after this upgrade:
```
undefined method `size' for nil:NilClass
```

Maybe `@workset` can be `nil`?

mperham:
@ngan Please provide the full backtrace.

ngan:
Apologies, this might be something on our end. We're investigating further. If it's a bug on Sidekiq's side, I'll report back with lots more information.

PR #5559, opened 2022-10-03
Speedup iterating over `WorkSet`

### LINKED ISSUE #5472: Proposal to improve `WorkSet#each` method
**Sidekiq / Pro / Enterprise version(s)**: Sidekiq 6.5.1 (but also related to latest version -- 7)

It has come to my attention that it seems like [current implementation of WorkSet#each](https://github.com/mperham/sidekiq/blob/7-0/lib/sidekiq/api.rb#L1048-L1069) can be improved. I have an idea to improve it but I'd like to understand first the reasons behind the current approach has been used in there.

The amount of `pipelined` calls to redis will depend on the amount of Sidekiq processes running. With that, if `Sidekiq::WorkSet.new.any? { ... }` is called very frequently (which is my case -- because I'm using it to prevent jobs duplication to be run), reducing the amount of such `pipelined` calls could reduce significantly the load on Redis (which we've been facing lately on the company I work for). And I know that preventing duplicated jobs from running at same time is an included feature in Sidekiq Enterprise (we already use Sidekiq Pro and we might get to use Enterprise version quite soon), but my real question about this context is: why is that logic really bothering in filtering out inexistent workers with that `exists?` call on line 1054 while it could simply call all the `pipeline.hgetall("#{key}:work")` for all the processes at once in a single `pipelined` call to redis and just skip the returned empty hashes? Would such approach impact too much performance or anything? If not, I could send a PR to change that.

fatkodima:
Closes #5472.

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -1103,21 +1103,28 @@ module Sidekiq
 
     def each(&block)
       results = []
+      procs = nil
+      all_works = nil
+
       Sidekiq.redis do |conn|
-        procs = conn.sscan_each("processes").to_a
-        procs.sort.each do |key|
-          valid, workers = conn.pipelined { |pipeline|
-            pipeline.exists?(key)
+        procs = conn.sscan_each("processes").to_a.sort
+
+        all_works = conn.pipelined do |pipeline|
+          procs.each do |key|
             pipeline.hgetall("#{key}:work")
-          }
-          next unless valid
-          workers.each_pair do |tid, json|
-            hsh = Sidekiq.load_json(json)
-            p = hsh["payload"]
-            # avoid breaking API, this is a side effect of the JSON optimization in #4316
-            hsh["payload"] = Sidekiq.load_json(p) if p.is_a?(String)
-            results << [key, tid, hsh]
           end
+        end
+      end
+
+      procs.zip(all_works).each do |key, workers|
+        workers.each_pair do |tid, json|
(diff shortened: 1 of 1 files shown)


lib/sidekiq/api.rb
```diff
@@ -1105,13 +1105,17 @@ def each(&block)
       results = []
       Sidekiq.redis do |conn|
         procs = conn.sscan_each("processes").to_a
-        procs.sort.each do |key|
-          valid, workers = conn.pipelined { |pipeline|
-            pipeline.exists?(key)
+
+        all_workers = conn.pipelined do |pipeline|
+          procs.sort.each do |key|
```
mperham, line 1110: I think this will give the wrong results. You need to persist the sorted results or else the zip below will not interleave the results with the correct proc. In other words, just add `.sort` to the end of line 1107 and remove it here.

fatkodima: Yes, sure 🤦 Thank you.

mperham:
The only other obvious enhancement is to move up the `end` of the Sidekiq.redis call so we can return the redis connection back to the pool before spending a long time parsing JSON and sorting the results, it doesn't look like we use it in the last half of the method, can you return the connection after line 1113?

fatkodima:
Updated.

mperham:
Thank you for your good work here. It'll go into the next 6.5 patch release and I'll merge it into 7.0b2 also.

PR #5565, opened 2022-10-08
Remove deprecated worker reference.

seuros:
The worker naming was deprecated in sidekiq 6.3. 
In version 7, we can use directly Sidekiq:Job in tests.

Application that depend on Sidekiq::Worker can require "sidekiq/worker_compatibility_alias" in their  application.rb or sidekiq initalizer.

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -8,6 +8,7 @@ require "sidekiq/logger"
 require "sidekiq/client"
 require "sidekiq/transaction_aware_client"
 require "sidekiq/job"
+require "sidekiq/worker_compatibility_alias"
 require "sidekiq/redis_client_adapter"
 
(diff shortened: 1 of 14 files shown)


lib/sidekiq.rb
```diff
@@ -8,7 +8,6 @@
 require "sidekiq/client"
 require "sidekiq/transaction_aware_client"
 require "sidekiq/job"
-require "sidekiq/worker_compatibility_alias"
```
mperham, line 11: Everything else is fine but I don’t want to go this far. S::W is not officially deprecated, just no longer recommended.

seuros: Restored.  

I fixed the reference in more places.

mperham:
Thank you!

PR #5569, opened 2022-10-11
externalize sd_notify and require only when needed

seuros:
With this change, systemd integration will load only if sd_notify gem is present in the Gemfile.

- sd_notify is is the same gem used by puma.
- We do not need to test it here anymore.
- sd_notify can be loaded only in the watchdog thread.
- sd_notify is optional.


mperham:
I don’t like forcing the user to add the gem in order to use systemd. Is there an ENV we can use when running in systemd to conditionally load s/sd_notify?

seuros:
The gem has to be added in puma to use the  type=notify in the service.

On Jruby, sidekiq should have the type=normal.

ping @agis

lib/sidekiq/cli.rb
```diff
@@ -417,5 +417,5 @@ def rails_app?
   end
 end
 
-require "sidekiq/systemd"
+require "sidekiq/systemd" if Gem::Specification.find_all_by_name("sd_notify").any?
```
mperham, line 420: Too specific. What if the user wants to supply their own version or use a forked gem? Just catch LoadError like usual?

seuros: They can require the file manualy. 

The only requirement is that is should be named sd_notify.rb and SdNotify

mperham:
I think you missed the parts in `bin/sidekiq`. Honestly though I'm not seeing much value with this change. I don't want the user to have to add another gem to gain systemd support. I forked `lib/sidekiq/sd_notify.rb` years ago; it's very simple and has required zero maintenance.

Issue #5505, opened 2022-09-01
Metrics web pages should refresh minutely?


mperham:
The metrics home page is a great candidate for a dashboard on a TV on a wall but it needs to auto-refresh for that purpose. Should we configure /metrics to refresh at the top of each minute?

davidwessman:
I might take a look at this tomorrow, really looking forward to metrics 😍

davidwessman:
Should we use the existing live-poll-functionality, hard refresh or something custom?

If we are using the live-poll, we need to
- remove chart
- replace content
- initialize new chart

then the charts will flicker like in the video below:
https://user-images.githubusercontent.com/6763624/188074993-228ac085-ece0-4723-8d0e-27f4663d6157.mov

If we want flicker-free updates we need to handle the table via javascript as well.

mperham:
I envision the page only fetching the latest minute value and updating the graph/table. I don't think it would be good to refresh the entire page/content.

davidwessman:
Will look into reworking the table to fit with that vision.

davidwessman:
@mperham What branch should I base my changes on now until version 7 is released?

mperham:
Main. 6.x is closed except for important bug fixes.

PR #5574, opened 2022-10-16
Metrics: Refreshes chart and table every minute

davidwessman:
- Related to issue #5505, only updating the index view and not
  the /metrics/:job view.
- It was hard to introduce a new paradigm of updating a html table in pure javascript,
  would probably be possible to do some other abstractions to make it easier to work with.

### DIFF
--- a/lib/sidekiq/web/application.rb
+++ b/lib/sidekiq/web/application.rb
@@ -61,9 +61,31 @@ module Sidekiq
     end
 
     get "/metrics" do
-      q = Sidekiq::Metrics::Query.new
-      @query_result = q.top_jobs
       erb(:metrics)
+    end
+
+    get "/metrics.json" do
+      query = Sidekiq::Metrics::Query.new
+      job_results = query.top_jobs.job_results.sort_by { |(kls, jr)| jr.totals["s"] }.reverse.first(20)
+      json_data = {
+        labels: query.top_jobs.buckets,
+        marks: query.top_jobs.marks.map { |m| [m.bucket, m.label] },
+        tableData: job_results.map do |(kls, jr)|
+          [
+            kls,
+            {
+              success: jr.dig("totals", "p") - jr.dig("totals", "f"),
+              failure: jr.dig("totals", "f"),
+              total: jr.dig("totals", "s").round(2),
+              average: jr.total_avg("s").round(2)
+            }
+          ]
+        end.to_h,
+        series: job_results.map { |(kls, jr)| [kls, jr.dig("series", "s")] }.to_h,
+        server_utc_time: server_utc_time,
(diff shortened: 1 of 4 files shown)


davidwessman:
@mperham It took a few tries, but here is my first version of a fully refreshing chart and table in the Metrics view.

mperham:
Looks fine to me. @adamlogic do you have any comments?

adamlogic:
IMO this feature doesn't add enough benefit to justify the added complexity here. I feel like this adds a lot more indirection and coupling between the HTML, JS, and Ruby, and I'm having to jump around a lot to understand what's going on.

I realize the source of the complexity is needing to update both the table and the chart. We can't simply fetch new data and feed it to the chart the way we do on the dashboard.

Since we're reloading essentially the entire page, I wonder if we couldn't vastly simplify this by calling `location.reload()`. The challenge there is losing the state of the checkboxes. We could sync those up with the URL by updating the query string and reading that query string on page load, but again, is it worth the added complexity?

One functional gripe here is that I don't think this page should auto-refresh by default. It only shows the last hour of data, and if I'm doing some analysis, I don't want to run the risk of losing the data I'm analyzing. Auto-refresh should be opt-in, or at the very least have a pause button.

davidwessman:
> IMO this feature doesn't add enough benefit to justify the added complexity here. I feel like this adds a lot more indirection and coupling between the HTML, JS, and Ruby, and I'm having to jump around a lot to understand what's going on.

That makes sense to me as all, and if we would like live-reloading of charts + tables on other pages we almost need a front-end framework to make better abstractions than this.

> I realize the source of the complexity is needing to update both the table and the chart. We can't simply fetch new data and feed it to the chart the way we do on the dashboard.
> 
> Since we're reloading essentially the entire page, I wonder if we couldn't vastly simplify this by calling `location.reload()`. The challenge there is losing the state of the checkboxes. We could sync those up with the URL by updating the query string and reading that query string on page load, but again, is it worth the added complexity?

The coupling between the charts and the tables are already a little bit hacky, is there a ChartJS equivalent for tables that would make sense to try out? Would be easier to initialize all the data in javascript if this is the level of interaction that we want.

> One functional gripe here is that I don't think this page should auto-refresh by default. It only shows the last hour of data, and if I'm doing some analysis, I don't want to run the risk of losing the data I'm analyzing. Auto-refresh should be opt-in, or at the very least have a pause button.

It would absolutely make sense to have a button for enabling / disabling the auto-refresh.

mperham:
We could have the existing "Live Poll" button mean "reload this page every minute" but I'm also very much a fan of simplicity and Cmd+R is pretty dang simple. I'm ok with closing this as a prototype which didn't work out.

davidwessman:
Will see if I can extract any of the changes, but will close this for now. Should we close #5505 as well?

PR #5577, opened 2022-10-18
Extend perform_bulk to accept custom options hash similar to push_bulk

shayonj:
Some sidekiq plugins use `push_bulk` in sending batch jobs, along with some metadata to be passed along the job that can later be read from the Sidekiq server middleware. However, such a thing is not possible when with `perform_bulk`. It'd be really nice to support this where we can pass in the custom options/attributes and get the benefit of batched redis calls too with `batch_size`

Ex of passing custom options with `push_bulk`:

```ruby
 Sidekiq::Client.push_bulk('class' => YourWorker, 'args' => args, 'tenant_id' => 3)
```

We can do the same with perform_bulk now

```ruby
  set = MySetJob.set('queue' => 'bar')
  set.perform_bulk(args, custom_opts: { 'tenant_id' => 3})
```

Happy to open a tracking issue if thats helpful, let me know. Open to feedback and iterations 🙌🏾

### DIFF
--- a/lib/sidekiq/job.rb
+++ b/lib/sidekiq/job.rb
@@ -237,10 +237,10 @@ module Sidekiq
       end
       alias_method :perform_sync, :perform_inline
 
-      def perform_bulk(args, batch_size: 1_000)
+      def perform_bulk(args, batch_size: 1_000, custom_opts: {})
         client = @klass.build_client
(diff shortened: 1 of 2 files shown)


mperham:
You can you use set to add any custom attributes not just queue.

Issue #5585, opened 2022-10-25
Web UI - Busy page - Sorting processes by name


mobilutz:
Ruby version: 3.1.2-p20
Rails version: 6.1.7
Sidekiq / Pro / Enterprise version(s): Sidekiq ~6.1.7~ 6.5.7 / Pro 5.5.5

The processes listed on the WebUIs budy page are currently not sorted alphabetically, see screenshot.

![Screen Shot 2022-10-25 at 15 53 24](https://user-images.githubusercontent.com/491877/197793239-fd1cb129-e252-4e29-890a-313d963b4c16.png)

This is more a design issue, but something that always takes me a bit to make sure the correct workers are running.

I tried to look into this, but didn't find any way to ass sorting to `SSCAN` on redis. And maybe I am even looking at the wrong command.

mperham:
Upgrade.

mobilutz:
Ah, I copied the wrong information, we are on Sidekiq 6.5.7.
Will update the original message (with leaving the old data in it)
![Screen Shot 2022-10-25 at 17 49 18](https://user-images.githubusercontent.com/491877/197821304-800041ab-a218-4f95-877e-9164b37522ea.png)

mperham:
Sorting was reworked in #5398. Sounds like there’s an issue. Was the previous sorting behavior better?

mobilutz:
Thanks for the hint. I will check out if reverting this results in better output.

mobilutz:
You are correct, the implementation of #5398 is causing our processes list to be incorrectly sorted.

The problem seems to be, that the new implementation defines the `split_characters = /[\._-]/` are bit too aggressive.

I now have a place to start, so I will look into opening a pull-request is the next couple of days.

mobilutz:
This PullRequest fixes the sorting issue:
https://github.com/mperham/sidekiq/pull/5587

PR #5587, opened 2022-10-25
Fix natural sorting for processes

mobilutz:
The processes sorting was not correct for hostnames with `_` in it.
Ref: #5585 

This changes it so that all cases work.

Kudos to `shurikk` on Stackoverflow:
https://stackoverflow.com/a/15170063/575547


mperham:
Get rid of the pry require.

mobilutz:
> Get rid of the pry require.

Forgot to remove it after debugging.

mperham:
@KevinBongart Are you ok with this revision to your sorting logic? Any feedback?

PR #5600, opened 2022-10-30
Metrics: Removes polling button

### LINKED ISSUE #5595: Metrics page breaks with polling enabled
I've just upgraded to Sidekiq 7.0 and noticed that the metrics page breaks when the polling is enabled and it refreshes the data. The first load works, but then the graph breaks.

Ruby version: 3.1.2
Rails version: 7.0.4
Sidekiq / Pro / Enterprise version(s): 7.0

davidwessman:
- Fixes #5595

### DIFF
--- a/lib/sidekiq/web/helpers.rb
+++ b/lib/sidekiq/web/helpers.rb
@@ -323,6 +323,16 @@ module Sidekiq
       Time.now.utc.strftime("%H:%M:%S UTC")
     end
 
+    # Routes of the web UI where Live Poll should not be available
+    NOT_POLLABLE = ["metrics"]
+
+    # Split current_path to be able to match for both "metrics" and "metrics/:name"
+    def pollable?
+      return false if current_path.blank?
+
+      NOT_POLLABLE.exclude?(current_path.split("/").first)
+    end
+
     def retry_or_delete_or_kill(job, params)
       if params["retry"]
         job.retry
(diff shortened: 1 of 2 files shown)


web/views/_poll_link.erb
```diff
@@ -1,4 +1,4 @@
-<% if current_path != '' %>
+<% if ['', 'metrics'].exclude?(current_path.split("/").first) %>
```
davidwessman, line 1: Maybe it would make more sense to configure it from the actons?
E.g. adding @polling_disabled to the /sidekiq and the metric-pages.

mperham: Yeah, this looks ugly but I don't have a good solution. Maybe a helper like `def pollable?` in web/helpers.rb?

davidwessman: Yeah, I will give it a go tomorrow 🙂

davidwessman: A helper method looks much cleaner!

lib/sidekiq/web/helpers.rb
```diff
@@ -323,6 +323,16 @@ def server_utc_time
       Time.now.utc.strftime("%H:%M:%S UTC")
     end
 
+    # Routes of the web UI where Live Poll should not be available
+    NOT_POLLABLE = ["metrics"]
+
+    # Split current_path to be able to match for both "metrics" and "metrics/:name"
+    def pollable?
+      return false if current_path.blank?
+
```
mperham, line 332: return !current_path.starts_with("metrics")

?

davidwessman: We also want to disable it for `/sidekiq`, which works with `current_path.blank?`.

But maybe it should be:

```ruby
def pollable?
  return false if current_path.blank? # /sidekiq
  
  NOT_POLLABLE.none? { |path| current_path.starts_with?(path) }
  end
```

PR #5618, opened 2022-11-04
Renames test files to `*_test.rb`

### LINKED ISSUE #5610: Rename test files
For Sidekiq 7.0, we renamed the test suite files from `test/test_feature.rb` to `test/feature.rb` to reduce unnecessary filename noise. Maxitest (and other test libraries?) don't detect test files with this naming scheme. Instead, rename these files again to `test/feature_test.rb` which it will detect so we can run the entire suite with a simple `bundle exec mtest`.

davidwessman:
- Fixes #5610


mperham:
Thank you!!!

PR #5621, opened 2022-11-05
Use start_with? instead of starts_with? which is only available on Rails

marcelolx:
Just noticed this failing in the sidekiq-scheduler test suite https://github.com/sidekiq-scheduler/sidekiq-scheduler/actions/runs/3399698575/jobs/5653621656#step:5:52

https://github.com/mperham/sidekiq/commit/2bfe30667a6be18a9bdce6415ee5ec3c4bf1a6de

### DIFF
--- a/lib/sidekiq/web/helpers.rb
+++ b/lib/sidekiq/web/helpers.rb
@@ -324,7 +324,7 @@ module Sidekiq
     end
 
     def pollable?
-      !(current_path == "" || current_path.starts_with?("metrics"))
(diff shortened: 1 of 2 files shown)


marcelolx:
not sure why 3.1 is broken 🤔

mperham:
Argh, Matz English strikes again.

tisba:
I think this hasn't been released yet, right? 👀

mperham:
Not yet, I'll prepare 7.0.2 this week. There's a major Enterprise bug which needs a 7.0.2 release too.

Issue #5613, opened 2022-11-01
Feature Proposal: Hooks in job info ERB file to allow for custom rows in the UI


pharris2411:
I've created a bit of a Javascript hack to parse job ID information in the UI and append additional rows for my own personal use, and I'd love to make this a more formalized capability. While it'd be easy to grab [_job_info.erb](https://github.com/mperham/sidekiq/blob/main/web/views/_job_info.erb) and override locally, we'd like to avoid this if possible to make upgrading easier. I think this could work by adding hooks into `_job_info.erb` similar to how custom tabs work, with additional rows being appended and content coming from a `call!` method of classes we pass in. Those classes would need access to the job details but read only.

I've attached an example screenshot -- the Datadog links are in the Job ID row, but could easily be their own row. (I also added a Sidekiq Batch link to make linking between batches and jobs easier -- I think this would be a great addition to the base product even without the need for a custom hook.)

If this seems like a feature you'd be interested I'd be happy to put together a PR. Thanks!

<img width="1247" alt="196208589-f83a452c-24da-41b0-8185-befcc460b5dc" src="https://user-images.githubusercontent.com/1631278/199349189-8077ab36-c79c-447a-bd3c-f43b0b69832c.png">

mperham:
Much of that could be auto-detected and built in (e.g. if the payload contains a BID, link to /batches/BID). I'd be happy to get a PR with any improvements.

pharris2411:
Closing as this PR has been merged: https://github.com/mperham/sidekiq/pull/5624

PR #5623, opened 2022-11-06
Adds a helper function to get batch id, and provides a link from the job info section

pharris2411:
This PR is in reference to [this issue](https://github.com/mperham/sidekiq/issues/5613), and adds parsing the job details to look for `bid`, indicating this is from a Sidekiq Batch. If found it outputs the ID as a link to the batch status.

Example (with a fake ID):

Screenshot 2022-11-06 at 1 06 00 PM

### DIFF
--- a/lib/sidekiq/web/helpers.rb
+++ b/lib/sidekiq/web/helpers.rb
@@ -262,6 +262,10 @@ module Sidekiq
       end
     end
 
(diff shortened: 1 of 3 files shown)


mperham:
I appreciate the PR, I went with a slightly different impl but the end result should be the same. Thank you!

PR #5624, opened 2022-11-06
Adds a hook into job info output to allow custom rows like links to external logs

pharris2411:
This PR adds a hook into the job info output to allow appending custom rows. For example this could be used to generate links to external log providers such as Datadog.

To use, a class must implement `show_row`, `row_name` and `row_contents`, for example:

```ruby
class CustomJobInfoRowLink
  class << self
    
    EXTERNAL_LOG_LINK = 'https://example.com/logs/job_id='
    
    def show_row(job)
      true
    end
    
    def row_name
      "Logs"
    end
    
    def row_contents(job)
      "<a href='#{EXTERNAL_LOG_LINK}#{job.jid}'>External logs</a>"
    end
  end
end
```

Then during setup of the Sidekiq web component, pass in to `custom_job_info_rows`:

```ruby
Sidekiq::Web.custom_job_info_rows << CustomJobInfoRowLink
```
Screenshot 2022-11-06 at 1 45 11 PM

### DIFF
--- a/lib/sidekiq/web.rb
+++ b/lib/sidekiq/web.rb
@@ -48,6 +48,10 @@ module Sidekiq
       end
       alias_method :tabs, :custom_tabs
 
+      def custom_job_info_rows
+        @custom_job_info_rows ||= []
+      end
+
       def locales
         @locales ||= LOCALES
       end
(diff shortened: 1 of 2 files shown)


mperham:
Thank you! I'm going to adjust the API to be a little more Ruby idiomatic. These will be in 7.0.2 (no ETA).

PR #5635, opened 2022-11-14
Allow shift-clicking to select ranges of jobs

jdelStrother:
Any interest in something like this to allow you to shift-click ranges of checkboxes on the retries/scheduled-jobs pages?


mperham:
Seems fine to me.

Issue #5632, opened 2022-11-10
RFC: Sidekiq.local_config


bf4:
Sidekiq / Pro / Enterprise version(s): 7.x

In https://github.com/mperham/sidekiq/blob/b4092e33f97e32c0e4e3f57c38c351867c5bc87b/lib/sidekiq.rb#L68-L74

We have

```ruby
  def self.redis_pool
    (Thread.current[:sidekiq_capsule] || default_configuration).redis_pool
  end

  def self.redis(&block)
    (Thread.current[:sidekiq_capsule] || default_configuration).redis(&block)
  end

  def self.default_configuration
    @config ||= Sidekiq::Config.new
  end
```

Right now, to get the current concurrency or timeout you have to 

```ruby
sidekiq_config = Thread.current[:sidekiq_capsule]&.config || Sidekiq.default_configuration
sidekiq_config[:concurrency] #=>  5
sidekiq_config[:timeout] #=> 25
```

I get that you don't want something that looks like a global config state, so maybe it could be called `Sidekiq.local_config` or something, but I think it would be nice to expose what is essentially, I think, a public API.

Side note:  While investigating this change, I discovered the reason for the ssl errors I got in my original upgrade were because default_configuration is `Sidekiq::Config.new` which doesn't take into account `Sidekiq.configure_client do |config|  config.redis = redis_config end` or `Sidekiq.configure_server do |config|  config.redis = redis_config end`, which means that calling `Sidekiq.redis {|c| c.call("PING") }` could fail when defined as `Sidekiq::Config.new.redis {|c| c.call("PING") }`.  It appears that even though `default_configuration` is defined as `Sidekiq::Config.new` in production, Sidekiq.default_configuration.redis {|c| c.call("PING") }` works while `Sidekiq::Config.new.redis {|c| c.call("PING") }` gives me the SSL_connect error. Relevance of this comment to this issue is just around surprises to find in finding or building the config.

mperham:
The timeout is global so you can always go to `Sidekiq.default_configuration[:timeout]`.

I think your issue is that concurrency is no longer a single value in Sidekiq. There is the concurrency for the default capsule but there are concurrency values for other capsules too. Rather than asking me for a specific implementation, maybe tell me what you are trying to achieve?

bf4:
When logging certain errors, we add context like

- primary database concurrency/statement timeout
- replica database concurrency/statement timeout
- if is running in sidekiq, then concurrency/timeout

[...]

PR #5647, opened 2022-11-18
Use configured Redis pool in Sidekiq::Deploy

bf4:
I noticed this when discussing https://github.com/mperham/sidekiq/issues/5632#issuecomment-1320451422

now

```ruby
Sidekiq::default_configuration.redis_pool.with {|c| c.call("PING") }
=> "PONG"
```

was

```ruby
Sidekiq::RedisConnection.create.with {|c| c.call("PING") }
/app/vendor/bundle/ruby/3.1.0/gems/redis-client-0.11.1/lib/redis_client/ruby_connection.rb:62:in `connect_nonblock': SSL_connect returned=1 errno=0 peeraddr=IP:PORT state=error: certificate verify failed (self signed certificate in certificate chain) (RedisClient::CannotConnectError)
```

### DIFF
--- a/lib/sidekiq/deploy.rb
+++ b/lib/sidekiq/deploy.rb
@@ -25,7 +25,7 @@ module Sidekiq
       Sidekiq::Deploy.new.mark(label: label)
     end
 
-    def initialize(pool = Sidekiq::RedisConnection.create)
+    def initialize(pool = Sidekiq.default_configuration.redis_pool)
       @pool = pool
     end
 


mperham:
I had to roll this back. One core idea behind deploy is that it should require as little of Sidekiq as possible.  It's designed to run inside deployment scripts which should run very fast and not require tens or hundreds of MB of RAM.

By depending on Sidekiq.default_configuration, we have to require sidekiq.rb which pulls in a lot of Sidekiq core. As is, S::Deploy only requires `sidekiq/redis_connection`. Use REDIS_URL to control where the deploy mark goes.

mperham:
In 7.0.2 you'll be able to do this if you don't want it to use REDIS_URL and use the Sidekiq initializer instead:

```ruby
Sidekiq::Deploy.new(Sidekiq.default_configuration.redis_pool).mark!
```

PR #5674, opened 2022-11-30
Only modify logger if it's an instance of `::Logger`.

ioquatix:
https://github.com/mperham/sidekiq/discussions/5673

At least don't assume that the `logger` argument is compatible with `extend(Sidekiq::LoggingUtils)`.

### DIFF
--- a/lib/sidekiq/config.rb
+++ b/lib/sidekiq/config.rb
@@ -248,7 +248,10 @@ module Sidekiq
         return
       end
 
-      logger.extend(Sidekiq::LoggingUtils)
+      if logger.is_a?(::Logger)
+        logger.extend(Sidekiq::LoggingUtils)
+      end
+
       @logger = logger
     end
 


mperham:
8e3f8f50

mperham:
With this change, custom loggers will just silently lose the ability for job-specific log levels. The test shows how to extend S::LoggingUtils in order to restore this functionality.

ioquatix:
My PR didn't break this use case for cases where it worked, but still mutated the incoming logger argument. While I think 8e3f8f5 is the best approach, it's also the least compatible. The compatibility is probably unlikely to be missed, as the default logger retains this functionality, so it only affects people deliberately assigning to `Sidekiq::Config#logger=`.

ioquatix:
@mperham is there any chance to cut a release with this change? As we are blocked on it.

mperham:
I've got a number of other things in the fire right now but I hope to release 7.0.2 this week.

Issue #5686, opened 2022-12-06
No more `Sidekiq.logger =` ?


DannyBen:
Ruby version: 3.1.3p185
Rails version: Not on rails
Sidekiq / Pro / Enterprise version(s): 7.0.2

I am getting an error in my setup when trying to assign my custom logger with:

```ruby
Sidekiq.logger = ...
# => NoMethodError: undefined method `logger=' for Sidekiq:Module
```

This used to work with 6.x

Looked for hints in [the upgrade guide](https://github.com/mperham/sidekiq/blob/main/docs/7.0-Upgrade.md), and the [Logging wiki](https://github.com/mperham/sidekiq/wiki/Logging), which still shows it should work.

mperham:
You have to use configure_server:

```ruby
Sidekiq.configure_server do |config|
  config.logger = ...
end
```

DannyBen:
Thank you.
So just note that the first section in the Logging wiki is incorrect (in more than one aspect... 😈): 

> Sidekiq's default logger is pretty simple:
> 
> ```ruby
> Sidekiq.logger = Logger.new(STDOUT)
> ```
> 
> Why make things more complex? YAGNI.
```

mperham:
The wiki still mostly reflects Sidekiq 6.x as most people have not yet upgraded to 7.

DannyBen:
Then perhaps the nice version-specific upgrade guides should mention notable changes? Would probably save time for both the owners and the users.

mperham:
The biggest issue is that we don't really have an upgrade document that focuses on API changes. I'll go through the wiki and clean up things where I can.

DannyBen:
> Send me a PR which has the fix you want to see? 

I can do that.

> The biggest issue is that we don't really have an upgrade document

Well, my two cents as a user:

1. When any release of any software, is immediately accompanied by updated documentation - it helps.
2. When there is a short upgrade guide between version - sort of like Rails is doing - it helps.

I treat documenting in the same way I treat specs. If the docs are not ready, the release is not ready.

PR #5687, opened 2022-12-07
Update upgrade instructions

DannyBen:
This PR is proposed following the discussion in #5686.

This is what I would expect in a version specific upgrade document - a list of short and clear changes the developers need to perform in their code in order to be compatible.

I only know of this one change, related to `Sidekiq.logger=` being gone, but if there are any, they should go in there in my opinion.

Feel free to discard this PR if it is not in line with how you envision these upgrade docs.

### DIFF
--- a/docs/7.0-Upgrade.md
+++ b/docs/7.0-Upgrade.md
@@ -88,14 +88,27 @@ They expect this support to be available in July 2023 with their rollout of Redi
 ## Upgrade
 
 As always, please upgrade Sidekiq **one major version at a time**.
-If you are already running Sidekiq 6.x, then:
+If you are running Sidekiq 5.x, follow the instructions in 
+[docs/6.0-Upgrade.md](docs/6.0-Upgrade.md) first, and then come back to this
+guide.
 
-* Upgrade to the latest Sidekiq 6.x.
+If you are running Sidekiq 6.x, follow these steps to upgrade:
+
+### 1. Upgrade your Gemfile to the latest Sidekiq 7.x
+
 ```ruby
-gem 'sidekiq', '< 7'
+gem 'sidekiq', '~> 7.0'
 ```
-* Fix any deprecation warnings you see.
-* Upgrade to 7.x.
+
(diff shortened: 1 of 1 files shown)


DannyBen:
I don't know, it is your call. My opinion is this:

1. As a user of many gems, where I am spending about 30% of my time upgrading to latest versions, I want my upgrade process to hopefully be without the need to make any changes.
2. If there are changes needed, I want to see them in an easy to find place. I am mentioning it, since Sidekiq already has several places to search for this info, some of which is (intentionally apparently) out of date, such as the wiki, and in these discussions even more documents are mentioned. I say less, not more. Preferably one place, which is up to date + one "todo list" for upgrading.
3. When I am upgrading, I want to spend as little time as possible on it - this is why a list such as this helps. If there is only one thing that changed in the API, it is ok for this list to only include one thing. 
4. The upgrade checklist I am thinking about, is something that the rails documentation site is doing quite well. On each major release, there are a) up to date documentation and b) a step by step guide in the same spirit of "before it was this, now it is that".


As for this:

> The vast majority of Sidekiq users don't have `Sidekiq.logger=`

I am sure you realize that:
1. It is not necessarily a niche feature.
2. It is the first thing that appears on the logging Wiki page
3. The wiki is not showing version 7 changes intentionally
4. The upgrade guide does not mention it

So.... this leaves people who need to upgrade with no other choice other than to dig into the commit log, or open an issue/discussion.

I know (guess) that the "community edition" of Sidekiq is at the bottom of the priority list, but in my opinion at least one place should mention the correct and **up to date** way of doing things.

mperham:
I agree with everything you said! This feedback is how things improve. I will start an API migration doc.

DannyBen:
Beautiful and helpful.

The other upgrade doc now looks like a high level change log plus generic "how to upgrade ruby stuff" instructions. If this is its intention, I guess all is good.

The nice thing about this new document, is that now I can relax, knowing there are no changes that my specs might have missed. Easy to follow, short and copy-paste-able.

Issue #5693, opened 2022-12-11
Suggestion: Sidekiq 7 metrics history configurable in web UI


haffla:
Hello!

The new metrics in the web UI are great but only go back 60 minutes. I am not sure if this is an intentional limitation for the free Sidekiq version or for performance reasons? In the beginning it confused me a lot. I thought something is wrong because I knew for sure that some jobs had run last night. Only later I saw the tiny hint at the bottom of the page and realized that metrics only collect data from one hour ago.

What do you think about making it configurable in the web UI using a select input for instance with some standard options like 1h, 6h, 12h, 1day, 2days, 7days? Would you accept a PR? Again sorry if that's not something that you want to offer for users of the free version. Feel free to just close. Cheers.

mperham:
We started small to make sure the feature was solid before expanding the scope. I don’t want to expand into a true APM competitor, data storage can really fill up Redis if we aren’t careful. Right now I think the data lives for 8 hours so we could expand the UI to display 1, 2, 4 and 8 hour windows.

PR #5694, opened 2022-12-13
Make period on metrics page configurable via select

haffla:
Hey!

Following up on #5693. I've added a select to the metrics page (both index and detail) that let's you select between 1, 2, 4, and 8 hours of history. @mperham are you sure that data lives for 8 hours?

### Index page
Screenshot 2022-12-13 at 23 35 40

### Detail page
Screenshot 2022-12-13 at 23 35 54

### DIFF
--- a/test/web_test.rb
+++ b/test/web_test.rb
@@ -810,4 +810,54 @@ describe Sidekiq::Web do
       assert_equal :after, Thread.current[:some_setting]
     end
   end
+
+  describe "metrics" do
+    before do
+      result_mock = MiniTest::Mock.new
+      result_mock.expect(:job_results, [])
+      result_mock.expect(:marks, [])
+      result_mock.expect(:buckets, [])
+      result_mock.expect(:starts_at, Time.now - 3600)
+      result_mock.expect(:ends_at, Time.now)
+
+      @query_mock = Minitest::Mock.new
+      @query_mock.expect :top_jobs, result_mock do |minutes:|
+        assert_equal expected_minutes, minutes
+      end
+    end
+
+    subject do
+      Sidekiq::Metrics::Query.stub :new, @query_mock do
+        get "/metrics", {period: period}.compact
+        assert_equal 200, last_response.status
+        assert_match(/Metrics/, last_response.body)
+      end
+    end
+
+    context "when period param is not provided" do
+      let(:period) { nil }
+      let(:expected_minutes) { 60 }
+      it "calls top_jobs with minutes: 60" do
+        subject
+      end
+    end
+
+    context "when period param is unknown" do
+      let(:period) { "2d" }
+      let(:expected_minutes) { 60 }
+      it "calls top_jobs with minutes: 60" do
+        subject
+      end
+    end
+
+    Sidekiq::WebApplication::METRICS_PERIODS.each do |code, minutes|
+      context "when period param is #{code}" do
+        let(:period) { code }
+        let(:expected_minutes) { minutes }
+        it "calls top_jobs with minutes: #{minutes}" do
+          subject
+        end
+      end
+    end
+  end
 end
--- a/test/web_test.rb
+++ b/test/web_test.rb
@@ -26,7 +26,7 @@ module TestScenariosForMetrics
     context "when period param is not provided" do
       let(:period) { nil }
       let(:expected_minutes) { 60 }
-      it "calls top_jobs with minutes: 60" do
+      it "queries using 60 minutes" do
         subject
       end
     end
@@ -34,7 +34,7 @@ module TestScenariosForMetrics
     context "when period param is unknown" do
       let(:period) { "2d" }
       let(:expected_minutes) { 60 }
-      it "calls top_jobs with minutes: 60" do
+      it "queries using 60 minutes" do
         subject
       end
     end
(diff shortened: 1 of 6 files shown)


mperham:
Yes, the metrics data TTL is currently 8 hours: https://github.com/mperham/sidekiq/blob/52d2a6f155b909239b2778ca5d2dad2a5d263022/lib/sidekiq/metrics/tracking.rb#L53

test/web_test.rb
```diff
@@ -18,6 +18,39 @@ def add_pair(job)
...
 end
 
+# Common test scenarios for both metrics index and detail page.
+module TestScenariosForMetrics
+  extend Minitest::Spec::DSL
+
+  def self.included(base)
+    context "when period param is not provided" do
+      let(:period) { nil }
+      let(:expected_minutes) { 60 }
```
mperham, line 28: This abstract style of testing is not the Sidekiq test suite style at all. Your tests should be directly readable. No modules, no spec DSLs, just:

```ruby
def test_something
  setup_code
  run the code to test
  assert results
end
```

m-adnan-iqbal (approved):
LGTM

test/web_test.rb
haffla: > This abstract style of testing is not the Sidekiq test suite style at all. Your tests should be directly readable. No modules, no spec DSLs, just:
> 
> ```ruby
> def test_something
>   setup_code
>   run the code to test
>   assert results
> end
> ```

Completely understood. I've updated the tests to be more in line with the rest.

mperham: Wow, that looks a lot better, much easier to read. Thank you for the polish.

StanBright:
> SHORT_TERM = 8 * 60 * 60 

Thanks @mperham, I'm using Sidekiq v7.0.2 (OSS) in three different projects, and it seems that metrics data is kept for 2h only. I couldn't find a relevant config or something related. Do you know what could be the reason?

mperham:
Nope. The TTL is clearly 8 hours.

mperham:
This PR still has a fundamental flaw: it tries to render all N datapoints on the graph. Works great at 60 minutes because you have 60 data points to render in horizontal 500-1000 pixels, no problem. But when you have 240 or 480 data points, the graph is pure chaos and unusable. You need to add logic to drop data points with longer timespans.

One hour:
Screenshot 2022-12-21 at 3 08 13 PM

Eight hours:
Screenshot 2022-12-21 at 3 08 26 PM

mperham:
I just committed some changes to main in f68aa4ed that will likely require you to rebase.

haffla:
Thanks @mperham. I will try to come up with a solution if I find some time these days. I remember some stuff from a data analyst course I once did. Only taking every Nth data point is probably a viable solution, even random sampling in some cases. Also did some quick research and stumbled upon this: https://en.wikipedia.org/wiki/Kernel_smoother. 

But to be clear: You don't want the output of `Metrics::Query` to change but rather be preprocessed before it is plotted, right?

mperham:
> But to be clear: You don't want the output of `Metrics::Query` to change but rather be preprocessed before it is plotted, right?

Right. The existing API should still return the full dataset. We can provide a new Ruby method to return a "smoothed" dataset of 60 points which is graph-ready or we can add the JS necessary to do the same. It might be better to do the latter if we ever want to stream new datapoints into a live graph.

mperham:
Wait, now that I think about it: the metrics graphs all display elements which are Counts. We can just sum the values. Execution Time and Count can both be summed up from the N data points. No need for smoothing at all, just sum up the f, p and ms integers. So for the eight hour view, you just sum up every eight data points, etc.

StanBright:
Thanks for your answer @mperham. I know it's odd. I double checked, and the data being kept is only 1h actually ¯\_(ツ)_/¯

Screen Shot 2022-12-23 at 12 33 28 pm

StanBright:
OK, thanks... I went through the docs again:
`The graphs only show data for the last hour and the data is only retained in Redis for 8 hours.`

Apparently I had a misunderstanding between shown data and retained data.

mperham:
Any idea if you can find time to work on this? It's so close to mergeable! I'd be happy to pair with you on it via Tuple if interested.

haffla:
Hey! Sorry, I let this slip a little. But I managed to come up with something now. I remember you said that it might be safer to do it in JS. Hence the solution in `metrics.js`. Let me know what you think about it.

The job detail metrics view probably needs a similar treatment. I can start working on that once you approve the general approach.

mperham:
Can you post a screenshot so I don't have to spend 20 minutes reproducing a realistic graph?

haffla:
### Before

Screenshot 2023-01-23 at 23 57 01

### After

Screenshot 2023-01-23 at 23 57 25

mperham:
Does your code properly handle holes in the dataset? If I process a bunch of jobs in one minute and then things go quiet for 10 minutes and then I get another flurry of jobs, will the buckets be calculated correctly even if there's no data for a particular bucket?

haffla:
Yes certainly. It's sums up the values for each bucket. If there aren't any values it goes on to the next bucket. I added some comments. You also see in the picture: All the buckets starting from 14:58 until somewhere around 20:30 do not have any data. It's full of wholes actually.

mperham:
Thank you and well done!

Issue #5704, opened 2022-12-22
Outdated instructions for max_retries config option in JobRetry class


severinkaelin:
Hello
First of all, thanks for the project, especially the recent 7.0 release.

While researching the `max_retries` option I came across the following comment in the codebase:

```ruby
# The default number of retries is 25 which works out to about 3 weeks
# You can change the default maximum number of retries in your initializer:
#
#   Sidekiq.options[:max_retries] = 7
```
[See in source](https://github.com/mperham/sidekiq/blob/main/lib/sidekiq/job_retry.rb#L52)

If I'm not mistaken, this no longer works with sidekiq 7. Due to the `Sidekiq::Config` change: https://github.com/mperham/sidekiq/blob/main/docs/capsule.md#sidekiqconfig

If this is indeed the case, it would be good if the comment in the code could be adjusted.

Thanks.

Kind regards
Severin

mperham:
Send a PR which fixes it?

severinkaelin:
> Send a PR which fixes it?

Sure, I'm happy to do that: https://github.com/mperham/sidekiq/pull/5705

severinkaelin:
Closing as this has been implemented in PR (merged): https://github.com/mperham/sidekiq/pull/5705

PR #5705, opened 2022-12-22
Remove comment with outdated instructions for max_retries config option

severinkaelin:
Details in #5704 

@mperham I'm not quite sure if it's fine for you to just remove the outdated instructions or if you would prefer to replace it with a specific explanation taking the new "configuration strategy" into account. Feel free to change it again. Thanks.

### DIFF
--- a/lib/sidekiq/job_retry.rb
+++ b/lib/sidekiq/job_retry.rb
@@ -47,8 +47,11 @@ module Sidekiq
   # to the job and everyone is using an error service, right?
   #
   # The default number of retries is 25 which works out to about 3 weeks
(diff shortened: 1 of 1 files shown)


mperham:
You can now do this instead:

```ruby
Sidekiq.default_configuration[:max_retries] = 10
```

severinkaelin:
> You can now do this instead:
> 
> ```ruby
> Sidekiq.default_configuration[:max_retries] = 10
> ```

@mperham Thanks for pointing this out. I've adjusted the comment accordingly.

mperham:
thanks!

PR #5716, opened 2023-01-05
Don't allow quiet or stop on embedded processes

### LINKED ISSUE #5714: Stopping embedded server from the Web UI kills the whole process
Ruby version: 2.7.5p203
Rails version: 6.0.4.8
Sidekiq / Pro / Enterprise version(s): 7.0.2

As per title - if you stop sidekiq via the Web UI stop button, the web server it is embedded in also ends.

I would expect either:
1) embedded servers don't have the option to be controlled like this - they're embedded as part of the app so it doesn't really make sense
2) sidekiq stops but the process keeps running, this would probably leave an app that depends on an embedded sidekiq in a broken state though

mikebaldry:
- add embedded to Process
- don't allow quiet! or stop! 
- show in the Web UI as embedded 
- don't show the buttons to quiet or stop for embedded

Fixes #5714

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -876,7 +876,7 @@ module Sidekiq
         end
       }
 
-      result.each do |info, busy, at_s, quiet, rss, rtt|
+      result.each do |info, busy, beat, quiet, rss, rtt_us|
         # If a process is stopped between when we query Redis for `procs` and
         # when we query for `result`, we will have an item in `result` that is
         # composed of `nil` values.
@@ -884,10 +884,10 @@ module Sidekiq
 
         hash = Sidekiq.load_json(info)
         yield Process.new(hash.merge("busy" => busy.to_i,
-          "beat" => at_s.to_f,
+          "beat" => beat.to_f,
           "quiet" => quiet,
           "rss" => rss.to_i,
-          "rtt_us" => rtt.to_i))
+          "rtt_us" => rtt_us.to_i))
       end
     end
 
@@ -946,6 +946,24 @@ module Sidekiq
   #   'embedded' => true,
   # }
   class Process
+    def self.find(identity)
+      exists, (info, busy, beat, quiet, rss, rtt_us) = Sidekiq.redis { |conn|
+        conn.multi { |transaction|
+          transaction.sismember("processes", identity)
+          transaction.hmget(identity, "info", "busy", "beat", "quiet", "rss", "rtt_us")
+        }
+      }
+
+      return nil if exists == 0 || info.nil?
+
+      hash = Sidekiq.load_json(info)
+      Process.new(hash.merge("busy" => busy.to_i,
+        "beat" => beat.to_f,
+        "quiet" => quiet,
+        "rss" => rss.to_i,
+        "rtt_us" => rtt_us.to_i))
+    end
+
     # :nodoc:
     # @api private
     def initialize(hash)
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -828,6 +828,24 @@ module Sidekiq
   class ProcessSet
     include Enumerable
 
+    def self.[](identity)
+      exists, (info, busy, beat, quiet, rss, rtt_us) = Sidekiq.redis { |conn|
+        conn.multi { |transaction|
+          transaction.sismember("processes", identity)
+          transaction.hmget(identity, "info", "busy", "beat", "quiet", "rss", "rtt_us")
+        }
+      }
+
+      return nil if exists == 0 || info.nil?
+
+      hash = Sidekiq.load_json(info)
+      Process.new(hash.merge("busy" => busy.to_i,
+        "beat" => beat.to_f,
+        "quiet" => quiet,
+        "rss" => rss.to_i,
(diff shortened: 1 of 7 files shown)


lib/sidekiq/web/application.rb
```diff
@@ -82,11 +82,14 @@ def self.set(key, val)
 
     post "/busy" do
       if params["identity"]
-        p = Sidekiq::Process.new("identity" => params["identity"])
-        p.quiet! if params["quiet"]
-        p.stop! if params["stop"]
+        pro = processes.find { |p| p.identity == params["identity"] }
```
mikebaldry, line 85: Is there a way to get a single `Sidekiq::Process` by identity efficiently instead of loading the whole process list to get one? I figured this was fine anyway as the page you're on shows the whole list.

test/web_test.rb
```diff
@@ -86,6 +86,10 @@ def job_params(job, score)
     it "can quiet a process" do
       identity = "identity"
       signals_key = "#{identity}-signals"
+      @config.redis do |conn|
```
mikebaldry, line 89: these were required to have the process findable via the `ProcessSet` now

lib/sidekiq/launcher.rb
```diff
@@ -172,17 +172,15 @@ def ❤
               "quiet", @done.to_s,
               "rss", kb)
             transaction.expire(key, 60)
-            transaction.rpop("#{key}-signals")
+            transaction.rpop("#{key}-signals") unless @embedded
           }
```
mperham, line 176: It's possible to send a signal into Redis for the process and that signal will never get popped, leading to a memory leak. It might be better to pop the signal and change the Process.kill logic below to `if signal && !@embedded`.

mikebaldry: yes, that makes sense :+1: I'll update.

test/helper.rb
```diff
@@ -28,8 +28,10 @@
 
 def reset!
   # tidy up any open but unreferenced Redis connections so we don't run out of file handles
-  existing_pool = Sidekiq.default_configuration.instance_variable_get(:@redis)
-  existing_pool&.shutdown(&:close)
+  if Sidekiq.default_configuration.instance_variable_defined?(:@redis)
```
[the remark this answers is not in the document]
mikebaldry: It was `/home/mikeyb/projects/sidekiq/test/helper.rb:32: warning: instance variable @redis not initialized`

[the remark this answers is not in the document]
mperham: I get that, I just don't know how to reproduce it. The suite does not emit any warnings for me. I'm using Ruby 3.0.3, I'll try 3.1.

lib/sidekiq/web/application.rb
mikebaldry: This has been added. From what I can tell, the key that goes in `processes` is actually the identity but I would like to confirm this, as some of the tests use a different key to the identity they have in their info.

test/helper.rb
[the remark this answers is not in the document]
mikebaldry: I was accidentally on `2.7.5p203` - can confirm they don't show in `3.1.2`. Can drop this change if you want.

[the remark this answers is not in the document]
mperham: Please do.

lib/sidekiq/launcher.rb
mikebaldry: Updated. Please confirm the key that goes in to `processes` is the same as `identity`? I had to update the tests to pass as it was using a different key and identity, but `Launcher` has `key = identity` - so I think this is ok

mikebaldry:
@mperham Please confirm the key that goes in to `processes` is the same as `identity`? I had to update the tests to pass as it was using a different key and identity, but `Launcher` has `key = identity` - so I think this is OK

mperham:
Yes, the identity is effectively the primary key for a process so that's what you'd use to look up its data in Redis.

mikebaldry:
Great, I think it's good then, unless you have any other changes in mind?

lib/sidekiq/api.rb
```diff
@@ -943,8 +943,27 @@ def leader
   #   'busy' => 10,
   #   'beat' => <last heartbeat>,
   #   'identity' => <unique string identifying the process>,
+  #   'embedded' => true,
   # }
   class Process
+    def self.find(identity)
```
mperham, line 949: I'd prefer to avoid method names that clash with Enumerable, e.g. `ProcessSet.new.find(...)`. How about `Sidekiq::Process.lookup(identity)` or `Sidekiq::ProcessSet[identity]`, i.e. `def self.[](id)`?

mikebaldry: `Process` isn't `Enumerable` and I prefer to keep names consistent with what a user would expect (e.g. `ActiveRecord` `find` loads a model by id). Your repo though, I'll change to `lookup`.

mikebaldry: Actually, I prefer `ProcessSet#[]`. Makes sense that you can enumerate or get a single one via the same interface

mikebaldry: This has been updated

mperham:
Thank you!

PR #5730, opened 2023-01-17
Backport worker_compatibility_alias.rb to Sidekiq 6

adamniedzielski:
I believe that it's the missing bit to the renaming. This allows Sidekiq extensions to support both Sidekiq 6 and Sidekiq 7 by using `require "sidekiq/worker_compatibility_alias"` and referencing `Sidekiq::Worker`.

### DIFF
--- a/lib/sidekiq/worker_compatibility_alias.rb
+++ b/lib/sidekiq/worker_compatibility_alias.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+# use require "sidekiq/worker_compatibility_alias" if you want your code to
+# work both with Sidekiq 6 and Sidekiq 7
+
+require "sidekiq/worker"


mperham:
I'm not sure I want to do this. Most extensions need some changes to work with Sidekiq 7. I'd recommend that extensions bump their major version and set their gemspec to `add_dependency "sidekiq", ">= 6.5.0"` so that they can depend on the S::{Client,Server}Middleware modules and S::Job being available. Users can stay on previous versions if they don't want to upgrade Sidekiq. WDYT?

adamniedzielski:
Before falling asleep I realised that it's possible to simply do:

```ruby
require "sidekiq/job"

Sidekiq::Job
```

and this should work for Sidekiq >= 6.2.2. It's also a forward-looking change so better than what I proposed in the PR. Thanks for your input!

Issue #5743, opened 2023-01-24
Tools for tracking deprecations more programmatically


technicalpickles:
Currently, there are a few different ways that deprecations have been communicated through logs and code comments. In a large and ~old~ mature enough application, it can be really easy to miss during development, CI, and production because they are often just log messages in a see of other log messages

We are still on Sidekiq 6.x, and I know there are some notable change we still need to tackle for 7.0, which will be risky if we can't be sure we've addressed the deprecations.

I think ideally, we'd like a way to be notified of deprecations, and handle it ourselves: either raise an error (like in dev and CI), or log a message (in production). This is basically what you can do with [ActiveSupport::Deprecation](https://api.rubyonrails.org/classes/ActiveSupport/Deprecation.html), which we've had a lot of success with during big upgrades.

I assume a dependency on `active_suppor` would be out of the question? 😅 

I could imagine a `deprecation_handler` similar to how `error_handlers` work currently for hooking into them.

## Deprecation examples

On main:

https://github.com/mperham/sidekiq/blob/9feefc2d3653fc4c187ca3e46155d2736f2f3c62/lib/sidekiq/client.rb#L47

On 6.x:

https://github.com/mperham/sidekiq/blob/ad0f13cac5138b891d54ef2f2df3460d5151e34c/lib/sidekiq.rb#L97

https://github.com/mperham/sidekiq/blob/ad0f13cac5138b891d54ef2f2df3460d5151e34c/lib/sidekiq.rb#L225

https://github.com/mperham/sidekiq/blob/ad0f13cac5138b891d54ef2f2df3460d5151e34c/lib/sidekiq/redis_connection.rb#L14

https://github.com/mperham/sidekiq/blob/ad0f13cac5138b891d54ef2f2df3460d5151e34c/lib/sidekiq/rails.rb#L51-L56

https://github.com/mperham/sidekiq/blob/ad0f13cac5138b891d54ef2f2df3460d5151e34c/lib/sidekiq/job_util.rb#L27-L31

There are also some in sidekiq-ent and sidekiq-pro that are similar.

mperham:
Does Ruby have a solution for deprecations? Seems like this is a bigger API question for the community. I don't want a dependency on activesupport but if they are willing to extract the API into a separate gem which is intended to be a community-wide standard, I would consider that dependency.

technicalpickles:
`ActiveSupport::Deprecation` is the best I've seen, and I've used that in libraries that already has it.

[...]

PR #5745, opened 2023-01-24
Add Sidekiq::Config#handle_deprecation

technicalpickles:
A possible implementation for https://github.com/mperham/sidekiq/issues/5743 . It follows the pattern established by `error_handlers`.

The main difference from `error_handlers` is that it only allows _one_ handler. My intention here is that you'd only want one thing to actually output, so it doesn't make sense to allow multiple. My mind can be changed though.

If the approach is generally amenable, there are a few things I'd like to get in before merging:

- [x] include information about the `caller`, so users can find what is causing it
- [x] a public API with more natural name (`Sidekiq.deprecated`?)
- [ ] update existing deprecation warnings to use it

### DIFF
--- a/lib/sidekiq/config.rb
+++ b/lib/sidekiq/config.rb
@@ -41,9 +41,16 @@ module Sidekiq
       l.warn(ex.backtrace.join("\n")) unless ex.backtrace.nil?
     }
 
+    DEPRECATION_HANDLER = ->(msg, ctx) {
+      cfg = ctx[:_config] || Sidekiq.default_configuration
+      l = cfg.logger
+      l.warn "DEPRECATION WARNING: #{msg}"
+    }
+
     def initialize(options = {})
       @options = DEFAULTS.merge(options)
       @options[:error_handlers] << ERROR_HANDLER if @options[:error_handlers].empty?
+      @options[:deprecation_handler] = DEPRECATION_HANDLER if @options[:deprecation_handler].nil?
       @directory = {}
       @redis_config = {}
       @capsules = {}
@@ -217,6 +224,14 @@ module Sidekiq
       @options[:error_handlers]
     end
 
+    def deprecation_handler
+      @options[:deprecation_handler]
+    end
+
+    def deprecation_handler=(deprecation_handler)
+      @options[:deprecation_handler] = deprecation_handler
+    end
+
     # Register a block to run at a point in the Sidekiq lifecycle.
     # :startup, :quiet or :shutdown are valid events.
     #
@@ -249,6 +264,11 @@ module Sidekiq
       end
 
       @logger = logger
+    end
+
+    def handle_deprecation(msg, ctx = {})
+      ctx[:_config] = self
+      @options[:deprecation_handler].call(msg, ctx)
     end
 
     # INTERNAL USE ONLY
--- a/lib/sidekiq/config.rb
+++ b/lib/sidekiq/config.rb
@@ -41,10 +41,11 @@ module Sidekiq
       l.warn(ex.backtrace.join("\n")) unless ex.backtrace.nil?
     }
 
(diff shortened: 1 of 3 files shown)


mperham:
I'm not yet convinced. Any gem can have deprecations over time. I would rather see something which could work with any gem rather than something specific to Sidekiq.

scottjacobsen:
I already have my rails app setup to send all Active Support deprecations to Honeybadger which makes rails upgrades much easier. I know you wouldn't want to make SK depend on active support, but as an optional dependency that would be fantastic. For example `bundler install sidekiq_active_support_deprecations` and bam! - all my sk deprecations go to honeybadger with no additional work on my end. Please and thank you!

lib/sidekiq/config.rb
```diff
@@ -41,9 +41,17 @@ class Config
       l.warn(ex.backtrace.join("\n")) unless ex.backtrace.nil?
     }
 
+    DEPRECATION_HANDLER = ->(msg, ctx, caller) {
+      logger = Sidekiq.default_configuration.logger
+
+      line = caller[0]
+      logger.warn "DEPRECATION WARNING: #{msg} (called from #{line})"
```
davetron5000, line 48: (On Masto where @mperham's toot was mis-rendered for me as a poll and I responded instead of coming here)

A "print out a deprecation warning" message like this is challenging in two ways:
* It does not include the name of the gem producing this, so it can require grepping the bundle-installed gems to figure out where it's coming from
* The user of `self.deprecate` is not encouraged (or allowed?) to provide a solution to address the deprecation.

If I were making this more general, I would do:

```ruby
def self.deprecate(problem:, solution:)
  default.configuration.handle_deprecation(problem: problem, solution: solution, caller: caller)
end

# ...

DEPRECATION_HANDLER => ->(problem, solution, context, caller) {
  logger = Sidekiq.default_configuration.logger

  line = caller[0]
  logger.warn("[Sidekiq] DEPRECATION: #{problem}: This can be avoided by #{solution}")
}
```

Or maybe something better, but seeing this in my logs I would a) know it was about Sidekiq and b) be given some more info about what to do about removing the deprecated stuff I am using.  I realize the caller of `deprecate` can given a non-helpful `solution`, but it would at least encourage helpfulness.

technicalpickles:
> CI / test (3.1) (pull_request) Failing after 37s

I tried reproducing this locally, but at first it seemed to be passing. Then tried with `rake TESTOPTS="--seed=46039 "` and it fails consistently. It does not fail on main though.

mperham:
I put into `main` what should be the fix for that failure (global state not getting reset).

mperham:
I like what AS::Deprecation does in terms of documenting, in code, how the API should change so that log messages can be auto-generated with the calling site. Like you showed, we have 4-5 different usecases for deprecation. If we can build an API which supports those changes, I think we'll have something we can ship as Sidekiq::Deprecated in 7.1 and possibly extract it as a gem in 7.2. I have no problems borrowing liberally from AS::Deprecation's API to keep compatibility as much as possible and retain the MIT license.

technicalpickles:
Started https://github.com/mperham/sidekiq/pull/5751 to explore making it `Sidekiq::Deprecation`

lib/sidekiq/config.rb
frederikspang: I think adding `logger.warn("[Sidekiq]...")` (the [Sidekiq] prefix), would be kind of an antipattern, right? 

Wouldn't that (imho) belong in the logger itself? 
`logger = MyLoggerClass.new(prefix: '[Sidekiq'])`, instead of the outputter?

The output should be the raw message imo

technicalpickles:
Closing for now, moving discussion to https://github.com/sidekiq/sidekiq/pull/5745

PR #5751, opened 2023-01-27
Introduce Sidekiq::Deprecation as a way to mark deprecations

technicalpickles:
This is a different take on https://github.com/mperham/sidekiq/pull/5745 to address https://github.com/mperham/sidekiq/issues/5743 This borrows a ton from ActiveSupport::Deprecation.

For Sidekiq contributors, they can use like:

```ruby
Sidekiq::Deprecation.warn "this thing is changing, blah balh"
```

And Sidekiq extension developers can utilize it too:

```ruby
Sidekiq::Deprecation.warn "this other thing changes soon", gem_name: "sidekiq-awesomeness"
```

Either of these uses can also specify the deprecation horizon, when it is going away:

```
Sidekiq::Deprecation.warn "this thing is changing, blah balh", deprecation_horizon: "8.0"
```

Users of Sidekiq can configure the behavior when these come in. The default is so use Sidekiq.logger.warn. It takes a symbol for a name of a predefined logger (:log, :silence), or an object that responds to call (ie a lambda/proc/etc or your own class defining it):

```ruby
Sidekiq::Deprecation.behavior = proc {|msg, callstack, details|
  puts "#{details.gem_name} deprecation:"
  puts "for version #{details.deprecation_horizon}" if details.deprecation_horizon
  puts msg
  puts callstack
}
```

---

Some follow on things I'm still thinking about:

- the API is a little different than
- thinking to include a ActiveSupport::Deprecation behavior out of the box. That would just mean a development dependency on it to validate
- maybe Rails should default to that too?

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -28,6 +28,7 @@ rescue LoadError
 end
 
 require "sidekiq/config"
+require "sidekiq/deprecation"
 require "sidekiq/logger"
 require "sidekiq/client"
 require "sidekiq/transaction_aware_client"
@@ -101,6 +102,10 @@ module Sidekiq
   def self.freeze!
     @frozen = true
     @config_blocks = nil
+  end
+
+  def self.deprecate(msg)
+    default_configuration.handle_deprecation(msg, caller: caller)
   end
 
   # Creates a Sidekiq::Config instance that is more tuned for embedding
--- a/lib/sidekiq/client.rb
+++ b/lib/sidekiq/client.rb
@@ -44,7 +44,7 @@ module Sidekiq
     # @param chain [Sidekiq::Middleware::Chain] use the given middleware chain
     def initialize(*args, **kwargs)
       if args.size == 1 && kwargs.size == 0
-        warn "Sidekiq::Client.new(pool) is deprecated, please use Sidekiq::Client.new(pool: pool), #{caller(0..3)}"
+        Sidekiq::Deprecation.warn "Sidekiq::Client.new(pool) is deprecated, please use Sidekiq::Client.new(pool: pool)"
         # old calling method, accept 1 pool argument
         @redis_pool = args[0]
         @chain = Sidekiq.default_configuration.client_middleware
--- a/lib/sidekiq/deprecation.rb
+++ b/lib/sidekiq/deprecation.rb
@@ -1,0 +1,53 @@
+module Sidekiq
+  class Deprecation
+    class << self
+      attr_reader :behavior
+    end
+
+    DEFAULT_BEHAVIORS = {
+      log: ->(msg, callstack, details) {
+        logger = Sidekiq.default_configuration.logger
+        logger.warn "DEPRECATION(#{details}): #{msg} (called from #{callstack[0]})"
+      },
+
+      silence: ->(message, callstack, details) { },
+    }
+
+    def self.behavior=(behavior)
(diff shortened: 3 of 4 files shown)


mperham:
I don't like the notion of deprecations as a fancy logger. `S::D.warn` isn't valuable enough to continue down that path.

I like the notion of creating a Deprecation object for each change. There's a lot of functionality we can codify: explanations, timeframe, routing old style calls automatically to the new style API, usage metrics, etc.

```ruby
dep = ActiveSupport::Deprecation.new("7.0", "Sidekiq")
dep.deprecate_methods(Sidekiq.singleton_class,
        default_worker_options: :default_job_options,
        "default_worker_options=": :default_job_options=)
```

mperham:
I'm sorry if "fancy logger" was disrespectful or unfair.

The issue is that I really don't want to pull in a dependency on activesupport. That's a lot of code which some teams don't want in their app. I'd prefer to work with other gem authors to standardize some community-led API rather than bless ActiveSupport.

In other words, I don't see how to get what you want without more interest from the community. Should we start our own gem project and see if other gem makers want to sign onto an de facto community API for deprecations? Maybe we could provide a "compatibility" mode which plugs into AS::Deprecated (e.g. remove "AS::" to get our shim).

https://github.com/contribsys/deprecatedrb

What do you think? Do you want to help flesh out this project and gem?

technicalpickles:
> I don't see how to get what you want without more interest from the community

I think the original PR would be a middle ground. It doesn't do anything fancy, while providing an extension point. I could do a gem which hooks it up to ActiveSupport::Behavior.warn. It'd also be a chance to experiment with those more elaborate deprecation patterns it offers.

> Should we start our own gem project and see if other gem makers want to sign onto an de facto community API for deprecations? Maybe we could provide a "compatibility" mode which plugs into AS::Deprecated (e.g. remove "AS::" to get our shim).
>
> https://github.com/contribsys/deprecatedrb
>
>What do you think? Do you want to help flesh out this project and gem?

Sure 👍🏻 

I was chatting with @shageman about this, and we found https://github.com/cbeer/deprecation . Hilariously, it depends on active-support 😮‍💨 

This morning looking for other options, and found:

- [Gem::Deprecate](https://github.com/ruby/ruby/blob/master/lib/rubygems/deprecate.rb): rubygems' internal way of marking things deprecated. It is year based for some reason, rather than version. It's a module that can be extended, and can create method wrappers to mark something deprecated with a replacement
- [ruby-deprecate_public: Warn when calling private methods via public interface](https://github.com/jeremyevans/ruby-deprecate_public) by @jeremyevans. More narrowly defined for making public methods private over time

mperham:
Ok, it's good to see what's already out there. Are you driving this forward or are you expecting me to do so? What are the usecases? Is the plan to treat each of Sidekiq's recent deprecations as a usecase and develop an API to handle that deprecation? Or build a clean-room impl of AS::Deprecated that can be used by anyone?

technicalpickles:
I have been out sick nearly the entire time since I last posted so haven't gotten to anything yet. The first thing I planned to do is summarize comments from here and on https://github.com/mperham/sidekiq/pull/5745 onto the issue I started: https://github.com/mperham/sidekiq/issues/5743

technicalpickles:
Closing for now, moving discussion to https://github.com/sidekiq/sidekiq/pull/5745

PR #5765, opened 2023-02-04
Capsule default weights

Exterm1nate:
This PR fixes wrong weights value of Sidekiq process with implicit queues list.

If you do not provide `queues` in configuration explicitly, the default Sidekiq capsule initializes with `default` queue, but the `weights` stays `nil`, so the process's `weights` is `[nil]` instead of `[{"default"=>0}]` with explicit `queues` configuration. And it breaks Sidekiq web UI busy tab.

_Explicit_ queues list:
```ruby
# In config
sidekiq = Sidekiq.configure_embed do |config|
  config.queues = %w[default]
  config.concurrency = 2
end

# In Rails console
Sidekiq::ProcessSet.new.to_a
# => [#<Sidekiq::Process @attribs={..., "queues"=>["default"], "weights"=>[{"default"=>0}], ...}>, ...]
```

_Implicit_ queues list:
```ruby
# In config
sidekiq = Sidekiq.configure_embed do |config|
  config.concurrency = 2
end

# In Rails console
Sidekiq::ProcessSet.new.to_a
# => [#<Sidekiq::Process @attribs={..., "queues"=>["default"], "weights"=>[nil], ...}>, ...]
```

### DIFF
--- a/lib/sidekiq/capsule.rb
+++ b/lib/sidekiq/capsule.rb
@@ -28,6 +28,7 @@ module Sidekiq
       @name = name
       @config = config
       @queues = ["default"]
+      @weights = { "default" => 0 }
       @concurrency = config[:concurrency]
       @mode = :strict
     end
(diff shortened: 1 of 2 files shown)


mperham:
The default weight for any queue which is not given a weight should be 1, not 0?

mperham:
Oh never mind, that would turn strict priority into random weighting.

PR #5766, opened 2023-02-04
Adjust push_bulk return value

mperham:
Previously we only returned the JIDs that were succesfully pushed but there was no way for the caller to determine which payloads, if any were not successfully pushed.

Now we return both nils and JIDs. If you push 5 jobs, you will get 5 results even if all five did not get pushed. This change is helpful for the upcoming ActiveJob bulk enqueue helper method so AJ can know which jobs pushed successfully.

See rails/rails#46603

### DIFF
--- a/lib/sidekiq/client.rb
+++ b/lib/sidekiq/client.rb
@@ -128,10 +128,11 @@ module Sidekiq
           copy
         end
         result || nil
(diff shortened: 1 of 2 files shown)


mperham:
This is expected with Sidekiq 7.1.

PR #5787, opened 2023-02-14
Fix detecting json unsafe item within job arguments

fatkodima:
I apologize, I realized that I made a mistake in https://github.com/sidekiq/sidekiq/pull/5786.

Code like 
```ruby
when Array
  item.find { |e| !json_unsafe_item(e).nil? }
```
incorrectly returns *the child* of an array if it is not valid or has invalid children, instead of the innermost child - see the updated test case. The code became a little more uglier, but this time should work as expected.

### DIFF
--- a/lib/sidekiq/job_util.rb
+++ b/lib/sidekiq/job_util.rb
@@ -98,12 +98,19 @@ module Sidekiq
       when String, Integer, Float, TrueClass, FalseClass, NilClass
         nil
       when Array
-        item.find { |e| !json_unsafe_item(e).nil? }
+        item.each do |e|
+          unsafe_item = json_unsafe_item(e)
+          return unsafe_item unless unsafe_item.nil?
+        end
+        nil
       when Hash
         item.each do |k, v|
           return k unless String === k
-          return v unless json_unsafe_item(v).nil?
+
+          unsafe_item = json_unsafe_item(v)
+          return unsafe_item unless unsafe_item.nil?
         end
+        nil
       else
         item
       end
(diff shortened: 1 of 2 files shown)


mperham:
Can you use detect if you remove the !nil? logic?

fatkodima:
Did you mean

```ruby
when Array
  item.detect { |e| json_unsafe_item(e) }
```
?

It will return an array element. But if we have something like `[{"foo" => 1, :bar => 2}]` - we want `:bar` to be returned, but will get the whole hash.

jeremyw:
7.0.4 broke our ability to push a job with a nested hash (just strings and integers). Seems pretty critical to fix or yank that version.

fatkodima:
Can you provide an example of such a hash?

jeremyw:
Correction: our nested hash has boolean values. That may be key. Essentially the payload looks like this:

```
{
  "responder_uuid": "uuid",
  "consents": {
    "terms_and_conditions": true,
    "privacy_policy": true
  }
}
```

fatkodima:
Seems like you defined the hash incorrectly. Should be (note the `=>`):
```ruby
{
  "responder_uuid" => "uuid",
  "consents" => {
    "terms_and_conditions" => true,
    "privacy_policy" => true
  }
}
```

jeremyw:
That's JSON @fatkodima

fatkodima:
> 7.0.4 broke our ability to push a job with a nested hash (just strings and integers)

> Can you provide an example of such a hash?

I mean, can you provide an example of a ruby hash which started raising?

jeremyw:
@fatkodima I think the issue is when the job args are actually a `ActiveSupport::HashWithIndifferentAccess`, not a standard ruby Hash.

mperham:
7.0.5 is now out with this enhancement.

PR #5790, opened 2023-02-15
Handle correctly subclasses when finding json unsafe item

### LINKED ISSUE #5789: Unsafe JSON arguments appear to be rebroken in 7.0.5
Ruby version: 3.2.1
Rails version: 7.0.4.2
Sidekiq / Pro / Enterprise version(s): 7.0.5 / 7.0.6 / 7.0.3

We are seeing jobs in CI fail in 7.0.5 that didn't fail in 7.0.3 (but did fail in 7.0.4) for flagged unsafe JSON arguments. Our job takes arguments like this:

```
perform_async(1, 2, {"description"=>"fun", "max_points"=>"50"})
```

Calling `Hash#stringify_keys` on the hash above produces an equivalent hash. In 7.0.4 we saw unexpected failures, but the 7.0.5 failure now says something new about `nil`:

```
ArgumentError:

Job arguments to GoogleClassroomWorkers::CreateLessonWorker must be native JSON types, but nil is a NilClass.
--
  | See https://github.com/sidekiq/sidekiq/wiki/Best-Practices.
  | To disable this error, add `Sidekiq.strict_args!(false)` to your initializer.
  | # /bundle/ruby/3.2.0/gems/sidekiq-7.0.5/lib/sidekiq/job_util.rb:33:in `verify_json'
  | # /bundle/ruby/3.2.0/gems/sidekiq-7.0.5/lib/sidekiq/client.rb:91:in `push'
  | # /bundle/ruby/3.2.0/gems/sidekiq-7.0.5/lib/sidekiq/job.rb:365:in `client_push'
  | # /bundle/ruby/3.2.0/gems/sidekiq-7.0.5/lib/sidekiq/job.rb:198:in `perform_async'
  | # /bundle/ruby/3.2.0/gems/sidekiq-7.0.5/lib/sidekiq/job.rb:290:in `perform_async'
  | # ./app/controllers/api/v1/google_classrooms/lessons_controller.rb:12:in `block in create'
```

The same failure in 7.0.4 looked liked this:

```
ArgumentError:
--
  | Job arguments to GoogleClassroomWorkers::CreateLessonWorker must be native JSON types, see https://github.com/sidekiq/sidekiq/wiki/Best-Practices.
  | To disable this error, add `Sidekiq.strict_args!(false)` to your initializer.
```

fatkodima:
Fixes #5789.

HWIA is correctly detected as unsafe, but there is a problem in detecting the culprit. We can not use a `case-when`, because it checks for instances of classes (*or subclasses*), which is not correct in this case 🤦 We need to do exact comparison.

Am I missing any other test cases, so this won't happen again? 🤔

### DIFF
--- a/lib/sidekiq/job_util.rb
+++ b/lib/sidekiq/job_util.rb
@@ -21,8 +21,7 @@ module Sidekiq
       mode = Sidekiq::Config::DEFAULTS[:on_complex_arguments]
 
       if mode == :raise || mode == :warn
-        unless json_safe?(args)
-          unsafe_item = json_unsafe_item(args)
+        if (unsafe_item = json_unsafe?(args))
           msg = <<~EOM
             Job arguments to #{job_class} must be native JSON types, but #{unsafe_item.inspect} is a #{unsafe_item.class}.
             See https://github.com/sidekiq/sidekiq/wiki/Best-Practices.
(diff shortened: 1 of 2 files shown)


lib/sidekiq/job_util.rb
```diff
@@ -94,16 +94,16 @@ def json_safe?(item)
     end
 
     def json_unsafe_item(item)
-      case item
-      when String, Integer, Float, TrueClass, FalseClass, NilClass
+      klass = item.class
+      if [String, Integer, Float, TrueClass, FalseClass, NilClass].include?(klass)
```
mperham, line 98: Promote this Array to a Constant otherwise it will allocate a new array for every call.

lib/sidekiq/job_util.rb
```diff
@@ -94,16 +94,16 @@ def json_safe?(item)
...
         nil
-      when Array
+      elsif klass == Array
         item.each do |e|
           unsafe_item = json_unsafe_item(e)
           return unsafe_item unless unsafe_item.nil?
         end
         nil
-      when Hash
+      elsif klass == Hash
```
mperham, line 106: Should we allow Hash and Array subclasses with the assumption they have the same behavior for serialization purposes? I think HWIA is going to be a big support issue otherwise.

lib/sidekiq/job_util.rb
fatkodima: I can do that, but this method will be called only on exceptional cases when we have an unsafe json. Is it worth it?

lib/sidekiq/job_util.rb
fatkodima: These classes are already rejected by `json_safe?`. Should we enable it there too?

If we have HWIA with symbol keys, they will be serialized as string keys (and not a HWIA in the `perform`, as the user may be expecting). The user will try to access them via a symbol key and get a nil. If we assume, that the user knows and remembers that hashes in `perform` always have string keys, then probably yes. Wdyt?

lib/sidekiq/job_util.rb
mperham: Oh I see, I forgot RECURSIVE_JSON_SAFE is separate. Can we combine the two codepaths? Can RECURSIVE_JSON_SAFE change to return the unsafe item, so `json_safe?` becomes `json_unsafe?`?

lib/sidekiq/job_util.rb
mperham: I believe HWIA stores keys internally as Strings so we're safe in that regard.

So we either fail in serializing or it will silently fail when they use Symbols to access hash values. Ugh. We have to fail early and loudly. Failing silently allows bugs to slip past. You're right, we can't allow HWIA.

fatkodima:
Combined two code paths.

lib/sidekiq/job_util.rb
```diff
@@ -70,50 +70,41 @@ def normalized_hash(item_class)
...
-    RECURSIVE_JSON_SAFE.default = ->(_val) { false }
-    RECURSIVE_JSON_SAFE.compare_by_identity
-    private_constant :RECURSIVE_JSON_SAFE
+    RECURSIVE_JSON_UNSAFE.default = ->(val) { val }
+    RECURSIVE_JSON_UNSAFE.compare_by_identity
+    private_constant :RECURSIVE_JSON_UNSAFE
 
     def json_safe?(item)
-      RECURSIVE_JSON_SAFE[item.class].call(item)
+      json_unsafe_item(item).nil?
```
mperham, line 103: I think you're calling `json_unsafe_item` twice when the payload is unsafe? If you switch this method to `json_unsafe?` and return the value rather than using `.nil?` you can avoid the second call.

fatkodima: Yes, thanks! Fixed.

fatkodima: Eh, 2 commits, please squash when merging.

mperham:
It looks like you've accidentally included some standardrb formatting changes. Can you force push a clean PR?

fatkodima:
Crap. Reverted that. I used to manually fixing rubocop offenses, but standardrb automatically corrects on `rake`.

mperham:
It was quite a journey but we got there. 😎👍🏼

PR #5794, opened 2023-02-16
Exception logging with Rails backtrace cleaner

### LINKED ISSUE #5589: Feature Suggestion: Exception logging with Rails backtrace cleaner
We have been running a small extension in our app that swaps the previous [Sidekiq::ExceptionHandler::Logger](https://github.com/mperham/sidekiq/blob/fc7b7d8532549553673d4e510894b10a063792b3/lib/sidekiq/exception_handler.rb#L7) with our own, so that we can apply the [Rails backtrace cleaner](https://api.rubyonrails.org/classes/ActiveSupport/BacktraceCleaner.html) to it and only log app specific lines.

This changed with Sidekiq v6.5.0 where [the default error handler was moved to a method in Sidekiq](https://github.com/mperham/sidekiq/commit/67daa7a408b214d593100f782271ed108686c147#diff-2eecd3565b38823c09513903afa6686627d3ba8d883d520d3a83db61375e5ddeR63-R68), as part of the preparation for changes to Sidekiq 7. We can still wrap that method and apply the cleaner on our side, but I wanted to reach out and ask if this is maybe something that'd be interesting to have implemented in Sidekiq directly?

My thinking is that Sidekiq could provide some sort of hook / config for backtrace cleaning (e.g. `config[:backtrace_cleaner]`, a proc that receives the backtrace), that could be applied to this exception logging, and perhaps also to the backtraces stored within the jobs (similar to https://github.com/mperham/sidekiq/issues/4451) if that'd be interesting too, and that within the [Rails context](https://github.com/mperham/sidekiq/blob/main/lib/sidekiq/rails.rb#L7) it could be set to the default `Rails.backtrace_cleaner` cleaner.

Would that be something of interest to have in Sidekiq itself? And I imagine this is something that'd be included only in the upcoming v7?

I'm happy to get a draft going, just wanted to start with some questions to gather initial interested and direction.

Thanks,
Carlos

fatkodima:
Closes #5589.

Backtrace cleaning applies to exception logging and backtraces stored within the jobs.

### DIFF
--- a/lib/sidekiq/config.rb
+++ b/lib/sidekiq/config.rb
@@ -30,7 +30,8 @@ module Sidekiq
       },
       dead_max_jobs: 10_000,
       dead_timeout_in_seconds: 180 * 24 * 60 * 60, # 6 months
-      reloader: proc { |&block| block.call }
+      reloader: proc { |&block| block.call },
+      backtrace_cleaner: ->(backtrace) { backtrace }
     }
 
     ERROR_HANDLER = ->(ex, ctx) {
@@ -38,7 +39,10 @@ module Sidekiq
       l = cfg.logger
       l.warn(Sidekiq.dump_json(ctx)) unless ctx.empty?
       l.warn("#{ex.class.name}: #{ex.message}")
-      l.warn(ex.backtrace.join("\n")) unless ex.backtrace.nil?
+      unless ex.backtrace.nil?
+        backtrace = cfg[:backtrace_cleaner].call(ex.backtrace)
+        l.warn(backtrace.join("\n"))
+      end
     }
 
     def initialize(options = {})
--- a/lib/sidekiq/job_retry.rb
+++ b/lib/sidekiq/job_retry.rb
@@ -71,6 +71,7 @@ module Sidekiq
     def initialize(capsule)
       @config = @capsule = capsule
       @max_retries = Sidekiq.default_configuration[:max_retries] || DEFAULT_MAX_RETRY_ATTEMPTS
+      @backtrace_cleaner = Sidekiq.default_configuration[:backtrace_cleaner]
     end
 
     # The global retry handler requires only the barest of data.
@@ -159,10 +160,11 @@ module Sidekiq
       end
 
       if msg["backtrace"]
+        backtrace = @backtrace_cleaner.call(exception.backtrace)
         lines = if msg["backtrace"] == true
-          exception.backtrace
+          backtrace
         else
-          exception.backtrace[0...msg["backtrace"].to_i]
+          backtrace[0...msg["backtrace"].to_i]
         end
 
         msg["error_backtrace"] = compress_backtrace(lines)
--- a/lib/sidekiq/rails.rb
+++ b/lib/sidekiq/rails.rb
@@ -50,6 +50,12 @@ module Sidekiq
       end
     end
 
+    initializer "sidekiq.backtrace_cleaner" do
+      Sidekiq.configure_server do |config|
+        config[:backtrace_cleaner] = ::Rails.backtrace_cleaner
+      end
+    end
+
     # This hook happens after all initializers are run, just before returning
     # from config/environment.rb back to sidekiq/cli.rb.
     #
@@ -59,5 +65,9 @@ module Sidekiq
         config[:reloader] = Sidekiq::Rails::Reloader.new
       end
     end
+  end
+
+  ActiveSupport::BacktraceCleaner.class_eval do
+    alias :call :clean
   end
 end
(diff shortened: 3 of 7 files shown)


lib/sidekiq/backtrace_cleaner.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module Sidekiq
+  # @api private
+  class BacktraceCleaner
```
mperham, line 5: Can we alias `clean` to `call` and use a no-op Proc by default so we don't need to create any classes for this?

fatkodima: But rails wants `clean` - https://api.rubyonrails.org/classes/ActiveSupport/BacktraceCleaner.html#method-i-clean
Or where did you mean to add an alias?

mperham: I mean monkeypatch AS::BacktraceCleaner to add an alias.

fatkodima: Added a monkey patch. This monkey patch can be even upstreamed, I think.

lib/sidekiq/rails.rb
```diff
@@ -60,4 +66,8 @@ def inspect
       end
     end
   end
+
+  ActiveSupport::BacktraceCleaner.class_eval do
```
mperham, line 70: Do you need to require this, just in case?

fatkodima: Seems like it is not necessary as it is always required. In newly created apps `rails` is required - https://github.com/sidekiq/sidekiq/blob/1813cbdbd5e2fff23bfed5488ebcf5c6eeebb2e2/myapp/config/application.rb#L3 and from there backtrace cleaner is loaded https://github.com/rails/rails/blob/9180e61802d1c025f15bf57db92f392400dce46a/railties/lib/rails.rb#L16

fatkodima: And even `rails` is already required from this file at the top.

mperham:
Can you rebase on 7-1?

fatkodima:
Rebased.

mperham:
Thank you! As you probably guessed, this will be in 7.1.

lib/sidekiq/rails.rb
carlosantoniodasilva: Instead of aliasing the method, maybe we could wrap setting the config into a proc? e.g.

```ruby
config[:backtrace_cleaner] = -> backtrace { ::Rails.backtrace_cleaner.clean(backtrace) }
```

lib/sidekiq/job_retry.rb
```diff
@@ -71,6 +71,7 @@ class Skip < Handled; end
     def initialize(capsule)
       @config = @capsule = capsule
       @max_retries = Sidekiq.default_configuration[:max_retries] || DEFAULT_MAX_RETRY_ATTEMPTS
+      @backtrace_cleaner = Sidekiq.default_configuration[:backtrace_cleaner]
```
carlosantoniodasilva, line 74: Nice, I hadn't gone this far into cleaning up the backtrace of jobs to store :+1:

lib/sidekiq/config.rb
```diff
@@ -30,15 +30,19 @@ class Config
...
     ERROR_HANDLER = ->(ex, ctx) {
       cfg = ctx[:_config] || Sidekiq.default_configuration
       l = cfg.logger
       l.warn(Sidekiq.dump_json(ctx)) unless ctx.empty?
       l.warn("#{ex.class.name}: #{ex.message}")
-      l.warn(ex.backtrace.join("\n")) unless ex.backtrace.nil?
+      unless ex.backtrace.nil?
+        backtrace = cfg[:backtrace_cleaner].call(ex.backtrace)
+        l.warn(backtrace.join("\n"))
+      end
```
carlosantoniodasilva, line 45: I was considering cleaning the backtrace when handling exceptions triggers an error too: https://github.com/fatkodima/sidekiq/blob/4a255eb6ff8299e0be1f3d13d46b9513e87aa6e8/lib/sidekiq/config.rb#L267-L270, would that make sense to do, or is it better to just let that log everything since it's likely an implementation issue there?

carlosantoniodasilva (commented):
Thank you @fatkodima! ❤️💜💛💚💙

lib/sidekiq/rails.rb
fatkodima: Yeah, that will work too 👍

carlosantoniodasilva: Cool, hopefully this should do it: https://github.com/sidekiq/sidekiq/pull/5796. Thanks again!

PR #5796, opened 2023-02-17
Wrap the Rails backtrace cleaner to avoid aliasing method

carlosantoniodasilva:
By wrapping the `Rails.backtrace_cleaner` with a proc/lambda and call `clean` ourselves, we can avoid the monkey-patch/extension to alias the method in the backtrace cleaner implementation so that it conforms to the `call` API.

Ref: https://github.com/sidekiq/sidekiq/pull/5794#discussion_r1109767123

/cc @fatkodima


fatkodima:
Lgtm, thanks 👍 
While I have rights to merge PRs on this repo, I am not confident to do this on someone-elses projects 😄

PR #5810, opened 2023-03-02
Fix uninitialized constant Sidekiq::Capsule (NameError) 

manastyretskyi:
Right now if you will try to access default capsule in the code like this `Sidekiq.default_configuration.default_capsule` you will get this error because `capsule` is not required in the config file
```irb
development> Sidekiq.default_configuration.default_capsule
/Users/myuser/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/sidekiq-7.0.6/lib/sidekiq/config.rb:107:in `block in capsule': uninitialized constant Sidekiq::Capsule (NameError)   

        cap = Sidekiq::Capsule.new(nm, self)
                     ^^^^^^^^^
```

### DIFF
--- a/lib/sidekiq/config.rb
+++ b/lib/sidekiq/config.rb
@@ -1,6 +1,7 @@
 require "forwardable"
 
 require "set"
+require "sidekiq/capsule"
 require "sidekiq/redis_connection"
 
 module Sidekiq


mperham:
That's correct. Capsules are only relevant inside Sidekiq processes so the code is not required by default. You need to put your code inside `configure_server`.

PR #5816, opened 2023-03-08
Fix docs for update to v7

moofkit:
Adds clarifications for updating to v7 about:
- custom logger formatter
- redis sentinel configuration

Issues:
- https://github.com/sidekiq/sidekiq/issues/5603

### DIFF
--- a/docs/7.0-API-Migration.md
+++ b/docs/7.0-API-Migration.md
@@ -47,8 +47,6 @@ Sidekiq.configure_server do |cfg|
 end
 ```
 
-Have more APIs which need a migration tip? Edit this file and open a PR!
-
 ## Redis sentinels
 
 Due to Sidekiq's use of the new redis-client gem for connections, the configuration for it has changed.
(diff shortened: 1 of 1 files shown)


docs/7.0-API-Migration.md
```diff
@@ -34,4 +47,36 @@ Sidekiq.configure_server do |cfg|
...
+      {host: "host2", port: 26379},
+      {host: "host3", port: 26379}
+    ]
+  }
+end
+
+# fixed - don't forget to fix connection settings for both client and server
+Sidekiq.configure_server do |cfg|
+  cfg.redis = {
+    url: "redis://myredis/0",
```
ivanovv, line 71: I don't think this line is correct since according to this https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/redis_client_adapter.rb#L83C8-L83C8

```ruby
opts.delete(:url) if opts.key?(:sentinels)
```

if we have `:sentinels` key, the `:url` key will be dropped silently

moofkit: @ivanovv nice catch!
Looks like it's unexpected and should be removed. I'll look into this, thanks!

PR #5825, opened 2023-03-10
Show ActionMailer params as the job arguments

cover:
[Rails 5.1](https://guiarails.com.br/5_1_release_notes.html#parameterized-mailers) added the ability to pass data to mailers as params, which is also the current recommended way in [the guides](https://guides.rubyonrails.org/action_mailer_basics.html#calling-the-mailer).

Most of the time, emails are delivered immediately, so they exit the queue pretty quick, but sometimes you want to schedule a sending in the future (with `deliver_later(wait: ...)`) and the job might stay around for a while.

It'd be helpful to be able to see the params that are associated to the job, as at the moment it only shows the args, which might be empty most of the time (if you use params instead).

This changes it to show both of them.

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -397,7 +397,7 @@ module Sidekiq
           job_args.drop(3)
         elsif (self["wrapped"] || args[0]) == "ActionMailer::MailDeliveryJob"
           # remove MailerClass, mailer_method and 'deliver_now'
-          job_args.drop(3).first["args"]
+          job_args.drop(3).first.values_at("params", "args")
         else
           job_args
         end
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -391,7 +391,7 @@ module Sidekiq
     def display_args
       # Unwrap known wrappers so they show up in a human-friendly manner in the Web UI
       @display_args ||= if klass == "ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper"
-        job_args = self["wrapped"] ? args[0]["arguments"] : []
+        job_args = self["wrapped"] ? deserialize_argument(args[0]["arguments"]) : []
         if (self["wrapped"] || args[0]) == "ActionMailer::DeliveryJob"
           # remove MailerClass, mailer_method and 'deliver_now'
           job_args.drop(3)
@@ -466,6 +466,29 @@ module Sidekiq
     end
 
     private
+
+    ACTIVE_JOB_PREFIX = "_aj_"
+    GLOBALID_KEY = "_aj_globalid"
+
+    def deserialize_argument(argument)
+      case argument
+      when Array
+        argument.map { |arg| deserialize_argument(arg) }
+      when Hash
+        if serialized_global_id?(argument)
+          argument[GLOBALID_KEY]
+        else
+          argument.transform_values { |v| deserialize_argument(v) }
+            .reject { |k, _| k.start_with?(ACTIVE_JOB_PREFIX) }
+        end
+      else
+        argument
+      end
+    end
+
+    def serialized_global_id?(hash)
+      hash.size == 1 && hash.include?(GLOBALID_KEY)
+    end
 
     def uncompress_backtrace(backtrace)
       decoded = Base64.decode64(backtrace)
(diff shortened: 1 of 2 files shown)


mperham:
Can you post a screenshot before/after?

cover:
Sure, here the before:
Screenshot_20230311_013839

And the after:
Screenshot_20230311_014113

It's a very quick app with random mailers and models to show it out. `#comment` is receiving both params and an arg, `#like` is only receiving an arg, and `#reminder` is only receiving params.

To be honest, I'm not a big fan of having all those `_aj_` keys, which make it a little bit harder to see the interesting data, but I haven't clean them out as they aren't in the current args either.

mperham:
I suppose that’s an improvement. Does AJ have a helper method for getting or displaying job arguments?

cover:
I've checked its doc and code and [there is a method](https://api.rubyonrails.org/classes/ActiveJob/Arguments.html#method-i-deserialize) to deserialize the arguments, so you could potentially do:
```ruby
job = Sidekiq::ScheduledSet.new.first
ActiveJob::Arguments.deserialize(job.item['args'])
```

And that'd return them without their internal keys:
```ruby
{ ...
  "arguments"=>
   ["ActivityMailer",
    "comment",
    "deliver_now",
    {:params=>
      {:user=>
        #<User:0x00007fcdfd9c8fd8
         id: 1,
         email: "email@example.com",
         created_at: Sat, 11 Mar 2023 00:34:30.785471000 UTC +00:00,
         updated_at: Sat, 11 Mar 2023 00:34:30.785471000 UTC +00:00>},
     :args=>[{:include_history=>true}]}],
... }
```

But I don't think you'd want to run that here as it'd likely cause some slowdowns on the page with many jobs, as it'd have to deserialize everything, and also run db queries as well to retrieve records through globalid. I couldn't find a method to just return them in a "pretty" way (ie. serialized and without internal keys).

For these specific args, another option could be to use `slice` instead of `values_at`, that way it'd show what those values are (eg. `{ 'params' => { ... }, 'args' => [ ... ] }`), instead of showing just the values (eg. `{ ... }, [ ... ]`), that'd make it a little bit clearer, even though with a bit more text on that page (this would also play well with the example below).

A further improvement would be to clean up the AJ args, but that wouldn't be just for ActionMailer jobs, but for any AJ job, as you'd get similar args even when using standard AJ jobs and using globalid, symbol arguments, etc. It might still incur some performance hit as it'd have to transform all args to remove them. So potentially for a separate PR/work, what do you think?

Ideally, it'd be nice to show something like:
```ruby
{"params"=>{"user"=>{"_aj_globalid"=>"gid://test-app/User/1"}, "project_id"=>42, "_aj_symbol_keys"=>["user", "project_id"]},
 "args"=>
  [42,
   {"_aj_globalid"=>"gid://test-app/Project/1"},
   {"user"=>{"_aj_globalid"=>"gid://test-app/User/3"}, "include_history"=>true, "_aj_ruby2_keywords"=>["user", "include_history"]}]}
```
as:
```ruby
{"params"=>{"user"=>"gid://test-app/User/1", "project_id"=>42},
 "args"=>[42, "gid://test-app/Project/1", {"user"=>"gid://test-app/User/3", "include_history"=>true}]}
```

mperham:
I'm not sure we want to use `deserialize`; I think that touches the database and I don't want Sidekiq's Web UI to do that. Can we write a simpler `deserialize` which does not touch the database?

cover:
I agree; that was also my point, to not use it as it'd call the db as well (ie. to fetch records through globalid).

Yes, it'd totally be possible to write some custom code for that. How would you feel about having this PR just to add the ActionMailer params, and then a second PR to make the AJ data prettier? In that case I could update this PR to maybe also add the extra `params`/`args` keys I mentioned above, to make it more clear what they are.

And then I'll look into the other improvement (it might take a little bit as I'll be traveling in the next few weeks)

cover:
Alright, I went ahead and actually made that change :slightly_smiling_face: 

Before/current:
Screenshot_20230313_032527

After:
Screenshot_20230313_033004

It's now deserializing by only removing internal AJ keys, and compacting global id ones to just show the value and not the entire hash. I find the result a bit more clear to read what's going on, with less internal AJ stuff; but I might be totally biased on that.

I've kept `values_at` instead of `slice` as I see that there is [a check](https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/web/helpers.rb#L233) to make sure the value is an Array (and slice would return a Hash, and show that message instead). I haven't checked the reason behind that (it's 3am, lol)

mperham:
That looks a ton better. Well done!

PR #5827, opened 2023-03-11
`Sidekiq::Client.push_bulk` works in batches

### LINKED ISSUE #5814: Sidekiq::Client.push_bulk should work in batches
Sidekiq::Job.perform_bulk allows the `batch_size` parameter but `push_bulk` does not. Add support for a default batch_size of 1000.

NB: the term "batch_size" has nothing to do with Sidekiq Pro's Batches feature and is an unfortunate overlap in terminology. We could also call it "group size", "window size", "bulk size" or other?

fatkodima:
Fixes #5814.

Example of usage:
```ruby
push_bulk('class' => 'MyJob', 'args' => (1..100_000).to_a) # will push into redis in batches of size 1000
push_bulk('class' => 'MyJob', 'args' => (1..100_000).to_a, batch_size: 10_000)
```

[Nicer diff without whitespace changes](https://github.com/sidekiq/sidekiq/pull/5827/files?diff=split&w=1)

### DIFF
--- a/lib/sidekiq/client.rb
+++ b/lib/sidekiq/client.rb
@@ -96,8 +96,9 @@ module Sidekiq
 
     ##
     # Push a large number of jobs to Redis. This method cuts out the redis
-    # network round trip latency.  I wouldn't recommend pushing more than
-    # 1000 per call but YMMV based on network quality, size of job args, etc.
+    # network round trip latency. It pushes jobs in batches if more than
+    # `:batch_size` (1000 by default) of jobs are passed. I wouldn't recommend making `:batch_size`
+    # larger than 1000 but YMMV based on network quality, size of job args, etc.
     # A large number of jobs can cause a bit of Redis command processing latency.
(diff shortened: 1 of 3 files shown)


mperham:
I like that the argument hash is the job and the keyword hash is the method options.

Also, as a committer, I'd expect you to include a line in the changelog if the change is worthy of user attention. I think this change qualifies, I'm happy to do it if you are not confident in your English skills.

fatkodima:
Yes, this is for sure worth user's attention. I'm used to the fact that you write the changelog entries 😄 

> I'm happy to do it if you are not confident in your English skills
I would be thankful.

mperham:
Sorry I forgot: I want to target this to 7.1 as 7.0 is closed to new features.

mperham:
I give up, keyword arguments are nasty if you're trying to add them to an API which already accepts a Hash.

PR #5877, opened 2023-04-14
redis 5

dpep:
I'm trying to upgrade to redis 5 but can't because sidekiq [pins](https://github.com/sidekiq/sidekiq/blob/6-x/sidekiq.gemspec#L25) the version.  The constraint was added [here](https://github.com/sidekiq/sidekiq/commit/d566154107e6543ccb42e0d2e7eddfb529f0933e) due to a compatibility issue.  Fortunately, [redis 5.0.1](https://github.com/redis/redis-rb/commit/7fbd60a0f16c6828f19a6a2824ffe155c83d3ec7) introduced a backwards compatibility fix.

I would love to upgrade to Sidekiq 7, but can't because of the [known issue](https://github.com/sidekiq/sidekiq/blob/main/docs/7.0-Upgrade.md#known-issues) of certain [redis cloud providers](https://github.com/sidekiq/sidekiq/issues/5594) lacking support.

Could we please relax the sidekiq constraint for redis so folks can upgrade the two libraries independently?  🙏🏻

### DIFF
--- a/sidekiq.gemspec
+++ b/sidekiq.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |gem|
     "source_code_uri" => "https://github.com/mperham/sidekiq"
   }
 
-  gem.add_dependency "redis", ["<5", ">= 4.5.0"]
+  gem.add_dependency "redis", [">= 4.5.0", "!= 5.0.0"]
   gem.add_dependency "connection_pool", ["<3", ">= 2.2.5"]
   gem.add_dependency "rack", "~> 2.0"
 end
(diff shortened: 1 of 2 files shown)


dpep:
shout out to @casperisfine for [attempting this](https://github.com/sidekiq/sidekiq/pull/5489)

Do we need to add `opts[:driver] ||= Redis::Connection.drivers.last || "ruby"` back into `lib/sidekiq/redis_connection.rb`?  (deleted [here](https://github.com/sidekiq/sidekiq/commit/d566154107e6543ccb42e0d2e7eddfb529f0933e#diff-96029d2f394258e62375fcff42f91cee34b5e5571bdb099c1f9d82104047c4acL49))

mperham:
This seems like a bad idea; allowing a completely new redis driver on a branch which is supposed to be stable could create a big support headache over the next year.

If you want Redis 5, upgrade to Sidekiq 7. If you are stuck on Redis Enterprise, wait 3 more months for their RESP3 support. I don’t want to risk Sidekiq 6 stability.

PR #5881, opened 2023-04-17
Use input attribute over button on busy view

shevaun:
Switch buttons to input attributes, so they get the JS listener that provides the confirmation dialog.

Fixes https://github.com/sidekiq/sidekiq/issues/5878

Screenshot 2023-04-18 at 11 35 20 AM

### DIFF
--- a/test/web_test.rb
+++ b/test/web_test.rb
@@ -83,6 +83,20 @@ describe Sidekiq::Web do
       assert_match(/WebJob/, last_response.body)
     end
 
+    it "can quiet all processes" do
+      identity = "identity"
+      signals_key = "#{identity}-signals"
+      @config.redis do |conn|
+        conn.sadd("processes", [identity])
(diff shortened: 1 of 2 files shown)


test/web_test.rb
```diff
@@ -83,6 +83,20 @@ def job_params(job, score)
       assert_match(/WebJob/, last_response.body)
     end
 
+    it "can quiet all processes" do
```
mperham, line 86: This doesn't really test your change, yes? Were you including it for coverage?

mperham: I think I understand. 👍🏼

shevaun:
Yeah for coverage and to make sure the functionality still worked when the
input value changed from "1" to "Quiet All"

mperham:
Thanks for the fix!

test/web_test.rb
shevaun: (moving my reply inline:) Yeah for coverage and to make sure the functionality still worked when the
input value changed from "1" to "Quiet All"

PR #5904, opened 2023-05-03
Support multiple current attributes in middleware

manu3569:
Currently, when calling `Sidekiq::CurrentAttributes.persist(...)` for multiple current attributes, the middleware doesn't correctly load the serialized data in the server middleware. In fact, it fails because the attributes from the first current attribute are used for "rehydrating" the second one.

The backwards compatible solution in this PR allows for a hash argument to be passed to the aforementioned method, with a key that is used in the `jobs` hash.

```ruby
Sidekiq::CurrentAttributes.persist({
  cattr: "Myapp::Current",
  cattr_other: "Myapp::OtherCurrent"
})
```

### DIFF
--- a/lib/sidekiq/middleware/current_attributes.rb
+++ b/lib/sidekiq/middleware/current_attributes.rb
@@ -7,26 +7,32 @@ module Sidekiq
   # This can be useful for multi-tenancy, i18n locale, timezone, any implicit
   # per-request attribute. See +ActiveSupport::CurrentAttributes+.
   #
+  # For multiple current attributes, pass an array of current attributes.
+  #
   # @example
   #
   #   # in your initializer
   #   require "sidekiq/middleware/current_attributes"
   #   Sidekiq::CurrentAttributes.persist("Myapp::Current")
+  #   # or multiple current attributes
+  #   Sidekiq::CurrentAttributes.persist(["Myapp::Current", "Myapp::OtherCurrent"])
   #
   module CurrentAttributes
     class Save
       include Sidekiq::ClientMiddleware
 
-      def initialize(cattr)
-        @strklass = cattr
+      def initialize(cattrs)
+        @cattrs = cattrs
       end
 
       def call(_, job, _, _)
-        if !job.has_key?("cattr")
-          attrs = @strklass.constantize.attributes
-          # Retries can push the job N times, we don't
-          # want retries to reset cattr. #5692, #5090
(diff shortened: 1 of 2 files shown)


mperham:
> when calling Sidekiq::CurrentAttributes.persist(...) for multiple current attributes

But why do you have multiple?

mperham:
And your suggestion supports 2. What if I want 3?

manu3569:
> > when calling Sidekiq::CurrentAttributes.persist(...) for multiple current attributes
> 
> But why do you have multiple?

In our case, we need to handle global state for two separate concerns, each with their own `CurrentAttribute` class.

----


> And your suggestion supports 2. What if I want 3?

The proposed solution supports an arbitrary number of `CurrentAttribute`s – one per key/value pair in the argument hash.

```ruby
Sidekiq::CurrentAttributes.persist({
  cattr: "Myapp::Current",
  cattr_other: "Myapp::OtherCurrent",
  cattr_three: "Myapp::ThirdCurrent",
  cattr_fourth: "Myapp::FourthCurrent",
  ...
})
```

mperham:
I understand your usecase. I don't like making the user configure the attribute name. I'm wondering if it would be simpler to do something like:

```ruby
Sidekiq::CurrentAttributes.persist("Myapp::Current", "Myapp::OtherCurrent")
```

The first would be stored in `cattr` so it is backwards compatible with jobs already persisted in Redis. The second and beyond would be stored at `cattr_1`, `cattr_2`, etc. Effectively zero-based array indexes. You could add a new CurrentAttribute instance to the list and existing jobs data will remain backwards compatible. You could not change the ordering of attributes without breaking existing data.

WDYT?

manu3569:
>  You could not change the ordering of attributes without breaking existing data.

That was the main reason I proposed the user defined key, but maybe documenting the caveat about changing order is good enough.

Do you prefer the suggested variable length method signature over accepting an array in the first position?  We currently have the signature below, and we would need to type check the last argument to see if it's a config of a class string.

```ruby
def persist(klass, config = Sidekiq.default_configuration)
```

Let me know how to proceed, and I'll make the changes. Thanks!

mperham:
I think a single klass_or_array argument would be my initial thought but push back if there’s a good reason to.

manu3569:
Made the suggested updates. Please, let me know if it's acceptable for merging now or if further tweaks are needed. Thanks!

PR #5926, opened 2023-05-19
Accept more enqueuing options for `Job.perform_bulk`

fatkodima:
It was not possible to use `perform_bulk`, when for example, you wanted to enqueue into the custom queue or specify some delay.

Now it is possible:
```ruby
MyJob.perform_bulk([[1], [2], [3]], "queue" => "other_queue")
```


mperham:
I'd prefer it to support the more standard convention of `MyJob.set(options).perform_bulk(args...)`

fatkodima:
Yes, you are right!

PR #5951, opened 2023-06-26
Made all cache-control headers lowercase.

mattalbright:
This avoids "Rack::Lint::LintError: uppercase character in header name: Cache-Control".


mperham:
Thank you!

PR #5975, opened 2023-07-18
Enhance metrics tab

kwugirl:
As discussed yesterday in https://github.com/sidekiq/sidekiq/issues/5974

I can't test the use of the `filtering` hook locally but I referenced the Scheduled and Retries tabs. It might look funny because of dropdown menu for selecting the time period, though?

I also looked into paginating the same way that those tabs do but the Metrics query looked different from how `Sidekiq::Paginator` works so I dropped that.

Screenshot 2023-07-18 at 11 19 46 AM

### DIFF
--- a/lib/sidekiq/metrics/query.rb
+++ b/lib/sidekiq/metrics/query.rb
@@ -20,7 +20,7 @@ module Sidekiq
       end
 
       # Get metric data for all jobs from the last hour
-      def top_jobs(minutes: 60)
+      def top_jobs(minutes: 60, substr: nil)
         result = Result.new
 
         time = @time
@@ -35,10 +35,12 @@ module Sidekiq
           end
         end
 
+        rx = /#{Regexp.escape(substr)}/i if substr
         time = @time
         redis_results.each do |hash|
           hash.each do |k, v|
             kls, metric = k.split("|")
+            next if rx && !kls.match?(rx)
             result.job_results[kls].add_metric metric, time, v.to_i
           end
           time -= 60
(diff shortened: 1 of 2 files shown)


mperham:
Filtering requires server-side endpoints defined in Sidekiq Pro (see lib/sidekiq/pro/web.rb) and so this PR is incomplete as is. I can look into completing it next month when I get back from vacation.

mperham:
I feel dumb but do you know the magical git incantation to push commits to your repo/branch in order to collaborate on this PR? I fail to see how to do it with GitHub's UI or git.

mperham:
I've determined that the size of this change is surprisingly large (need to integrate the filter input and period selector into one form) so I'm going to push this back to 7.2.0.

I'm also committing to make the filtering feature free in Sidekiq 8.0, promoting it from Sidekiq Pro to OSS. It will make OSS more useful for open source apps like Mastodon and maintenance easier for me at the same time.

kwugirl:
@mperham I think I needed to set you as a collaborator on my fork for you to push onto it directly? Or you would need to open a PR against my branch on my fork, and then I would merge those in and then the branch would be updated for this PR?

But that's really nice to hear that filtering will be a free OSS feature!

mperham:
I thought I read somewhere that GitHub is smart enough to allow owners of an upstream repo to push to PR branches in collaborating repos (i.e. yours). IOW, I should be able to push to your branch for this PR but I don't know the git incantation to do so.

mperham:
Boom, got it working by adding your remote `git remote add kwu git@github.com:kwugirl/sidekiq.git` and running `git push kwu enhance_metrics_tab`.

mperham:
Filtering is part of the 7.2.0 release.

PR #5994, opened 2023-08-04
Add support for sidekiq_retry_in and sidekiq_retries_exhausted_block for ActiveJob

pierrickouw:
Hello 👋 

Trying to make the `sidekiq_retries_exhausted` and `sidekiq_retry_in` works with ActiveJob.
There some issue describing this behavior: https://github.com/sidekiq/sidekiq/issues/4496, and some workaround existing (for `sidekiq_retries_exhausted`), but none (as far as I know) for `sidekiq_retry_in`.

It is also written in the Sidekiq Wiki as currently not supported: https://github.com/sidekiq/sidekiq/wiki/Active-Job#customizing-error-handling, but it would be a huge bonus for us

Open to suggestions, other idea I had was to modify the [ActiveJob Adapter](https://github.com/rails/rails/blob/main/activejob/lib/active_job/queue_adapters/sidekiq_adapter.rb) to keep it in sync and modify the Job Instance on the fly, but thought it would be better to keep the current support as much as possible inside Sidekiq. [(That what we did as a workaround in the meantime)](https://gist.github.com/pierrickouw/da5fb6bffda2b17ab50d427ade1beb83)

There might be better places aswell in Sidekiq codebase. 

Thanks!

### DIFF
--- a/test/retry_exhausted_test.rb
+++ b/test/retry_exhausted_test.rb
@@ -64,10 +64,6 @@ describe "sidekiq_retries_exhausted" do
     @old_worker ||= OldJob.new
   end
 
-  def wrapped_worker
-    @wrapped_worker ||= WrappedJob.new
-  end
-
   def handler
     @handler ||= Sidekiq::JobRetry.new(@config.default_capsule)
   end
@@ -160,7 +156,7 @@ describe "sidekiq_retries_exhausted" do
 
   it "supports wrapped jobs" do
     assert_raises RuntimeError do
-      handler.local(wrapped_worker, job("retry_count" => 0, "retry" => 1, "wrapped" => "WrappedJob"), "default") do
+      handler.local(WrappedJob.new, job("retry_count" => 0, "retry" => 1, "wrapped" => WrappedJob.to_s), "default") do
         raise "kerblammo!"
       end
     end
--- a/test/retry_exhausted_test.rb
+++ b/test/retry_exhausted_test.rb
@@ -29,6 +29,14 @@ class Foobar
   include Sidekiq::Job
 end
 
+class WrappedJob < ActiveJob::Base
+  class_attribute :exhausted_called
+
+  sidekiq_retries_exhausted do |job|
+    WrappedJob.exhausted_called = true
+  end
+end
+
 describe "sidekiq_retries_exhausted" do
   def cleanup
     [NewJob, OldJob].each do |worker_class|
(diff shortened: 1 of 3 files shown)


mperham:
IMO if you are using ActiveJobs, you should be using its retry subsystem and semantics. If you want Sidekiq::Job's retry subsystem, use Sidekiq::Job.

pierrickouw:
Thanks for your answer!

I kind of agree, but as there is already a support for simple things like `sidekiq_options`, it often make people confused on why it doesn't work (and it is silent) when they are going through the documentation. As AJ is the default in Rails, many people goes this way aswell (I understand that Sidekiq is not intended _for_ Rails). The usage is a bit different (like the global identifier), so the migration from one system to the other is not straightforward (and addition like this one would certainly helps people move to a full Sidekiq implementation and benefits the performance upgrade once they are ready, cause the retry will already have the proper mechanism and semantics - _at least it would help us for sure, that's why we did the monkey patch_)

Retry system in AJ is very basic, and does not have the richness of Sidekiq's one, so it helps bridge the gap. (but no problem if you think it doesn't bring any values / is not aligned with Sidekiq philosophy!)

mperham:
I'm reconsidering this. Can you add a test for the exhausted block, because you have a bug as is. 😃

pierrickouw:
Ok sure, will do tomorrow! Thanks!

pierrickouw:
@mperham Fixed it and added a test!

test/retry_exhausted_test.rb
```diff
@@ -55,6 +64,10 @@ def old_worker
     @old_worker ||= OldJob.new
   end
 
+  def wrapped_worker
```
mperham, line 67: Do you need this?

test/retry_test.rb
```diff
@@ -381,6 +387,12 @@ def job
         end
         assert_equal 1, ds.size
       end
+
+      it "supports wrapped jobs" do
+        strat, count = handler.__send__(:delay_for, ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper, 2, StandardError.new, {"wrapped" => "ActiveJobRetry"})
```
mperham, line 392: Use `ActiveJobRetry.to_s` instead so the code works if we refactor the class name.

mperham (changes requested):
Minor nits. Almost there!

pierrickouw:
Should be good, thanks!

test/retry_exhausted_test.rb
pierrickouw: Not really, can remove it

mperham (approved):

mperham:
Thank you, will be available in 7.1.3.

PR #5999, opened 2023-08-10
Restore compatibility with Rack 2

### LINKED ISSUE #5988: Downcasing Rack response headers incompatible with Rack 2
Ruby version: any
Rails version: < 7.1
Sidekiq / Pro / Enterprise version(s): any including f97d16acad300f596c2ab4841553e4d8b7ff0116

See rails/rails#48815

The issue is that downcasing response headers on its own (while valid according to Rack 2 SPEC as written) is not really compatible with how people have been using Rack 2 headers in practice. In the linked issue, this leads to duplicate CSP headers (`Content-Security-Policy` and `content-security-policy`) in `Sidekiq::Web` when mounted in any released version of Rails with a CSP configured.

There's been some discussion in skipkayhil/tmp-rails#1 about the best way to support Rack 2 (in practice) and Rack 3 at the same time. The two approaches are:
- make headers conditional on Rack version
- wrap response headers in `Rack::Utils::HeaderHash`/`Rack::Headers` depending on Rack version

I'm happy to make a PR to do either of these, but figured I would make an issue to ask if you had a preference

skipkayhil:
Fixes #5988

Response headers were previously downcased in order to support Rack 3. However, this led to an issue where the content-security-policy header was not found by the Rails 7 (Rack 2) ContentSecurityPolicy middleware and the middleware would add its own (conflicting) header.

This commit restores compatibility with Rack 2 by making the casing of response headers conditional on the Rack version. A few of these are able to use the constants from Rack itself, and those not included in Rack were added as constants under Sidekiq::Web.

### DIFF
--- a/lib/sidekiq/web.rb
+++ b/lib/sidekiq/web.rb
@@ -34,6 +34,18 @@ module Sidekiq
       "Metrics" => "metrics"
     }
 
+    if Gem::Version.new(Rack::RELEASE) < Gem::Version.new("3")
+      CONTENT_LANGUAGE = "Content-Language"
+      CONTENT_SECURITY_POLICY = "Content-Security-Policy"
+      LOCATION = "Location"
+      X_CASCADE = "X-Cascade"
+    else
+      CONTENT_LANGUAGE = "content-language"
+      CONTENT_SECURITY_POLICY = "content-security-policy"
+      LOCATION = "location"
+      X_CASCADE = "x-cascade"
+    end
+
     class << self
       def settings
         self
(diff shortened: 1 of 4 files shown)


alissonbrunosa:
Hi 👋🏿 ,
Is there any chance of getting a version with this fix in?

mperham:
I’m hoping for a patch release next week. Still polishing it.

guigs:
@mperham Should this be backported to 7.0.x?

guigs:
Alright, I thought that security issues could be backported. We had an issue upgrading to 7.1, but I think it would be not so hard to fix.

PR #6017, opened 2023-08-31
Have `Thread.handle_interrupt` wrap the entire `ensure` block for job acknowledgement

bensheldon:
While reviewing different uses of `Thread.handle_interrupt`, I observed that Sidekiq's usage does not fully guarantee that job acknowledgement will happen once the `ensure` block is reached.  It's possible that an interrupt exception could be raised between `ensure` and `Thread.handle_interrupt`. Similar to the issue written about here: 

https://itnext.io/why-puma-workers-constantly-hung-and-how-we-fixed-by-discovering-the-bug-of-ruby-v2-5-8-and-v2-6-6-7fa0fd0a1958

This is arguably not a problem if the intent was only to guard the job acknowledgement once invoked but not ensure it takes place. I also allowed the `rescue` statements to be interruptible, but maybe they also should be moved into the non-interruptable block.

Hope this is helpful and I'm also open to the possibility I'm misunderstanding `Thread.handle_interrupt` 🙇🏻

### DIFF
--- a/lib/sidekiq/processor.rb
+++ b/lib/sidekiq/processor.rb
@@ -148,6 +148,8 @@ module Sidekiq
 
     IGNORE_SHUTDOWN_INTERRUPTS = {Sidekiq::Shutdown => :never}
     private_constant :IGNORE_SHUTDOWN_INTERRUPTS
+    ALLOW_SHUTDOWN_INTERRUPTS = {Sidekiq::Shutdown => :immediate}
+    private_constant :ALLOW_SHUTDOWN_INTERRUPTS
 
     def process(uow)
       jobstr = uow.job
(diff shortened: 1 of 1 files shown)


mperham:
This is the code I'm most scared to change because it is so subtle and tricky. 😬

Do we know if this is causing any issues or problems? Sidekiq does not guarantee exactly once execution so an un-acknowledged job can be rescued and re-run. Sounds like this issue, at worse, may lead to rare instances of duplicate job execution?

bensheldon:
@mperham no, no observed problems; this is solely theoretical Ruby-structural problems. The nature of the changes are a little narrower hiding the whitespace:

https://github.com/sidekiq/sidekiq/pull/6017/files?w=1

...but I also completely understand about the risk of changes here if it still doesn't change the fundamental contract of this code. No harm to close this PR 🙇🏻

mperham:
Thanks, Ben, good job!

PR #6035, opened 2023-09-12
Default retry_for to nil, not 0

afonsowaters:
Coercing to int is making it so whenever the arg is not passed in, it assumes the value 0, which prevents all jobs from ever retrying.

### DIFF
--- a/lib/sidekiq/job_util.rb
+++ b/lib/sidekiq/job_util.rb
@@ -55,7 +55,7 @@ module Sidekiq
       item["jid"] ||= SecureRandom.hex(12)
       item["class"] = item["class"].to_s
       item["queue"] = item["queue"].to_s
-      item["retry_for"] = item["retry_for"].to_i
+      item["retry_for"] = item["retry_for"].to_i if item["retry_for"]
       item["created_at"] ||= Time.now.to_f
       item
     end


mperham:
Uh, oops.

mperham:
7.1.4 is now available.

afonsowaters:
@mperham Thanks for the quick action! Do you have any suggestion how to handle the jobs that were already scheduled with the new version which will now be stored with a 0 on our Redis?

mperham:
I think you can use a server middleware to remove the key if the value is 0.

mperham:
Something like this:

```ruby
class CleanRetryFor
  def call(w, job, q, pool)
    job.delete("retry_for") if job["retry_for"] == 0
    yield
  end
end

  cfg.server_middleware.add CleanRetryFor
end
```

Not tested.

gmcabrita:
@mperham seems the middleware doesn't correctly clean up the `retry_for`, even setting it as both server and client middleware.

We opted for this temporary monkeypatch instead:

```ruby
module Sidekiq
  class << self
    alias_method :old_load_json, :load_json
  end

  def self.load_json(string)
    msg = old_load_json(string)
    msg.delete("retry_for") if msg["retry_for"] == 0
    msg
  end
end
```

This ensures https://github.com/sidekiq/sidekiq/blob/5203fc70e798cddc73cc36d7c5ee97982d8cd046/lib/sidekiq/job_retry.rb#L91-L93 and https://github.com/sidekiq/sidekiq/blob/5203fc70e798cddc73cc36d7c5ee97982d8cd046/lib/sidekiq/job_retry.rb#L124-L130 won't pass a `msg` to `process_retry` with `retry_for` set to `0`.

jjb:
I wonder why the rest of the sidekiq test suit, with no retry_for given, didn't have failures.

mperham:
Yeah, that's what caught me off-guard. We need a failing test to ensure this doesn't happen again.

PR #6056, opened 2023-09-26
Fix broadcasting sidekiq logger to rails logger for rails 7.1

### LINKED ISSUE #6054: `Logger.broadcast` removed from ActiveSupport causing sidekiq to break
Ruby version: 3.2
Rails version: master

#### The problem
When attempting to run Sidekiq with the latest Rails master branch, the following error is encountered:

```console
➜  core git:(main) ✗ bundle exec sidekiq
undefined method `broadcast' for ActiveSupport::Logger:Class
/Users/rasmus/.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/sidekiq-7.1.4/lib/sidekiq/rails.rb:52:in `block (2 levels) in <class:Rails>'
/Users/rasmus/.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/sidekiq-7.1.4/lib/sidekiq.rb:98:in `configure_server'
/Users/rasmus/.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/sidekiq-7.1.4/lib/sidekiq/rails.rb:47:in `block in <class:Rails>'
```

This issue seems to be related to changes in ActiveSupport, specifically the removal of the broadcast method in ActiveSupport::Logger. 

#### Steps to reproduce:

1. Use the latest Rails master branch.
2. Attempt to run Sidekiq.

#### Related PR:
See https://github.com/rails/rails/pull/48615

fatkodima:
Fixes #6054.

### DIFF
--- a/lib/sidekiq/rails.rb
+++ b/lib/sidekiq/rails.rb
@@ -45,7 +45,11 @@ module Sidekiq
         # it will appear in the Sidekiq console with all of the job context. See #5021 and
         # https://github.com/rails/rails/blob/b5f2b550f69a99336482739000c58e4e04e033aa/railties/lib/rails/commands/server/server_command.rb#L82-L84
         unless ::Rails.logger == config.logger || ::ActiveSupport::Logger.logger_outputs_to?(::Rails.logger, $stdout)
-          ::Rails.logger.extend(::ActiveSupport::Logger.broadcast(config.logger))
+          if ::Rails::VERSION::STRING >= "7.1"
+            ::Rails.logger = ActiveSupport::BroadcastLogger.new(::Rails.logger, config.logger)
+          else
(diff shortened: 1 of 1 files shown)


mperham:
I don't like Sidekiq explicitly overriding the `Rails.logger` instance with `logger=`. We don't own that attribute and it's VERY common for users to muck around with the Logger and we don't want to secretly undo their changes.

I've opened an issue, ideally Rails will make the logger instance a BroadcastLogger so we can add our logger to the broadcast set easily.

mperham:
I spent some time verifying this is the correct approach with Rafael and in doing so, reinvented your patch. So it landed without credit to you, sorry about that. Main should work now.

PR #6060, opened 2023-09-29
Change rails logger broadcasting configuration

fatkodima:
Rails actually improved configuring logging broadcasting - https://github.com/rails/rails/pull/49417.


mperham:
Yay, that's the feedback I gave when I initially made this change. Good.

PR #6067, opened 2023-10-05
Replace use of ActiveSupport::Logger.broadcast

thomasleese:
This has been removed in Rails 7.1 and replaced with the `ActiveSupport::BroadcastLogger` class. This has already been fixed for Sidekiq 7.x in #6060 but this commit fixes it for the older 6.x release of Sidekiq.

This builds on top of f67a7abccffc9337f144e7be96bb1ed4b0fee49a which did the same thing but didn't entirely remove all usage of `ActiveSupport::Logger.broadcast`.


mperham:
thanks, 6.5.11 is now out.

thomasleese:
Thanks!

PR #6071, opened 2023-10-10
Adds test and documents capsule configuration through the configuration file

hbontempo-br:
It's possible to configure capsules through the configuration file, but this feature nowhere in the documentation.
Documents and tests feature already implemented.


mperham:
This is on purpose. I don’t want to support that.

PR #6076, opened 2023-10-13
Allow nested Sidekiq::Testing

cjubb39:
Hello!

With the upgrade from version `7.1.1` -> `7.1.6`, I started seeing failures on having nested `Sidekiq::Testing.inline!` blocks.  The `inline` status seemed to be lost after the inner block exited.  This behavior change seemed to be introduced in e9f212902185591831af514df6313f8ae2cce31a.  I've included a test that fails on `main` but passes with my change.

I'm a bit of a Ruby noob so happy for whatever stylistic changes :)

### DIFF
--- a/lib/sidekiq/testing.rb
+++ b/lib/sidekiq/testing.rb
@@ -13,10 +13,10 @@ module Sidekiq
       def __set_test_mode(mode)
         if block_given?
           begin
-            self.__local_test_mode = mode
+            __local_test_mode_push(mode)
             yield
           ensure
-            self.__local_test_mode = nil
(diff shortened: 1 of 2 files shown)


mperham:
I don’t want to support reentrancy. Nesting fake and inline modes makes it hard to follow what the code will actually do.

cjubb39:
Fair enough.  In that case, are you open to a change that raises some sort of error on nesting?

mperham:
I think that’s reasonable.

cjubb39:
Closing in favor of #6078

PR #6078, opened 2023-10-13
Raise exception on nested Sidekiq::Testing modes

cjubb39:
Hello!

With the upgrade from version `7.1.1` -> `7.1.6`, I started seeing failures on having nested `Sidekiq::Testing.inline!` blocks.  This behavior change seemed to be introduced in e9f212902185591831af514df6313f8ae2cce31a.

As discussed on #6076, I've added code to raise an exception in this situation so it doesn't silently fail. I've included a test that fails on `main` but passes with my change.

I'm a bit of a Ruby noob so happy for whatever stylistic changes :)

### DIFF
--- a/lib/sidekiq/testing.rb
+++ b/lib/sidekiq/testing.rb
@@ -5,6 +5,7 @@ require "sidekiq"
 
 module Sidekiq
   class Testing
+    class TestModeAlreadySetError < RuntimeError; end
     class << self
       attr_accessor :__global_test_mode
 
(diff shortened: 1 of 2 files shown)


mperham:
Make the error message more specific: "You cannot nest test modes" or similar.

mperham:
I'll massage it a little bit but good job. 💪🏼😎

cjubb39:
No worries.  Thanks for merging!

PR #6079, opened 2023-10-17
feat: Notify strict args violation with Proc

ixti:
This way one can "track" all the "bad" places in production via bug tracking systems (like Sentry), or write those to logger.

### DIFF
--- a/lib/sidekiq/job_util.rb
+++ b/lib/sidekiq/job_util.rb
@@ -18,22 +18,18 @@ module Sidekiq
 
     def verify_json(item)
       job_class = item["wrapped"] || item["class"]
-      args = item["args"]
-      mode = Sidekiq::Config::DEFAULTS[:on_complex_arguments]
 
-      if mode == :raise || mode == :warn
-        if (unsafe_item = json_unsafe?(args))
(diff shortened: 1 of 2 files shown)


fatkodima:
Imo, strict args violations should be caught in development. Having those in production is an error. Don't think we need to have a special handling except the raising, as now. There is no point in silently logging it, or having a callback that reports it somewhere. 
Exception trace services should already report unhandled exceptions from sidekiq to their services (at least the ones I used before, bugsnag for example, already does this).

ixti:
@fatkodima I agree with that. But if you test suite is not 95% coverage and like 60% you end up with:

---

testing in production

---

Although, after opening the PR I just realized that probably it's not a good proposal indeed, LOL

fatkodima:
Even without good coverage, that error will be reported by the error reporting service automatically and just logging is probably 100% useless.

ixti:
> Even without good coverage, that error will be reported by the error reporting service automatically and just logging is probably 100% useless.

That's true. But if deployment takes 20 mins, cathing a failure in production that is facing 100% customers is not the most fun experience. :D

PR #6080, opened 2023-10-18
Metrics formatting.

benubois:
Hi @mperham,

I finally got around to updating to Sidekiq 7, great work!

I dig the new metrics tab, but it got me thinking about numbers in tables. The idea here is to make numbers more scannable by:

- Right aligning
- Using a monospace font
- Forcing precision (i.e `1.4 -> 1.40` so everything aligns within the column)
- Adding delimiters
- Moving units to the table header

If you prefer the way this looks, I could finish up this PR by formatting every numeric column the same way.

Here's a before and after example:

Table comparison

Thanks!

### DIFF
--- a/web/assets/javascripts/application.js
+++ b/web/assets/javascripts/application.js
@@ -33,6 +33,7 @@ function addListeners() {
 
   addShiftClickListeners()
   updateFuzzyTimes();
+  updateNumbers();
   setLivePollFromUrl();
 
   var buttons = document.querySelectorAll(".live-poll");
@@ -100,6 +101,20 @@ function updateFuzzyTimes() {
   var t = timeago()
   t.render(document.querySelectorAll('time'), locale);
   t.cancel();
+}
+
+function updateNumbers() {
+  document.querySelectorAll("[data-number-with-precision]").forEach(node => {
+    let number = parseInt(node.textContent);
+    let precision = parseInt(node.dataset["numberWithPrecision"] || 0);
+    if (typeof number === "number") {
+      let formatted = number.toLocaleString(undefined, {
+        minimumFractionDigits: precision,
+        maximumFractionDigits: precision,
+      });
+      node.textContent = formatted;
+    }
+  });
 }
 
 function setLivePollFromUrl() {
--- a/web/assets/javascripts/application.js
+++ b/web/assets/javascripts/application.js
@@ -104,9 +104,9 @@ function updateFuzzyTimes() {
 }
 
 function updateNumbers() {
-  document.querySelectorAll("[data-number-with-precision]").forEach(node => {
-    let number = parseInt(node.textContent);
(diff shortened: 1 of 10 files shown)


mperham:
Good to see you here again, Ben. 😎

Is number formatting something we should hand off to the browser JS, so it can do locale-specific formatting for the user?

web/assets/javascripts/application.js
```diff
@@ -102,6 +103,20 @@ function updateFuzzyTimes() {
...
+    if (typeof number === "number") {
+      let formatted = number.toLocaleString(undefined, {
+        minimumFractionDigits: precision,
+        maximumFractionDigits: precision,
+      });
+      node.textContent = formatted;
+    }
+  });
+}
+
```
benubois, line 119: Sure! I could tackle number localization too. What do you think about this as an API?

To format any number you would add a data attribute like:

```html
<samp data-number-with-precision="2">
```

mperham: I think my only nitpick is to use a much shorter descriptor. Use “data-nwp” or similar to keep page size down.

benubois:
Ok! Ready to review. I think every number should be getting formatted by `toLocaleString`, except for the Redis stats on the dashboard, since those come formatted from `Sidekiq::Stats`.

Scherm­afbeelding 2023-10-18 om 22 06 09

Also, I wasn't sure about modifying `number_with_delimiter`, as I don't know if anything else relies on it. Happy to make a new helper and leave `number_with_delimiter` alone.

web/assets/javascripts/dashboard-charts.js
```diff
@@ -86,6 +86,7 @@ class RealtimeChart extends DashboardChart {
     updateStatsSummary(this.stats.sidekiq);
     updateRedisStats(this.stats.redis);
     updateFooterUTCTime(this.stats.server_utc_time);
+    updateNumbers();
```
mperham, line 89: This is in application.js. Is it needed here?

web/assets/stylesheets/application.css
```diff
@@ -370,6 +370,15 @@ img.smallogo {
 .stat p {
   font-size: 0.9em;
 }
+
+.metric {
```
mperham, line 374: Is `fn` (formatted number) or `num` a better term since this class is used on non-metric pages?

web/assets/javascripts/dashboard-charts.js
benubois: Yeah, it has to run again after `updateStatsSummary` because the numbers get overwritten by polling mechanism.

web/assets/stylesheets/application.css
benubois: Sounds good!

benubois:
Deployed main to production. I like it!

Frame 1

PR #6082, opened 2023-10-21
CSP fix for metric period.

benubois:
I noticed that on /metrics, there's an `unsafe-inline` error triggered by the [new csp](#6074). This moves the functionality out of the page.


mperham:
Should that go in metrics.js?

benubois:
Yes! That makes sense.

PR #6091, opened 2023-10-31
Support :discard return value in sidekiq_retries_exhausted block

quainjn:
Provides the same functionality to `sidekiq_retries_exhausted` as `:discard` for `sidekiq_retry_in` of not sending these jobs to the dead set or calling global death handlers. This is useful for jobs that have retries disabled, which as a result means `sidekiq_retry_in` is never called.

Note: this is slightly different than the existing behavior of jobs with `dead: false`, in which case those jobs skip the dead set, but global death handlers are still called. Added tests to document/confirm that behavior, but didn't change it as part of this.

### DIFF
--- a/lib/sidekiq/job_retry.rb
+++ b/lib/sidekiq/job_retry.rb
@@ -226,7 +226,7 @@ module Sidekiq
     end
 
     def retries_exhausted(jobinst, msg, exception)
-      begin
+      rv = begin
         block = jobinst&.sidekiq_retries_exhausted_block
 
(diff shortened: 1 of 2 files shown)


mperham:
Thanks! This looks really simple and it seems like a good quality of life option for certain jobs. I just snuck it into today's 7.2.0 release.

henrik:
Took us a while to realise this was supported – would be good if it this was made clearer in https://github.com/sidekiq/sidekiq/wiki/Error-Handling.

mperham:
Thanks, done.

PR #6097, opened 2023-11-05
fix: reuse Sidekiq redis pool with the user defined options

rainerborene:
If you don't reuse the Sidekiq pool you must set it up on your `sidekiq` initializer like this:

```ruby
Sidekiq::Deploy.new(Sidekiq.redis_pool).mark! IO.read("#{Rails.root}/.git_sha")
```

Otherwise, you will receive a connection error in case you do not use the default `REDIS_URL` environment config.


mperham:
`sidekiq/deploy` is designed to work without needing to call `require 'sidekiq'`. Your change requires 'sidekiq'.

PR #6102, opened 2023-11-09
Proposed fix for minor memory leak in tracking

i7an:
The affected code:
```ruby
array = [..., @grams]
@grams = Hash.new { "this blocks captures `array` and previous `@grams` value" }
```

What is going on:
* `@grams` block  captures `array` variable.
* On the next iteration, when reset is called again, new `@grams` block captures previous variables, which hold the initial variables
* ...
* As the result GC cannot collect "unused" variables

The leakage is pretty minor. Don't expect a visible impact.

I discovered this while experimenting with heap dumps as [this article](https://blog.skylight.io/hunting-for-leaks-in-ruby/) explains.


mperham:
Thanks, good catch!

PR #6121, opened 2023-11-30
Fix Sidekiq boolean methods

rubendinho:
The ruby-doc documentation says that methods like `Sidekiq.server?` return a boolean, but this is not the case, because they use `defined?` under the hood. 

```ruby
foo = "bar"
defined?(foo) # => true
defined?(bar) # => nil 
```

This means on a non-Sidekiq process, `Sidekiq.server?` will return nil.

This ensures that a boolean is always returned.

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -46,7 +46,7 @@ module Sidekiq
   end
 
   def self.server?
-    defined?(Sidekiq::CLI)
+    defined?(Sidekiq::CLI) || false
   end
 
   def self.load_json(string)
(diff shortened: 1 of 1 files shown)


mperham:
It is perfectly idiomatic to return falsy values.

rubendinho:
Perhaps more elegant solution would be to just cast whatever the return value is as a boolean, so both strings and nils are always handled as expected.

rubendinho:
> I'm not sure where the return type in that RDoc came from; the code does not have any explicit RDoc for it. The API works fine if you use it in an idiomatic manner and at this point I would rather not change the impl.
> 
> Good: `if Sidekiq.server?` 
> Bad: `if Sidekiq.server? == true`

I ran this by a few experienced Ruby engineers and consistently heard that a predicate method like `Sidekiq.server?` would be expected to return `true` or `false`, not a string or `nil`. Off the top of my head, I can't think of other predicate methods in Ruby or Rails (except `defined?` ironically), that behave this way.

Finally I would argue that the current implementation goes against the "principle of least surprise". 

All this said, I understand that changing the return type for a core method like this could be a breaking change, but I hope you'll consider it for a future major version.

PR #6151, opened 2024-01-02
Remove dependency on base64 gem

### LINKED ISSUE #6150: Either declare or remove dependency on base64 to avoid a Ruby 3.3 warning
Ruby version: 3.3.0
Sidekiq / Pro / Enterprise version(s): master

Ruby 3.3 emits a warning when base64 is required without being specified in the gemspec. Ruby 3.4 will error.

There are a few places in sidekiq where base64 is being made use of. These should either be refactored to not use the gem, or it should be added to the gemspec.

You can observe this warning in CI: https://github.com/sidekiq/sidekiq/actions/runs/7358335463/job/20031428359
image
Ignore the other warnings, they originate from other gems. drb is from rails I believe.

I've gone through a few gems I use and made PRs to replace this gem with pack/unpack, since that is basically everything that it's doing. However, here the url safe variants are used which does a bit more than just calling a differently named method and inlining that here doesn't feel that great.
https://github.com/ruby/base64/blob/9669a7d3b0e3b9a739969404daf58f912c58c6b3/lib/base64.rb#L328-L333
https://github.com/ruby/base64/blob/9669a7d3b0e3b9a739969404daf58f912c58c6b3/lib/base64.rb#L351-L362

* https://github.com/rubocop/rubocop/pull/12313
* https://github.com/rack/rack/pull/2110
* https://github.com/newrelic/newrelic-ruby-agent/pull/2378
* https://github.com/lostisland/faraday/pull/1541

Earlopain:
This avoids a warning with Ruby 3.3

Closes #6150

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -4,7 +4,6 @@ require "sidekiq"
 
 require "zlib"
 require "set"
-require "base64"
 
 require "sidekiq/metrics/query"
 
@@ -491,8 +490,8 @@ module Sidekiq
     end
 
     def uncompress_backtrace(backtrace)
-      decoded = Base64.decode64(backtrace)
-      uncompressed = Zlib::Inflate.inflate(decoded)
+      strict_base64_decoded = backtrace.unpack1("m0")
+      uncompressed = Zlib::Inflate.inflate(strict_base64_decoded)
       Sidekiq.load_json(uncompressed)
     end
   end
--- a/lib/sidekiq/job_retry.rb
+++ b/lib/sidekiq/job_retry.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require "zlib"
-require "base64"
 require "sidekiq/component"
 
 module Sidekiq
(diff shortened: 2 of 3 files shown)


lib/sidekiq/api.rb
```diff
@@ -491,8 +490,8 @@ def serialized_global_id?(hash)
     end
 
     def uncompress_backtrace(backtrace)
-      decoded = Base64.decode64(backtrace)
-      uncompressed = Zlib::Inflate.inflate(decoded)
+      base64_decoded = backtrace.unpack1("m")
```
mperham, line 493: I think you should use `m0` for pack/unpack here. We don't want or need newlines.

Earlopain: Sure, changed. Tests are still passing

mperham:
Thanks!

dorianmarie:
@mperham would it be possible to release a new version so we don't get this warning on ruby 3.3.0?

dorianmarie:
Thanks a lot @mperham <3

PR #6152, opened 2024-01-02
Add new Sidekiq::Work type

### LINKED ISSUE #6145: Payload field inside work from Sidekiq::WorkSet is not hash anymore
Sidekiq version(s): 7.2.0

In accordance with your documentation  https://github.com/sidekiq/sidekiq/wiki/API#workers 
payload inside work should be hash
image

But in https://github.com/sidekiq/sidekiq/commit/ad9442f116086424948d60ec059755f9168a6bc8 it was changed to json.

`work` in Sidekiq 6.5.10
image

`work` in Sidekiq 7.2.0
image

mperham:
Today, the third parameter in `Sidekiq::WorkSet#each {|process_id, thread_id, work| ... }` is a raw Hash which makes it difficult to change data formats and storage. Introduce a new type `Sidekiq::Work` which represents a job in progress.

Sidekiq 7.0 introduced a slight change in the data format. It was optimized to reduce unnecessary JSON serialization of the "payload" attribute but this changed the data type.

```ruby
Sidekiq::WorkSet.new.each do |pid, tid, work|
  # old, <= 6.x
  work["payload"]["jid"]
  # old, 7.0 -> 7.2.0
  Sidekiq.load_json(work["payload"])["jid"]
  # new, main
  work["payload"]["jid"]
  work.job["jid"]
end
```

This restores compatibility with 6.x but does break compatibility with 7.0 -> 7.2. Any opinions?

See #6145

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -5,6 +5,8 @@
 HEAD
 ----------
 
+- Add `Sidekiq::Work` type which replaces the raw Hash in
+  `Sidekiq::WorkSet#each { |pid, tid, hash| ... }` [#6145]
 - Fix Ruby 3.3 warnings around `base64` gem [#6151, earlopain]
 
 7.2.0
--- a/Changes.md
+++ b/Changes.md
@@ -5,8 +5,14 @@
 HEAD
 ----------
 
-- Add `Sidekiq::Work` type which replaces the raw Hash in
+- Add `Sidekiq::Work` type which replaces the raw Hash as the third parameter in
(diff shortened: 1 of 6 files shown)


mperham:
I don't think it's a good idea to break compatibility with 7.x to restore compatibility with 6.x. Instead the new Sidekiq::Work type will provide accessor methods to get the Hash'd job data; I will deprecate direct access to all attributes in the Hash.

PR #6158, opened 2024-01-07
Add a default 5 connections to the capsule pool

stanhu:
In Sidekiq 6, there was one connection pool of size concurrency +
5. In Sidekiq 7, a capsule has its own connection pool of size concurrency, and there is an internal Sidekiq connection pool with a fixed size of 10.

If you have Sidekiq jobs that run and access `Sidekiq.redis`--which accesses the capsule's connection pool--users can run into an issue where a connection doesn't become available until a BRPOP returns from another job.

This commit adds 5 additional connections to the capsule pool, which is configurable via the `extra_connections` parameter. This should be a reasonable default that ensures Sidekiq 7 jobs avoid pool contention delays.

Relates to https://github.com/sidekiq/sidekiq/issues/5929

### DIFF
--- a/docs/capsule.md
+++ b/docs/capsule.md
@@ -81,6 +81,7 @@ Sidekiq.configure_server do |config|
   config.capsule("unsafe") do |capsule|
     capsule.queues = %w(thread_unsafe)
     capsule.concurrency = 1
+    capsule.extra_connections = 5
   end
 end
 ```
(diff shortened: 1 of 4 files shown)


mperham:
But ultimately, why? A Thread should never use more than one Redis connection at a time. If there are N threads and N connections, why would you need 5 more? This is a hack because we still don’t understand the root cause.

PR #6183, opened 2024-01-31
chore: mention Dragonfly in README

Niennienzz:
- Follow-up on this PR https://github.com/sidekiq/sidekiq/pull/6175
- Since we now have benchmark results and other details on the Wiki page, we just mention Dragonfly as an option here.


mperham:
250cc1e4dde44e84d9b607a2ded84ec0835a3ef1

PR #6191, opened 2024-02-08
Fix sidekiq.service for use as user service

Yegorov:
When rebooting the system, the user service was not launched automatically.

### DIFF
--- a/examples/systemd/sidekiq.service
+++ b/examples/systemd/sidekiq.service
@@ -91,3 +91,5 @@ SyslogIdentifier=sidekiq
 
 [Install]
 WantedBy=multi-user.target
+# Uncomment this and remove the line above if you are going to use this as a user service
+# WantedBy=default.target


mperham:
Please explain why.

Yegorov:
Because, for user service only available [`default.target`](https://www.freedesktop.org/software/systemd/man/latest/systemd.special.html#Special%20User%20Units)
I had a similar problem as [here](https://unix.stackexchange.com/questions/666509/systemd-service-does-not-start-wantedby-multi-user-target)

mperham:
Sidekiq is not typically deployed as a user-specific service. That .service file works for most Sidekiq users but if you deploy it differently, you must expect to adjust the .service file.

Yegorov:
I think, this case must be documented in the template so that other users are not surprised by this behavior
Besides, here it is already indicated how to make user service #4938

mperham:
We can document it as a comment but I don't think the actual configuration should change.

PR #6198, opened 2024-02-15
fix: Avoid JSON dump on Rails.application

ixti:
Rails enhances `Object` with `#as_json` that either uses `#to_hash`, or builds Hash of instance variables to their values. During boot process Sidekiq sends debug event with context that includes `:reloader`. As Reloader class has `@app = Rails.application` that causes stak level too deep one-off failure.

### DIFF
--- a/lib/sidekiq/rails.rb
+++ b/lib/sidekiq/rails.rb
@@ -20,6 +20,10 @@ module Sidekiq
       def inspect
         "#<Sidekiq::Rails::Reloader @app=#{@app.class.name}>"
       end
+
+      def to_hash
+        { app: @app.class.name }
+      end
     end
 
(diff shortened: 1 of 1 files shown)


mperham:
Can you point me to the Rails code that uses `to_hash`?

mperham:
https://github.com/rails/rails/blob/9e01d93547e2082e2e88472748baa0f9ea63c181/activesupport/lib/active_support/core_ext/object/json.rb#L60

mperham:
`to_hash` looks to be supported at least back to Rails 6.1.

ixti:
@mperham yeah, 6.0 also uses `to_hash`. My initial idea was to define `as_json` but digging through ActiveSupport - I believe `#to_hash` is the best one. :D

PR #6200, opened 2024-02-16
Allow user to disable default job logging

mperham:
This PR adds a new configuration knob, :skip_default_job_logging, to disable job logging out of the box, see #6199.

The user can choose when and where to disable the logging. I would recommend leaving it enabled for a reasonable developer experience but in environments like a high-volume production setup, logging output can be significant. This allows the user to opt-out of the logging if it becomes an issue.

```ruby
  # disable job logging in production
  config[:skip_default_job_logging] = (config[:environment] == "production")

  # disable sidekiq's job logging everywhere
  config[:skip_default_job_logging] = true
end
```

### DIFF
--- a/lib/sidekiq/job_logger.rb
+++ b/lib/sidekiq/job_logger.rb
@@ -2,23 +2,36 @@
 
 module Sidekiq
   class JobLogger
-    def initialize(logger)
+    include Sidekiq::Component
+
+    def initialize(config)
+      @config = config
       @logger = logger
     end
 
+    # If true we won't do any job logging out of the box.
+    # The user is responsible for any logging.
+    def skip_default_logging?
+      config[:skip_default_job_logging]
+    end
+
     def call(item, queue)
-      start = ::Process.clock_gettime(::Process::CLOCK_MONOTONIC)
-      @logger.info("start")
+      return yield if skip_default_logging?
 
-      yield
+      begin
+        start = ::Process.clock_gettime(::Process::CLOCK_MONOTONIC)
+        @logger.info("start")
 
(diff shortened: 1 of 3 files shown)


mperham:
This does change the parameters to JobLogger. Since it breaks an internal component, it'll need a minor version bump to 7.3.0.

sirwolfgang:
Is there any reason why the `config.skip_default_job_logging = true` syntax doesn't work here? Seems out of place with the rest of the config:
```ruby
  config.redis = Redis.config_sidekiq

  config.logger = Sidekiq::Logger.new($stdout)
  config.logger.level = ENV.fetch('SIDEKIQ_LOG_LEVEL', 'info')
  config[:skip_default_job_logging] = (config[:environment] == 'production')
  
  config.super_fetch!
  config.reliable_scheduler!
```

mperham:
Yes, it was not deemed important enough option to justify accessors on Sidekiq::Config.

ahmedsayedabdelsalam:
will this work with version 6.1.2 ?

adamdullenty:
Can this be specified per-worker via `sidekiq_options`? 

I've been looking at having this disabled globally by default in our app, and then enabling it for specific workers via `sidekiq_options skip_default_job_logging: false`

mperham:
I don't use the term `worker` anymore, it's ambiguous.

You can turn off INFO logging per job type like so:

`sidekiq_options log_level: :warn`

https://github.com/sidekiq/sidekiq/blob/bd3604909687bb0a0f4e71e087eb5a75e69fdf83/lib/sidekiq/job_logger.rb#L46-L49

PR #6212, opened 2024-03-04
Add ability to find a currently running work by jid

### LINKED ISSUE #6211: Ability to find a job by jid from WorkSet
Recently, we had to debug Sidekiq's Ent unique jobs and for one of the jobs we get a log message like `Skipping enqueue for SomeJob, not unique (lock held by JID somejid)`. We looked through all the queues (regular, `schedule` and `retry`) and the job was not in any of them. But fortunately, for all of that queues, API already has a `find_job` helper and it was a quick search. 
It turned out that the job was currently running, and we needed to quickly get its arguments. But it was not so quick, because there is no `find_job` helper for `WorkSet`.

Are you open to such an addition? I will help with the PR.

fatkodima:
Closes #6211.

Example:
```ruby
work = WorkSet.find_work_by_jid("abcde")
```


mperham:
Should be find_work and return a Sidekiq::Work, yes?

mperham:
Did you want to add a changelog entry and give yourself credit?

fatkodima:
Added a changelog entry.

PR #6217, opened 2024-03-07
Add language selector and session language persistance

FrancoARossi:
This pull request introduces a language selector to the footer of the web UI. Additionally, it ensures that the selected locale persists throughout the current session. If the user doesn't choose a language, it will default to their preferred one.

### Screenshots

**Desktop**
Screenshot 2024-03-08 at 19 56 02
Screenshot 2024-03-08 at 19 56 18
Screenshot 2024-03-08 at 19 56 33

**Mobile**
Screenshot 2024-03-08 at 19 57 46
Screenshot 2024-03-08 at 19 57 58

### DIFF
--- a/web/assets/stylesheets/application-rtl.css
+++ b/web/assets/stylesheets/application-rtl.css
@@ -155,11 +155,3 @@ div.interval-slider {
 .locale-select {
   float: right;
 }
-
-@media (max-width: 767px) {
-  .locale-select {
-    float: none;
-    line-height: 30px;
-    margin: 11px auto;
-  }
-}
--- a/web/assets/stylesheets/application-rtl.css
+++ b/web/assets/stylesheets/application-rtl.css
@@ -152,6 +152,12 @@ div.interval-slider {
   }
 }
 
-.locale-select {
+#locale-select {
   float: right;
+}
+
+@media (max-width: 767px) {
+  #locale-select {
+    float: none;
+  }
 }
--- a/web/assets/stylesheets/application-rtl.css
+++ b/web/assets/stylesheets/application-rtl.css
@@ -154,16 +154,6 @@ div.interval-slider {
 
 .locale-select {
   float: right;
-  color: #ddd;
-  width: 66px;
-  margin: 11px 15px;
-  padding: 4px;
-  border-radius: 4px;
-  background-color: transparent;
-  box-shadow: inset 0 1px 1px rgba(0,0,0,.075);
-  -webkit-box-shadow: inset 0 1px 1px rgba(0,0,0,.075);
-  transition: border-color ease-in-out .15s, -webkit-box-shadow ease-in-out .15s;
-  -webkit-transition: border-color ease-in-out .15s, -webkit-box-shadow ease-in-out .15s;
 }
 
 @media (max-width: 767px) {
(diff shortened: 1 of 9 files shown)


mperham:
But why? What’s the usecase?

FrancoARossi:
> But why? What’s the usecase?

I feel that enforcing the language based on the browser is somewhat restrictive. I though of doing this because a couple of days ago I was screen sharing to some co-workers from the US (me being from Argentina) to debug some failing jobs and it was a little awkward for them to see the UI in spanish without having the option to change it, I also would like to have it set to english as it's the language that I use daily while coding...

web/views/_footer.erb
```diff
@@ -15,7 +15,18 @@
             <p class="navbar-text"><a rel=help href="https://github.com/sidekiq/sidekiq/wiki">docs</a></p>
           </li>
           <li>
-            <p class="navbar-text"><a rel=external href="https://github.com/sidekiq/sidekiq/tree/main/web/locales"><%= locale %></a></p>
+            <form id="locale-form" action="<%= root_path %>change_locale" method="post">
+              <%= csrf_tag %>
+              <select id="locale-select" class="locale-select" name="locale">
+                <% available_locales.each do |locale_option| %>
+                  <% if locale_option === locale %>
```
mperham, line 22: Be careful. By using triple ===, I believe you may be risking a security hole, allowing anyone to inject any value for the `locale` parameter in the session. Thankfully you aren't rendering it.

web/assets/stylesheets/application-rtl.css
```diff
@@ -151,3 +151,25 @@ div.interval-slider {
     padding-left: 5px;
   }
 }
+
+.locale-select {
```
mperham, line 155: Can you DRY this up, especially since many rules overlap with application.css?

mperham:
Remove all of the explicit styling and add the bootstrap form styles. The dropdown should look like the other dropdown in the UI (the metrics time selector).

FrancoARossi:
> Remove all of the explicit styling and add the bootstrap form styles. The dropdown should look like the other dropdown in the UI (the metrics time selector).

I've applied the `.form-inline` and `.form-control` classes from Bootstrap, and I've also kept the bare minimum to align it properly with the other footer elements.

mperham:
The only obvious remaining issue is lack of dark mode support, we need a dark background for that form element. Our old Bootstrap CSS predates dark mode so I'm not sure if this can activate automatically. I've put in #6222 to upgrade to a modern Bootstrap.

web/views/_footer.erb
```diff
@@ -15,7 +15,19 @@
             <p class="navbar-text"><a rel=help href="https://github.com/sidekiq/sidekiq/wiki">docs</a></p>
           </li>
           <li>
-            <p class="navbar-text"><a rel=external href="https://github.com/sidekiq/sidekiq/tree/main/web/locales"><%= locale %></a></p>
+            <form id="locale-form" class="form-inline" action="<%= root_path %>change_locale" method="post">
+              <%= csrf_tag %>
+              <label class="sr-only" for="locale">Language</label>
```
mperham, line 20: Good work on this accessibility feature!

tierra:
In my instance, our footer is already quite wide, kicking this selector to the next line. I believe because of that, the footer is now larger height, causing some page content to be hidden behind the footer. This is especially problematic on pages with buttons on the bottom.

Note the slight window width changes, with both these being fully scrolled to bottom on the same page:

image

Slightly shorter, triggering the width break, and hiding the buttons:

image

PR #6225, opened 2024-03-12
Fix: Latency eq 0 is not a float

kayn1:
## What?

According to docs `latency` is 0 but looks like this is not true.
Zero is an exception there. 
Could we fix this? Tiny change but makes difference. :)

## Why?
```
[4] pry(main)> Sidekiq::Queue.all.last.latency.is_a?(Float)
=> false
```

`0-0` -> 0

Link:
https://www.rubydoc.info/github/mperham/sidekiq/Sidekiq/Queue#latency-instance_method


mperham:
What's wrong with that? Ruby is duck typed so it's normal for an API to return Float, Integer, etc.

PR #6231, opened 2024-03-16
Add `disable_retry_jitter` parameter to `sidekiq_options`

kuajia:
## Purpose of the Change

This pull request introduces a new parameter `disable_retry_jitter` to `sidekiq_options`, allowing users to disable the addition of random jitter to the retry delay.

## Details of the Change

- Introduced a new boolean parameter `disable_retry_jitter` in `sidekiq_options`
- The `process_retry` method now checks this parameter to determine whether to apply jitter to the `retry_at` calculation
- If `disable_retry_jitter` is true, no jitter is added; otherwise, the behavior remains as it was.

## Rationale

The addition of this parameter offers greater control over the retry mechanism, particularly in environments where predictable retry timing is essential (even if the retry mechanism is deliberately not designed to be precise). By allowing the disabling of retry jitter, users can ensure more deterministic retry intervals, which can be crucial for certain use cases or testing environments.

## Impact

This change is backward compatible and should not affect existing users who do not opt to use this new parameter. It introduces an additional feature that enhances configurability without altering the default behavior of Sidekiq.

### DIFF
--- a/lib/sidekiq/job_retry.rb
+++ b/lib/sidekiq/job_retry.rb
@@ -184,7 +184,8 @@ module Sidekiq
 
       # Logging here can break retries if the logging device raises ENOSPC #3979
       # logger.debug { "Failure! Retry #{count} in #{delay} seconds" }
-      jitter = rand(10) * (count + 1)
+      disable_rety_jitter = msg["disable_retry_jitter"].is_a?(TrueClass)
+      jitter = !disable_rety_jitter ? rand(10) * (count + 1) : 0
       retry_at = Time.now.to_f + delay + jitter
       payload = Sidekiq.dump_json(msg)
       redis do |conn|


mperham:
> particularly in environments where predictable retry timing is essential 

This has never been a goal of Sidekiq and still isn't. Sidekiq is not for "near real time" or synchronous functionality. I could be convinced to reduce the jitter but some jitter is required.

kuajia:
I agree with you, about the need of have some sort of jitter in order to avoid usless retry: if something is not responding now, probably won't respond anytime soon.

But I also think the user should be able to choose to use the retry system in the way which fits him best.
I don't know if you agree with the fact that some users may create Polling Jobs with Sidekiq, but in this case the developer may want to have control on the jitter, especially if the polling could continue for a long time (e.g. many days) with a little retry time (e.g. 1 minute).

You could introduce a 'max_jitter' option to cap the max jitter amount or a 'jitter_multiplier' to replace the default rand(10). This way, anyone using Sidekiq could tweak it to fit their needs exactly.

mperham:
Polling means you are trying to execute a loop over and over. That's not a Sidekiq job, jobs are designed to execute once to handle some event or workflow step.

You can implement polling badly with Sidekiq by executing a job and, upon success, have that job reschedule itself to run in N seconds, but that's not an officially supported usecase. Any network blip can lead to "breaking the chain" so the job doesn't get rescheduled and the loop stops.

For loops, I recommend people instead depend on minutely cron jobs. Every minute your cron job can check for necessary work to process and create child jobs to perform that work.

PR #6232, opened 2024-03-16
Add `disable_max_retries` option to `sidekiq_options` to Conditionally Disable Max Retries

kuajia:
## Purpose of the Change

This pull request adds a new `disable_max_retries` parameter to `sidekiq_options`, enabling users to disable the max retries limit under specific conditions. This feature requires both `disable_max_retries` and `retry_for` to be set, which ensures that it is used deliberately and reduces the risk of creating infinite job loops.

## Details of the Change

- A new boolean parameter `disable_max_retries` has been introduced, which works in conjunction with `retry_for`
- The job retry logic now checks both `retry_for` and `disable_max_retries` before deciding whether to exhaust retries
- This change allows jobs to bypass the max retry attempts limit if and only if `disable_max_retries` is true and `retry_for` is specified

## Rationale

This update gives users more flexibility in handling retries, particularly for jobs that might require special retry logic under certain conditions. By necessitating the concurrent setting of `retry_for`, it ensures that this feature is used judiciously, mitigating potential risks associated with disabling max retries. I think the current behavior of `retry_for` could be easily misunderstood due to the close, and in my opinion not very explicit, relationship that exists with the `retry` option.

## Impact

This enhancement is backward compatible and introduces a conditional feature that adds granularity to job retry management without affecting the existing functionality. It is especially useful in scenarios where advanced retry logic is necessary, providing users with the option to fine-tune job retries according to their specific requirements.

### DIFF
--- a/lib/sidekiq/job_retry.rb
+++ b/lib/sidekiq/job_retry.rb
@@ -169,7 +169,9 @@ module Sidekiq
         msg["error_backtrace"] = compress_backtrace(lines)
       end
 
-      return retries_exhausted(jobinst, msg, exception) if count >= max_retry_attempts
+      disable_max_retries = msg["retry_for"] && msg["disable_max_retries"].is_a?(TrueClass)
+
+      return retries_exhausted(jobinst, msg, exception) if !disable_max_retries && count >= max_retry_attempts
 
       rf = msg["retry_for"]
       return retries_exhausted(jobinst, msg, exception) if rf && ((msg["failed_at"] + rf) < Time.now.to_f)


mperham:
I don't like more feature flags. I don't really understand your Rationale (I agree about the weird relationship between `retry` and `retry_for` though), could you give me a clearer example of the behavior you are trying to fix?

kuajia:
Thanks for getting back to me, I really appreciate.

Let me break it down with an example that sheds light on what I'm trying to do. 
I've got a task that I want to keep trying for a set amount of time, like 2 days, or until something specific happens, like getting a certain response from an HTTP call. 
I figured out that I should tweak the 'retry' and 'retry_for' settings based on how often and how long I want to retry, and I adjusted these in my code as shown below.

```
class SampleJob
    include Sidekiq::Job

    RETRY_TIME = 2.days
    RETRY_INTERVAL = 1.minute

    sidekiq_options retry_for: RETRY_TIME,
                    retry: RETRY_TIME / RETRY_INTERVAL

    sidekiq_retry_in do |count|
        case exception
        when JOB_EXCEPTION::SoftRetry
            RETRY_INTERVAL.to_i
        else
            Sidekiq.logger.warn('unhandled exception :(')
            :discard
        end
    end

    def perform(token)
        status = an_http_call_which_return_a_status()

        case status
        when STATUS['ok']
            complete_the_job()
        else
            raise JOB_EXCEPTION::SoftRetry
        end
    end
end
```

Now, I might be off track here, and perhaps this isn't quite what you had in mind for using Sidekiq. 
It seems to me that there could be misleading when someone expects 'retry_for' to let the task run for the full time they've set. 
From my point of view, it looks like 'retry_for' is meant to keep the job active for the chosen period, but maybe that's not exactly right.

mperham:
I fundamentally disagree that a job should retry every minute for the next two days, ever. Exponential backoff is necessary; a failing job should not execute hundreds or even thousands of times. If you have exponential backoff, 25 retries becomes reasonable.

PR #6243, opened 2024-03-25
CSS refresh

mperham:
Upgrade Bootstrap and prune/simplify CSS rules where possible.

* [x] Upgrade to Bootstrap 3.4
* [ ] Migrate to Bootstrap 4.6 if it doesn't break RTL


mperham:
Bootstrap 4.5.3 with RTL support here

https://rtlcss.com/cdn/css-frameworks/bootstrap/
https://github.com/RTLCSS/bootstrap

mperham:
Aborting this. It's a quagmire of manual testing I don't want to get into. Nothing's actually broken today and old CSS has never been a source of security issues so only forward compatibility with future browsers is a concern; we can tackle those issues as necessary when they become apparent.

PR #6270, opened 2024-04-29
Tighten CSP around `script-src` and `style-src`

### LINKED ISSUE #6268: More strict CSP policy
Ruby version: 3.3.1
Rails version: 7.1.3.2
Sidekiq / Pro / Enterprise version(s): main

I saw the recent XSS and was a bit surprised since I remember protection against `unsafe-eval` for `script-src` being added in 7.2.0.

The current CSP declares `http:` and `https:` for a bunch of directives. I was tempted to open a PR to remove them since the UI just continues to work but there are two problems I see with that:
* Extensions that hook into the webui might cease to function
* Assets may be hosted outside of the main app, on something like S3 (https://github.com/sidekiq/sidekiq/pull/3913#issuecomment-410318634)

For the S3 case, I feel like the user should be able to configure the CSP in some way to add their bucket url to the CSP instead of just allowing everything. Similarly for addons which may use a cdn for js/css (though vendoring like here would be prefered I think).

Apart from benefitting sidekiq, extensions would also inherit these better protections.

Earlopain:
Instead of allowing every src, require scripts from unknown sources to have have a nonce. This makes it harder to exploit potential XSS vulnverabilities as the attacker needs to somehow know the nonce beforehand.

This also adds the nonce to all internal scripts. As per https://github.com/sidekiq/sidekiq/pull/3913#issuecomment-410318634 the assets may be hosted outside of the main app. Adding the nonce continues to support this usecase.

A nonce is incompatible with `unsafe-inline` (which only style has). Adapt the chart to not assign inline styles directly.

Extensions that load external scripts/styles must either vendor or add the nonce to their tags.

Closes #6268

----

Using nonces is the least disruptive change I believe. However, extensions that load resources from external sources (like from a CDN) will have the browser refuse to load them.

```erb
# Previous
<link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-QWTKZyjpPEjISv5WaRU9OFeRpok6YctnYmDr5pNlyT2bRjXh0JMhjY6hW+ALEwIH" crossorigin="anonymous">

# Newly required
<link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-QWTKZyjpPEjISv5WaRU9OFeRpok6YctnYmDr5pNlyT2bRjXh0JMhjY6hW+ALEwIH" crossorigin="anonymous" nonce="<%= csp_nonce %>">
```

Similarly, because nonces are incompatible with `unsafe-inline`, things like `<span style="background-color: red;"></span>` won't work anymore. You can look at the chart changes what this actually means since sidekiq itself requires a small change for this. It is very similar in scope to the changes about removing `unsafe-inline` from `script-src` which you did in the past.


mperham (changes requested):

lib/sidekiq/web.rb
```diff
@@ -5,11 +5,13 @@
 require "sidekiq"
 require "sidekiq/api"
 require "sidekiq/paginator"
+require "sidekiq/web/public_helpers"
```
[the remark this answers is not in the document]
mperham: Ah, poor wording. It's not a public API for applications and users. It's ok for web extensions to use them.

mperham:
Looks great. I'll need to do some work on Pro/Ent to make them compliant so give me a week or two to get this merged.

Earlopain:
Yeah, of course. If you have questions regarding CSP for that, let me know. While I'm no expert on the matter I do have the misfortune of knowing it more closely.

lib/sidekiq/web/application.rb
```diff
@@ -15,8 +15,8 @@ class WebApplication
       "manifest-src 'self'",
       "media-src 'self'",
       "object-src 'none'",
-      "script-src 'self' https: http:",
-      "style-src 'self' https: http: 'unsafe-inline'",
+      "script-src 'self' 'nonce-!placeholder!'",
+      "style-src 'self' 'nonce-!placeholder!'",
```
mperham, line 19: I'm wondering if we can allow `style-src 'unsafe-inline'` until Sidekiq 8.0. Most people can extract JavaScript pretty easily but a lot of dynamic CSS isn't as easy to extract. Are you aware of any issues or trade offs here?

Earlopain: `nonce` and `unsafe-inline` are not compatible with each other, when both are specified only `nonce` is respected, at least on modern browsers.

It seems fine to just undo the policy changes for `style-src` to allow for a smoother upgrade path, `script-src` is what this PR was all about. Loading arbitrary css is not exactly what you'd want to do but loading arbitrary scripts even less so.

Let me know what you think, I can make a PR for that. Or just do that yourself, either is fine for me.

mperham: Ok, we'll continue to implement nonce but need to allow unsafe-inline for styles for now. We can enable nonce and disable unsafe-inline in 8.0. Extensions can use 7.3 to implement nonce so they can remain compatible with 7.3 and 8.0. I don't want to require a hard break in compatibility between major versions if possible.

mperham: You're welcome to send a PR for that if you're in a good place to do so.

Earlopain: Yeah, sure. I will do so. I'm currently looking into something else but will probably have time to do that tomorrow

PR #6272, opened 2024-04-29
Allow indifferent access of strs to let locale be a symbol or string

mishaelajay:
Fix for: https://github.com/sidekiq/sidekiq/issues/6271


mperham:
Nope, that adds an ActiveSupport dependency to Sidekiq. Try another way or just call `to_s` on the value before lookup.

mishaelajay:
@mperham done! I went with a `to_s` only when fetching the locale from `session`

PR #6278, opened 2024-05-07
Convert Redis options nested keys to symbols

dmeremyanin:
[The docs says](https://github.com/sidekiq/sidekiq/wiki/Using-Redis#using-an-initializer)

> Keys can be strings or symbols and will be unified before being passed on to the Redis client.

But it doesn't work with nested keys in the options.

```ruby
  config.redis = {
    "name" => "mymaster",
    "sentinels" => [
      { "host" => "redis-node-1", "port" => 26379 },
      { "host" => "redis-node-2", "port" => 26379 },
      { "host" => "redis-node-3", "port" => 26379 }
    ]
  }
end
```

We're parsing the configuration from a YAML file and the hash comes out naturally with string keys. It would be nice to handle the conversion automatically.

### DIFF
--- a/lib/sidekiq/redis_connection.rb
+++ b/lib/sidekiq/redis_connection.rb
@@ -8,7 +8,7 @@ module Sidekiq
   module RedisConnection
     class << self
       def create(options = {})
-        symbolized_options = options.transform_keys(&:to_sym)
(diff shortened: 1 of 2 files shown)


mperham:
Nice work.

PR #6286, opened 2024-05-12
Add iteration support for long-running jobs

fatkodima:
This is a native implementation of the https://github.com/fatkodima/sidekiq-iteration pattern.
Closes https://github.com/fatkodima/sidekiq-iteration/issues/6.

At a high level, the API will look something like:
```ruby
class NotifyUsersJob
  include Sidekiq::Job
  include Sidekiq::Job::Iterable

  def build_enumerator(cursor:)
    active_record_records_enumerator(User.all, cursor: cursor)
  end

  def each_iteration(user)
    user.notify_about_something
  end
end
```

### DIFF
--- a/lib/sidekiq/job/iterable.rb
+++ b/lib/sidekiq/job/iterable.rb
@@ -1,0 +1,260 @@
+# frozen_string_literal: true
+
+require_relative "iterable/enumerators"
+
+module Sidekiq
+  module Job
+    module Iterable
+      class Interrupted < ::RuntimeError; end
+
+      include Enumerators
+
+      # @api private
+      def self.included(base)
+        base.extend(ClassMethods)
+      end
+
+      # @api private
+      module ClassMethods
+        def method_added(method_name)
+          if method_name == :perform
+            raise "Job that is using Iterable (#{self}) cannot redefine #perform"
+          end
+
+          super
+        end
+      end
+
+      # @api private
+      def initialize
+        super
+
+        @executions = 0
+        @cursor = nil
+        @times_interrupted = 0
+        @start_time = nil
+        @total_time = 0
+      end
+
+      # A hook to override that will be called when the job starts iterating.
+      #
+      # It is called only once, for the first time.
+      #
+      def on_start
+      end
+
+      # A hook to override that will be called around each iteration.
+      #
+      # Can be useful for some metrics collection, performance tracking etc.
+      #
+      def around_iteration
+        yield
+      end
+
+      # A hook to override that will be called when the job resumes iterating.
+      #
+      def on_resume
+      end
+
+      # A hook to override that will be called each time the job is interrupted.
+      #
+      # This can be due to throttling, `max_job_runtime` configuration,
+      # or sidekiq stopping.
+      #
+      def on_shutdown
+      end
+
+      # A hook to override that will be called when the job finished iterating.
+      #
+      def on_complete
+      end
+
+      # The enumerator to be iterated over.
+      #
+      # @return [Enumerator]
+      #
+      # @raise [NotImplementedError] with a message advising subclasses to
+      #     implement an override for this method.
+      #
+      def build_enumerator(*)
+        raise NotImplementedError, "#{self.class.name} must implement a '#build_enumerator' method"
+      end
+
+      # The action to be performed on each item from the enumerator.
+      #
+      # @return [void]
+      #
+      # @raise [NotImplementedError] with a message advising subclasses to
+      #     implement an override for this method.
+      #
+      def each_iteration(*)
+        raise NotImplementedError, "#{self.class.name} must implement an '#each_iteration' method"
+      end
+
+      # A hook to override that can be used to throttle a job when a given condition is met.
+      #
+      # If a job is throttled, it will be interrupted and retried after a backoff period
+      # (0 by default) has passed.
+      #
+      # The backoff can be configured via `sidekiq_options` per job:
+      #
+      #   sidekiq_options iteration: { retry_backoff: 30 } # in seconds
+      #
+      #  or globally:
+      #
+      #   Sidekiq::Config::DEFAULTS[:iteration][:retry_backoff] = 30 # in seconds
+      #
+      # @return [Boolean]
+      #
+      def throttle?
+        false
+      end
+
+      # @api private
+      def perform(*arguments)
+        fetch_previous_iteration_state
+
+        @executions += 1
+        @start_time = Time.now
+
+        enumerator = build_enumerator(*arguments, cursor: @cursor)
+        unless enumerator
+          logger.debug("'#build_enumerator' returned nil, skipping the job.")
+          return
+        end
+
+        assert_enumerator!(enumerator)
+
+        if @executions == 1
+          on_start
+        else
+          on_resume
+        end
+
+        completed = catch(:abort) do
+          iterate_with_enumerator(enumerator, arguments)
+        end
+
+        on_shutdown
+        completed = handle_completed(completed)
+
+        if completed
+          on_complete
+
+          logger.info {
+            format("Completed iterating. times_interrupted=%d total_time=%.3f", @times_interrupted, @total_time)
+          }
+        else
+          reenqueue_iteration_job
+        end
+      end
+
+      private
+
+      def fetch_previous_iteration_state
+        state = Sidekiq.redis { |conn| conn.hgetall("it-#{jid}") }
+
+        unless state.empty?
+          @executions = state["executions"].to_i
+          @cursor = Sidekiq.load_json(state["cursor"])
+          @times_interrupted = state["times_interrupted"].to_i
+          @total_time = state["total_time"].to_f
+        end
+      end
+
+      STATE_FLUSH_INTERVAL = 5 # seconds
+      STATE_TTL = 30 * 24 * 60 * 60 # 30 days
+
+      def iterate_with_enumerator(enumerator, arguments)
+        found_record = false
+        state_flushed_at = Time.now
+
+        enumerator.each do |object, cursor|
+          found_record = true
+          around_iteration do
+            each_iteration(object, *arguments)
+          end
+          @cursor = cursor
+          adjust_total_time
+
+          if Time.now - state_flushed_at >= STATE_FLUSH_INTERVAL || job_should_exit?
+            flush_state
+            state_flushed_at = Time.now
+          end
+
+          if job_should_exit?
+            return false
+          end
+        end
+
+        logger.debug("Enumerator found nothing to iterate!") unless found_record
+
+        true
+      end
+
+      def reenqueue_iteration_job
+        @times_interrupted += 1
+        flush_state
+        logger.debug { "Interrupting and re-enqueueing the job (cursor=#{cursor.inspect})" }
+
+        raise Interrupted
+      end
+
+      def adjust_total_time
+        @total_time += (Time.now - @start_time).round(3)
+      end
+
+      def assert_enumerator!(enum)
+        unless enum.is_a?(Enumerator)
+          raise ArgumentError, <<~MSG
+            #build_enumerator is expected to return Enumerator object, but returned #{enum.class}.
+            Example:
+              def build_enumerator(params, cursor:)
+                active_record_records_enumerator(
+                  Shop.find(params["shop_id"]).products,
+                  cursor: cursor
+                )
+              end
+          MSG
+        end
+      end
+
+      def job_should_exit?
+        max_job_runtime = self.class.get_sidekiq_options.dig("iteration", "max_job_runtime") ||
+          Sidekiq.default_configuration.dig(:iteration, :max_job_runtime)
+
+        ran_enough = max_job_runtime && @start_time && (Time.now - @start_time > max_job_runtime)
+        ran_enough || Sidekiq.stopping? || throttle?
+      end
+
+      def flush_state
+        key = "it-#{jid}"
+        state = {
+          "executions" => @executions,
+          "cursor" => Sidekiq.dump_json(@cursor),
+          "times_interrupted" => @times_interrupted,
+          "total_time" => @total_time
+        }
+
+        Sidekiq.redis do |conn|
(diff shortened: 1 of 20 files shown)


lib/sidekiq/job/iterable.rb
```diff
@@ -0,0 +1,245 @@
...
+              arguments.pop["sidekiq_iteration"]
+            else
+              {}
+            end
+
+          @executions = options["executions"] || 0
+          @cursor_position = options["cursor_position"]
+          @times_interrupted = options["times_interrupted"] || 0
+          @total_time = options["total_time"] || 0
+        end
```
fatkodima, line 115: The iterable job keeps its state inside the job hash metadata itself (as the last item of the args, if to be concrete). So, if the interrupted job is run again, for example, we restore its attributes from the previously stored metadata. Same when the job raises and error - we save the current state into the job and push it to redis.

Another approach is to keep the iteration state as a separate structure in redis (as a hash, e.g.). But then we need to decide on expiration time of that hash, decide if we should update it after each iteration (+1 redis call per iteration) or once in a while, write a server middleware to set job's attributes based on that hash.

Which one would you suggest/prefer?

lib/sidekiq/job/iterable.rb
```diff
@@ -0,0 +1,245 @@
...
+              "executions" => executions,
+              "cursor_position" => cursor_position,
+              "times_interrupted" => times_interrupted,
+              "total_time" => total_time
+            }
+          )
+
+          retry_backoff = self.class.get_sidekiq_options.dig("iteration", "retry_backoff") ||
+                          Sidekiq.default_configuration["retry_backoff"]
+          self.class.perform_in(retry_backoff, *arguments)
```
fatkodima, line 199: Is there a better way to reschedule itself than keeping arguments in the job and then explicitly calling `perform_in`?

For example, there is a helper in the ActiveJob for the similar purpose - https://api.rubyonrails.org/classes/ActiveJob/Exceptions.html#method-i-retry_job. Maybe we need something similar?

fatkodima:
@mperham Can you please answer questions, when you have time? 🙏

mperham:
This is a totally new feature to me and not something I've used before so please give me some time to study the code, understand what it is doing and understand how Sidekiq might need to change to make it work well. I apologize for the delay. Test cases would also be helpful to understand how the APIs will be used.

lib/sidekiq/job/iterable.rb
mperham: That's a really tough question. It's pretty fundamental to the design and I'm not sure I like the idea of mutating the job payload itself, as it's stored in Redis as a String which can't easily be modified without a JSON parse/dump round trip. I'm leaning toward defining a job iteration Hash, with a key like "it-{jid}". Redis has a number of useful H* commands to update/increment Hash values very quickly.

I think the default should be to update the iteration record in Redis only if the iteration is interrupted or 5 seconds have passed. It should not be updated after every record; if we are processing a million records we don't want a million Redis updates.

lib/sidekiq/job/iterable.rb
mperham: Good question, it might be time to consider a API for this purpose. Generally I don't like the idea of a job performing its own "meta-logic" around scheduling because it makes testing difficult. It's a massive violation of the Single Responsibility Principle. But this API wouldn't be used by the job itself, but rather your iteration logic.

lib/sidekiq/job/iterable.rb
```diff
@@ -0,0 +1,245 @@
...
+
+        def job_should_exit?
+          max_job_runtime = self.class.get_sidekiq_options.dig("iteration", "max_job_runtime") ||
+                            Sidekiq.default_configuration["iteration"]
+
+          return true if max_job_runtime && start_time && start_time.to_f + max_job_runtime > Time.now.to_f
+          return true if throttle?
+
+          # TODO: determine a better way to detect if the server/capsule is stopping.
+          Sidekiq.server? && Sidekiq::CLI.instance.launcher.stopping?
```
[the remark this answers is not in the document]
mperham: Nope, this is the best way to do it. I would use the global you set, rather than calling `launcher.stopping?`.

lib/sidekiq/job/iterable/active_record_enumerator.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module Sidekiq
+  module Job
+    module Iterable
+      # @private
+      class ActiveRecordEnumerator
```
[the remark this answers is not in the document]
mperham: Ideally you'll have a skeleton interface and people can develop their own Enumerators which work with a specific datastore. ActiveRecord is the primary datastore for 90% of Sidekiq apps though; it would make sense to support it well. I'm ok with a single column if it makes the code far simpler.

lib/sidekiq/job/iterable.rb
```diff
@@ -0,0 +1,261 @@
...
+      def reenqueue_iteration_job
+        @times_interrupted += 1
+        flush_state
+        logger.debug { "Interrupting and re-enqueueing the job (cursor=#{cursor.inspect})" }
+
+        retry_backoff = self.class.get_sidekiq_options.dig("iteration", "retry_backoff") ||
+          Sidekiq.default_configuration.dig("iteration", "retry_backoff")
+
+        # Preserve original jid.
+        self.class.set(jid: jid).perform_in(retry_backoff, *@arguments)
```
fatkodima, line 205: This will rerun client middleware again. Need to decide, if this is a problem or not.

lib/sidekiq/job/iterable.rb
```diff
@@ -0,0 +1,261 @@
...
+          }
+        else
+          reenqueue_iteration_job
+        end
+      end
+
+      private
+
+      def fetch_previous_iteration_state
+        state = Sidekiq.redis { |conn| conn.get("it-#{jid}") }
```
fatkodima, line 155: Should we handle errors here?

fatkodima:
@mperham This is ready now for review.

lib/sidekiq/job/iterable.rb
```diff
@@ -0,0 +1,261 @@
...
+
+        if state
+          state_hash = Sidekiq.load_json(state)
```
fatkodima, line 158: Should we handle errors here?

mperham: No, exceptional cases should generate exceptions. I do think you should use native Redis hashes instead of JSON, so you'd use `hgetall` rather than `get`.

lib/sidekiq/job/iterable.rb
```diff
@@ -0,0 +1,261 @@
...
+      def flush_state
+        state = {
+          "executions" => @executions,
+          "cursor" => @cursor,
+          "times_interrupted" => @times_interrupted,
+          "total_time" => @total_time
+        }
+
+        Sidekiq.redis do |conn|
+          conn.set("it-#{jid}", Sidekiq.dump_json(state), ex: STATE_TTL)
```
mperham, line 244: See `hmset` in Redis. https://redis.io/docs/latest/commands/hmset/

lib/sidekiq/job/iterable.rb
fatkodima: Yeah, I don't like the current approach too.
This is exactly how I implemented it initially, but in the added tests for this PR I now need to rescue this exception manually and do all the rescheduling logic for tests to pass. I will look at if I can somehow better solve this in tests.

lib/sidekiq/job/iterable.rb
fatkodima: We currently store the iteration cursor as part of this metadata. When using json, we get the correct data type when loading from json. But with redis hashes we will always get strings as values and the user will need to remember to manually cast the string value to the correct datatype. When people will use some custom enumerators or we start supporting custom columns for active record enumerator in the future, we will get other types of cursors, other than integers, and this will be a problem.

Wdyt?

mperham: You can make the cursor JSON if it is user-supplied but Sidekiq's internal iteration data can be stored element-by-element in the hash and coerced in Ruby. JSON is meant to be used for interoperability but for our internal data, I'd rather the data be directly addressable in Redis, rather than requiring deserialization.

fatkodima:
Updated with https://github.com/sidekiq/sidekiq/pull/6286#discussion_r1607155767 and converted to use redis hashes.

lib/sidekiq/job/iterable.rb
```diff
@@ -0,0 +1,260 @@
+# frozen_string_literal: true
+
+require_relative "iterable/enumerators"
+
+module Sidekiq
+  module Job
+    module Iterable
+      class Interrupted < ::RuntimeError; end
```
mperham, line 8: I think `Sidekiq::Job::Interrupted` would read better, wdyt?

lib/sidekiq/job/iterable.rb
```diff
@@ -0,0 +1,260 @@
...
+          if method_name == :perform
+            raise "Job that is using Iterable (#{self}) cannot redefine #perform"
+          end
+
+          super
+        end
+      end
+
+      # @api private
+      def initialize
```
mperham, line 29: This module should include a `stopping?` method which checks `@done` on Sidekiq::Processor. Here we use a :lifecycle accessor to store the Processor on the Job.

```ruby
# In Sidekiq::Processor...
job = SomeJob.new
job.lifecycle = self if job.is_a?(Sidekiq::Job::Iteratable) # Processor
job.perform(*args)
job.stopping? # => job.lifecycle.stopping?
```

It's a little more complex but I'm trying to avoid any more global APIs like `Sidekiq.stopping?` to be more Ractor-friendly.

lib/sidekiq/job/iterable/csv_enumerator.rb
```diff
@@ -0,0 +1,44 @@
...
+            .to_enum { (count_of_rows_in_file.to_f / batch_size).ceil }
+        end
+
+        private
+
+        def count_of_rows_in_file
+          filepath = @csv.path
+          return unless filepath
+
+          count = `wc -l < #{filepath}`.strip.to_i
```
mperham, line 37: This could be a security issue under the right circumstances. I think you can use `Shellwords.shellescape` for this filepath and I think you can use a simpler `wc -l filepath` without the redirection.

lib/sidekiq/job_retry.rb
```diff
@@ -134,6 +137,15 @@ def local(jobinst, jobstr, queue)
 
     private
 
+    def process_requeue(jobinst, jobstr)
+      retry_backoff = jobinst.class.get_sidekiq_options.dig("iteration", "retry_backoff") ||
```
mperham, line 141: Remember that sidekiq_options get persisted into the job payload when the job is created; you don't need to dig the value out of `get_sidekiq_options`, you'd use `job.dig("iteration", "retry_backoff")`.

mperham: Oh, I see; you didn't add these to the Job default options. That makes sense if overriding these elements is rare or you want to keep the job payload size as small as possible. If the elements are overridden, the job payload will contain the values so you can do:

```ruby
job.dig(...) || lifecycle.config.dig(...) # get from specific job or capsule's config
```

fatkodima:
Addressed feedback.

lib/sidekiq/processor.rb
```diff
@@ -79,7 +83,10 @@ def run
 
     def process_one(&block)
       @job = fetch
-      process(@job) if @job
+      if @job
+        @job.lifecycle = self if @job.is_a?(Job::Iterable)
```
mperham, line 87: Here, job will always be a `UnitOfWork` so this will never execute. You want to set this right before middleware is called, right after `JobClass.new`.

fatkodima: Right 🤦 The unit tests are too unit (with mocking), weren't able to detect this.

mperham:
Great work. I'll merge this and spend a little more time polishing it soon.

mperham:
We'll need a new Iteration wiki page which explains this major new feature. If you have any docs to use as a starting point, please free feel to create it.

mperham:
No problem, I will take care of it.

mperham:
Initial page here: https://github.com/sidekiq/sidekiq/wiki/Iteration

mperham:
I'm refactoring the API a bit and wondering if you considered using `to_enum` or `enum_for` instead of `build_enumerator`. I don't understand Enumerators very well so I'm trying to understand what would be the most idiomatic API.

https://www.rubydoc.info/stdlib/core/Enumerator

fatkodima:
`build_enumerator` is a method that user should define inside their job, which specifies what we should iterate over (and which should return an instance of an `Enumerator`). 
The user can use one of the convenient helpers we provide or specify their own (using `to_enum`/`enum_for`, for example). 
Some examples in https://github.com/fatkodima/sidekiq-iteration/blob/master/guides/custom-enumerator.md.

Let me know if I did not fully get the question.

mperham:
Why not Job#to_enum, instead of Job#build_enumerator? Do the methods have different meanings?

fatkodima:
If you mean why we use `build_enumerator` method name and not a `to_enum`, I don't know, with my english level, `Job#to_enum` sounds to me like we are converting this job to enum, which makes little sense. But `Job#build_enumerator` sounds like a method that builds an enumerator for this job. `to_enum` is a method already defined on `Object`, so we probably should not define our own.

You may provide some code example about how you imagine it to work, if that helps.

freemanoid:
I'm surprised this kind of functionality was integrated into sidekiq. I thought you should always break down long jobs into smaller ones if they require sidekiq-iteration. If it was built for ActiveRecord, that’s even more surprising to me.

mperham:
@freemanoid This provides an API for breaking down a long-running job into discrete chunks without having to fill Redis with thousands of small jobs. Some jobs have to remain serial and cannot be broken down into set of **concurrent** jobs.

The ActiveRecord portion is totally optional and does not force Rails upon the user.

PR #6291, opened 2024-05-14
Update example config

excid3:
This updates the example config to use default values for most things (like concurrency) and provide examples of how to customize.

It also includes some of the Rails default job queues for a more complete example.


mperham:
Thanks for adding the default set of Rails queues. That's a nice bit of polish for users.

PR #6292, opened 2024-05-14
New registration API for web extensions

mperham:
In order to lock down JS and CSS for better security, we need to provide a first-class API for web extensions so they can opt into better asset security while maintaining compatibility with existing versions.

Please see the content under `examples/webui-ext` and specifically the new `register` API:

```ruby
# DEPRECATED --- Old way to register
# Sidekiq::Web.register(SidekiqExt::RedisInfo::Web)
# Sidekiq::Web.tabs["Redis"] = "redis_info"

# NEW API IN 7.3.0
# This new way allows you to serve your own static assets
# and provide localizations for any strings in your extension.

Sidekiq::Web.register(SidekiqExt::RedisInfo::Web,
  name: "redis_info",
  tab: ["Redis"],              # The name on your Tab(s)
  index: ["redis_info"],       # The path to the root page(s) of your extension within the Web UI, usually "/sidekiq/" + index
  root_dir: SidekiqExt::RedisInfo::Web::ROOT, # root directory for all web ui content
  asset_paths: ["css", "js"],  # Paths within {root}/assets/{name} to serve static assets
  cache_for: 86400            # Enable http caching for one day of static assets
) do |sidekiq_web|
  # you can add your own middleware or additional settings here
end
```

The old way still works but will be removed in 8.0. Note also the new tag helpers `script_tag` and `style_tag` which handle the "nonce-sense" (ha ha) for you.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -5,12 +5,13 @@
 HEAD (7.3.0)
 ----------
 
-- Allow `Sidekiq::Limiter.redis` to use Redis Cluster [#6288]
-- Adjust CSP to disallow inline styling and scripts within the Web UI [#6270]
-  This should make Sidekiq immune to future XSS attacks. 3rd party extensions
-  will need to adjust their web assets accordingly.
+- **SECURITY** The Web UI no longer allows extensions to use `<script>`.
+  Adjust CSP to disallow inline scripts within the Web UI. Please see
+  `examples/webui-ext` for how to register Web UI extensions and use
+  dynamic CSS and JS. This will make Sidekiq immune to XSS attacks. [#6270]
 - Add config option, `:skip_default_job_logging` to disable Sidekiq's default
   start/finish job logging. [#6200]
+- Allow `Sidekiq::Limiter.redis` to use Redis Cluster [#6288]
 
 7.2.4
 ----------
--- a/Changes.md
+++ b/Changes.md
@@ -2,6 +2,16 @@
 
 [Sidekiq Changes](https://github.com/sidekiq/sidekiq/blob/main/Changes.md) | [Sidekiq Pro Changes](https://github.com/sidekiq/sidekiq/blob/main/Pro-Changes.md) | [Sidekiq Enterprise Changes](https://github.com/sidekiq/sidekiq/blob/main/Ent-Changes.md)
 
+HEAD (7.3.0)
+----------
+
+- Allow `Sidekiq::Limiter.redis` to use Redis Cluster [#6288]
+- Adjust CSP to disallow inline styling and scripts within the Web UI [#6270]
+  This should make Sidekiq immune to future XSS attacks. 3rd party extensions
+  will need to adjust their web assets accordingly.
+- Add config option, `:skip_default_job_logging` to disable Sidekiq's default
+  start/finish job logging. [#6200]
+
 7.2.4
 ----------
 
--- a/examples/webui-ext/readme.md
+++ b/examples/webui-ext/readme.md
@@ -1,0 +1,53 @@
+# Web UI Extensions in Sidekiq 7.3 and 8.0
+
+This is an example of how to extend Sidekiq's Web UI to add your own tab.
+
+## Initialization
+
+The Ruby code for your extension goes in `lib/` like any normal Rubygem.
+When the user includes the `my-gem` gem within their bundle, Bundler will call `require "my-gem"` which corresponds to `lib/my-gem.rb` in your project.
+This file should **not** require any web code or assets because you don't know if you're running within a Sidekiq process or a Web process.
+
+> Don't require your web extensions in Sidekiq's initializer. This is a common mistake.
+
+Your web extension should be activated by the user calling `require "my-gem/web"` in Rails' `config/routes.rb` as that guarantees we are in a Web process.
+
+## Registration
+
+Your Web extension must register itself with Sidekiq::Web to appear in the Sidekiq UI.
+See `lib/sidekiq-redis_info/web.rb` for a full-featured example.
+
+## Content Security
+
+Sidekiq 7.3 and 8.0 explicitly lock down the Web UI to prevent XSS and malicious content injection.
+In Sidekiq 7.3, the Web UI no longer allows ěmbedding JavaScript directly on the HTML page via
+`<script>...code...</script>`. In Sidekiq 8.0, inline styling via `<style>...css...</style>` will not work.
+
+### JavaScript
+
+JavaScript must be packaged in predefined .js files with a per-request nonce that proves it was rendered by the webapp:
+
+```html
+<script type="text/javascript" src="<%= root_path %>my-gem/js/somefile.js" nonce="<%= csp_nonce %>"></script>
+```
+
+### CSS
+
+Static CSS must be linked with the nonce:
+
+```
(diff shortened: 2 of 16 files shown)


markets:
Like the new API @mperham 👍 

Just a small suggestion. The `middlewares` object seems very flexible, but what about some high level flag to easily configure the assets path? Something like:

```ruby
Sidekiq::Web.register(SidekiqExt::RedisInfo::Web,
  tab: "Redis",
  index: "redis_info",
  asset_paths: ["/redis_info/css", "/redis_info/js"]
end
```

And rely on some sensible defaults for the rest.

Changes.md
```diff
@@ -2,6 +2,16 @@
 
 [Sidekiq Changes](https://github.com/sidekiq/sidekiq/blob/main/Changes.md) | [Sidekiq Pro Changes](https://github.com/sidekiq/sidekiq/blob/main/Pro-Changes.md) | [Sidekiq Enterprise Changes](https://github.com/sidekiq/sidekiq/blob/main/Ent-Changes.md)
 
+HEAD (7.3.0)
+----------
+
+- Allow `Sidekiq::Limiter.redis` to use Redis Cluster [#6288]
+- Adjust CSP to disallow inline styling and scripts within the Web UI [#6270]
```
Earlopain, line 9: Inline styling got reverted and inline scripts are already disallowed (#6074). This will be more accurate:

> Adjust CSP to disallow remote scripts within the Web UI.
> ... 3rd party extensions that load scripts from something like a CDN will need to adjust their web assets accordingly.

Also, perhaps something like:

```erb
# Before
<script type="text/javascript" src="https://code.jquery.com/jquery-3.7.1.min.js"></script>
# After
<script type="text/javascript" src="https://code.jquery.com/jquery-3.7.1.min.js" <%= defined?(csp_nonce) ? "nonce=\"#{csp_nonce}\"" : "" %>></script>
```

This will work in both 7.2 and 7.3 so extensions wouldn't have to worry about that.

examples/webui-ext/readme.md
```diff
@@ -0,0 +1,53 @@
...
+## Registration
+
+Your Web extension must register itself with Sidekiq::Web to appear in the Sidekiq UI.
+See `lib/sidekiq-redis_info/web.rb` for a full-featured example.
+
+## Content Security
+
+Sidekiq 7.3 and 8.0 explicitly lock down the Web UI to prevent XSS and malicious content injection.
+In Sidekiq 7.3, the Web UI no longer allows ěmbedding JavaScript directly on the HTML page via
+`<script>...code...</script>`. In Sidekiq 8.0, inline styling via `<style>...css...</style>` will not work.
```
Earlopain, line 24: A more common example (I think) would be something like `<div style="background=red">Hello World!</div>`. Listing both seems fine. You may also call out dynamic js like `myDiv.style.background = "red"`

examples/webui-ext/readme.md
```diff
@@ -0,0 +1,53 @@
...
+```html
+<script type="text/javascript" src="<%= root_path %>my-gem/js/somefile.js" nonce="<%= csp_nonce %>"></script>
+```
+
+### CSS
+
+Static CSS must be linked with the nonce:
+
+```
+<link href="<%= root_path %>redis_info/css/somefile.css" media="screen" rel="stylesheet" type="text/css" nonce="<%= csp_nonce %>" />
```
Earlopain, line 39: Perhaps now is the time to put a bit more thought into this: Styles/Scripts from the same origin don't need to supply a nonce. I added this somewhat defensivly per that one PR comment I read.

If the webui is served from `https://example.com/sidekiq` and styles/scripts are hosted under `https://example.com/sidekiq/assets` then all is fine. Is there actually a supported usecase where assets may be served from a different domain than where the webui is running from, like `https://s3.amazonaws.com/my_corp/assets/`? If that is not possible then adding nonces to these is not needed.

mperham: We could provide a helper method to include a script or style tag:

```erb
<%= stylesheet("redis_info/css/somefile.css", attribute: "value") %> <!-- local, nonce not required -->
<%= stylesheet("https://cdn.example.com/redis_info/css/somefile.css") %><!-- global, will add nonce -->

<%= javascript("redis_info/js/somefile.js", attribute: "value") %> <!-- local, nonce not required -->
<%= javascript("https://cdn.example.com/redis_info/js/somefile.js") %><!-- global, will add nonce -->
```

Earlopain: Yeah, that could be nice. Rails helpers have a `nonce` keyword option which transparently handles this like you proposed.

mperham:
> Just a small suggestion. The `middlewares` object seems very flexible, but a bit complex maybe 🤔? What about some more high level flag to easily configure the assets paths, something like:
> 
> ```ruby
> Sidekiq::Web.register(SidekiqExt::RedisInfo::Web,
>   tab: "Redis",
>   index: "redis_info",
>   asset_paths: ["/redis_info/css", "/redis_info/js"]
> ```

The issue is that I have to make a lot of assumptions to remove most of that boilerplate code. I want to include some caching of assets but can I assume that? Is a one day expiry ok? How do I know where the assets are located on disk (the root parameter)? I think I can do something like this:

```
Sidekiq::Web.register(SidekiqExt::RedisInfo::Web,
  tab: "Redis",
  index: "redis_info",
  asset_urls: ["/redis_info/css", "/redis_info/js"],
  root: SidekiqExt::RedisInfo::Web::ROOT,
  cache_for: 86400)
```

If we assume that `ROOT/assets`, `ROOT/locales`, `ROOT/views` is a directory convention we can rely on.

examples/webui-ext/readme.md
mperham: Implemented.

Earlopain: This looks fine, I think. The interpolation for key/value look a bit scary? Though I suppose it's always called with static parameters from the extension so I doubt it would matter much. Offhand I don't know if CGI even offers a method to escape attribute key/values.

You've left a debug puts in the html_tag method

markets:
Much simpler in my opinion 👌 

In general I feel that: Conventions > Configuration. 

I understand your concerns about some defaults, like caching, but since this panel is not critical or public facing, probably any sensible default is fine (and anyway, we can keep the flags so anyone is able to override it if needed).

mperham:
Here's the latest API from the (great!) feedback so far:

```ruby
Sidekiq::Web.register(SidekiqExt::RedisInfo::Web,
  name: "redis_info",
  tab: ["Redis"],              # The name on your Tab(s)
  index: ["redis_info"],       # The path to the root page(s) of your extension within the Web UI, usually "/sidekiq/" + index
  asset_paths: ["css", "js"],  # Paths within {root}/assets/{name} to serve static assets
  cache_for: 86400            # Enable http caching for one day of static assets
) do |sidekiq_web|
  # you can add your own middleware or additional settings here
end
```

Tabs/Index can now be Arrays because I noticed sidekiq-unique-jobs adds multiple Tabs.

markets:
Nice job @mperham 👍 Really like this version!

mperham:
Ok, I'm happy with this PR as is. Any further feedback? cc @mhenrixon

mhenrixon:
This looks great to me! Solves more than a few of the issues I've had reported 👍

PR #6307, opened 2024-05-23
Retain CurrentAttributes after inline execution

### LINKED ISSUE #6306: Current attributes lost after inline execution
Ruby version: 3.3.0
Rails version: 7.1.1
Sidekiq / Pro / Enterprise version(s): 7.2.4

While upgrading from Sidekiq 6 to 7, we ran into this problem through our tests. We have Sidekiq jobs which themselves call `perform_async` and found that the parent job lost their current attributes after performing the child job.

It's easiest to demonstrate this with a test against `perform_inline` directly though:

```ruby
# test/current_attributes_test.rb

class CurrentAttributesJob
  include Sidekiq::Job

  def perform; end
end

describe "Current attributes" do
  before do
    @config = reset!
  end

  it "persists after perform_inline" do
    Sidekiq::CurrentAttributes.persist(Myapp::Current, @config)
    Myapp::Current.user_id = 1
    assert_equal 1, Myapp::Current.user_id
    CurrentAttributesJob.perform_inline
    assert_equal 1, Myapp::Current.user_id # got nil
  end
end
```

The problem is `Sidekiq::CurrentAttributes::Load#call` resets all the attributes afterwards. 

We think that instead it should call `ActiveSupport::CurrentAttributes.set` on each class, i.e.

```ruby
Myapp::Current.set(user_id: 123) do
  Myapp::OtherCurrent.set(other_id: "abc") do
    yield
  end
end
```

One way of doing that:

```ruby
    def call(_, job, _, &block)
      klass_attrs = {}

      @cattrs.each do |(key, strklass)|
        next unless job.has_key?(key)

        klass_attrs[strklass.constantize] = job[key]
      end

      klass_attrs.inject(block) { |chain, (klass, attrs)|
        -> { klass.set(attrs, &chain) }
      }.call
    end
```

Or with recursion, easier to follow:

```ruby
    def call(_, job, _, &block)
      klass_attrs = {}

      @cattrs.each do |(key, strklass)|
        next unless job.has_key?(key)

        klass_attrs[strklass.constantize] = job[key]
      end

      wrap(klass_attrs.to_a, &block)
    end

    def wrap(klass_attrs, &block)
      klass, attrs = klass_attrs.shift
      return block.call unless klass

      klass.set(attrs) do
        wrap(klass_attrs, &block)
      end
    end
```

steveh:
Ensure CurrentAttributes are restored to their original values, rather than cleared.

Fixes #6306

### DIFF
--- a/test/current_attributes_test.rb
+++ b/test/current_attributes_test.rb
@@ -14,6 +14,13 @@ module Myapp
   end
 end
 
+class CurrentAttributesJob
+  include Sidekiq::Job
+
+  def perform
+  end
+end
+
 describe "Current attributes" do
   before do
     @config = reset!
@@ -91,6 +98,15 @@ describe "Current attributes" do
           assert_equal 17, job_hash["cattr_1"][:other_id]
         end
       end
+    end
+  end
+
+  it "persists after perform_inline" do
+    Sidekiq::CurrentAttributes.persist("Myapp::Current", @config)
+    with_context("Myapp::Current", :user_id, 16) do
+      assert_equal 16, Myapp::Current.user_id
+      CurrentAttributesJob.perform_inline
+      assert_equal 16, Myapp::Current.user_id
     end
   end
 
--- a/lib/sidekiq/middleware/current_attributes.rb
+++ b/lib/sidekiq/middleware/current_attributes.rb
@@ -46,22 +46,26 @@ module Sidekiq
       end
 
       def call(_, job, _, &block)
-        cattrs_to_reset = []
+        klass_attrs = {}
 
         @cattrs.each do |(key, strklass)|
-          if job.has_key?(key)
-            constklass = strklass.constantize
-            cattrs_to_reset << constklass
+          next unless job.has_key?(key)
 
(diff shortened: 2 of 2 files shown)


test/current_attributes_test.rb
```diff
@@ -94,6 +101,15 @@ class OtherCurrent < ActiveSupport::CurrentAttributes
     end
   end
 
+  it "persists after perform_inline" do
+    Sidekiq::CurrentAttributes.persist("Myapp::Current", @config)
+    with_context("Myapp::Current", :user_id, 16) do
+      assert_equal 16, Myapp::Current.user_id
+      CurrentAttributesJob.perform_inline
+      assert_equal 16, Myapp::Current.user_id
```
mperham, line 109: Is this really the semantic we want? We mutated a global variable but didn’t reset it. It feels wrong.

mperham: This code is really complex in its side effects and details so I’m struggling to understand the best approach. The value was already set when we called perform_inline so resetting it would be wrong too. I think this is better.

PR #6308, opened 2024-05-23
Refactor iteration

mperham:
Refactoring to iteration. Thoughts:

* Is the terminology correct?
* Do we have enough good code examples so people can grok the feature?
* Is this the right way to opt into the feature?

```ruby
class MyJob
  include Sidekiq::Job
  include Sidekiq::Job::Iterable

  def build_enumerator(cursor)
  end

  def each_iteration(item)
  end
end
```

What about instead?

```ruby
class UpdateUserBankAccount
  include Sidekiq::Job::Iterable

  def build_enumerator(cursor)
  end

  def process(item)
  end
end
```


fatkodima:
Not strong opinions, but :
1. I am not in favor of `process` method name, because it sounds like the same as `perform` and is confusing, imo. Maybe there is a better method name than `each_iteration` 🤷 
2. I see the ability to automatically interrupt jobs after some time was removed. This is still a useful feature, if some job allocates a good amount of memory or just to free the worker after some time. Do you prefer this to be implemented by the user itself?

mperham:
1. I agree, it's a bit generic and easy to confuse with perform.
2. I see your usecase. I'm still trying to figure out the best way to implement this, it may come back later.

mperham:
One API issue I want to fix, I don't like how the user includes two Modules here:

```ruby
class MyJob
  include Sidekiq::Job
  include Sidekiq::Job::Iterable
```

The contract for Sidekiq::Job is that Jobs must supply a `perform` method. But S::J::Iterable explicitly raises an error if the user supplies a `perform` method. Should the user pick one Module or the other?

fatkodima:
Just using `include Sidekiq::Job::Iterable` feels like we are adding some functionality to the (now missing) `Sidekiq::Job`. Maybe rename it to just be `include Sidekiq::IterableJob`?

People may still try to erroneously define the `perform` method 🤔.

mperham:
Yeah, I like that rename. `Sidekiq::IterableJob` is a good idea.

mperham:
What do you think about using a new term? Instead of `Sidekiq::IterableJob`, which seems awkward in English, `Sidekiq::Chunk`? A Chunk is a Job which processes a sequence of data which can be stopped and restarted.

fatkodima:
I am definitely not the right person to ask about correct English terms, but personally, I don't like it. `Chunk` is too generic to me. Maybe someone can come up with other ideas.

fatkodima:
Maybe just `IteratingJob`, I don't know 🤷

mperham:
We could lean into idiomatic Ruby and call it `Sidekiq::EnumerableJob`.

mperham:
The current API is `build_enumerator` and `each_iteration`. WDYT about changing that to `build_iterator` to keep the terminology the same?

```ruby
class SomeJob
  include Sidekiq::IterableJob
  sidekiq_options iteration: { retry_after: 30.seconds }

  def build_iterator(...)
    ...
  end

  def each_iteration(item)
    ...
  end
end

**Comment by mperham** [comment]
Alternative naming:

```ruby
class SomeJob
  include Sidekiq::ChunkableJob
  sidekiq_options chunk: { retry_after: 30.seconds }

  def build_chunk(...)
    # ...
  end

  def each_item(item)
    # ...
  end
end
```

mperham:
Yeah, you don't need to comment or answer every question. Some of this is just me throwing out suggestions. Anyone can chime in with their opinion.

fatkodima:
Like the 1st variant (with `build_iterator`).

rainerborene:
What about `Sidekiq::SequenceJob` or `Sidekiq::ChainableJob`? I would prefer something like `Sidekiq::ChainableJob` or `Sidekiq::IterableJob`. 

One question: does it make sense to "split" a cursor into multiple long-running jobs to parallelize and speed up the process? It would be great to mention some examples and use cases in the wiki on how to get the most out of this feature.

mperham:
I noticed you don't pass the job arguments to the callback methods but you do pass them to `build/each`. Has there been any demand for this?

```ruby
build_enumerator(*args, cursor:)
on_start
each_iteration(item, *args)
```

fatkodima:
There was no demand for arguments in callbacks, Shopify's `job-iteration` goes the same route, so not sure yet about its usefulness.

lib/sidekiq/iterable_job.rb
```diff
@@ -0,0 +1,51 @@
...
+#
+#     enum = ProcessUserSet#build_enumerator(123, cursor: updated_cursor)
+#
+# Note there are several APIs to help you build enumerators for
+# ActiveRecord Relations, CSV files, etc. See sidekiq/job/iterable/*.rb.
+module Sidekiq
+  module IterableJob
+    def self.included(base)
+      base.send(:include, Sidekiq::Job)
+      base.send(:include, Sidekiq::Job::Iterable)
```
fatkodima, line 33: Interesting, why rubocop does not suggest that the method is public for a long time
```suggestion
      base.include(Sidekiq::Job)
      base.include(Sidekiq::Job::Iterable)
```

mperham:
While I think we could bike shed the API a bit, it's also reasonable to point out that sticking with the same API as and retaining compatibility with the job-iteration and sidekiq-iteration gems is really valuable in itself. For that reason, I'm not going to change any APIs and just accept it as is.

Changes.md
```diff
@@ -5,14 +5,30 @@
...
+class ProcessArrayJob
+  include Sidekiq::IterableJob
+  def build_enumerator(*args, **kwargs)
+    array_enumerator(args, **kwargs)
+  end
+  def each_iteration(arg)
+    puts arg
+  end
+end
+ProcessArrayJob.perform_async(1, 2, 3)
```
fatkodima, line 22: This example looks not very much useful and confusing, imo. What about using something like here https://github.com/fatkodima/sidekiq-iteration?tab=readme-ov-file#getting-started ?

mperham: Yep, I'll remove it there and add 1-2 real world examples to the wiki page tomorrow.

mperham:
I found what seems to be a major issue in integration testing. I have this job:

```ruby
class PostCreator
  include Sidekiq::IterableJob

  def build_enumerator(start_at, count, **kwargs)
    p kwargs
    array_enumerator((start_at...(start_at + count)).to_a, **kwargs)
  end

  def each_iteration(pid, *args)
    puts pid
    Post.transaction do
      Post.create!(id: pid, title: "Post #{pid}", body: "Body of post #{pid}")
    end
  end
end
```

Its purpose is to create N Post records in the database. Call it like this:

> 100.times {|idx| PostCreator.perform_async(idx*1000, 1000) }

This will create 100 jobs, each of which creates 1,000 Posts in sqlite. If I let them run normally, it works fine.

Now if I interrupt one of those jobs, it persists the cursor **BUT** that cursor is for the record which was just created. So if I am creating Post # 545, it will finish the `each_iteration` call and save 545 as the cursor. When it restarts the job, it tries to re-create Post # 545, raising a uniqueness violation.

On interruption, should we advance the Enumerator and save **that next cursor** instead?

fatkodima:
> On interruption, should we advance the Enumerator and save that next cursor instead?

It was implemented to restart from the same cursor only to support the nested enumerator we have. Because otherwise, when restarting, we can potentially unintentionally skip iterations. But it is advised for people to have their `each_iteration` methods idempotent (similar to `perform` method).

fatkodima:
In `job-iteration` they basically have a non-working nested iterator which skips iterations, but probably they don't use it and it is not noticed yet.

mperham:
What's the usecase for the nested enumerator? It looks complicated enough that we might want to remove it. Complex code is hard to reason about, debug and use correctly. And most importantly for us, painful to support.

mperham:
I see: https://github.com/fatkodima/sidekiq-iteration?tab=readme-ov-file#nested-iteration

mperham:
I'd rather see people create a tree of jobs, rather than a single job which crawls a tree with this nested iterator.

> Shop.all.each {|shop| CrawlProductVariantsJob.perform_async(shop.id) }

fatkodima:
This can potentially create a lot of jobs, depending on the records sizes in each loop, and harder to manage these jobs 🤷.

It would be great to have some insights from Shopify about this (but as I mentioned, they probably don't use it yet because theirs implementation looks broken).

mperham:
I just fixed the cursor issue. I know we want idempotency but we can't require it due to a bug in our implementation.

We can add more complex nested enumerator support in the future if there is demand. Let's start with the easy and simple API first.

PR #6310, opened 2024-05-26
Fix polling interval formatting

getsnoopy:


getsnoopy:
@mperham Why was this closed without being merged?

mperham:
If you can’t spend the time to describe what the PRbis for, I can’t spend the time reviewing it.

getsnoopy:
The commit/PR subject is comprehensively self-explanatory, and the PR diff itself is a 2-line change, so I doubt it's a lack of effort on my part, but anyway:

The formatting for the polling interval is incorrect (it uses the incorrect symbol for *seconds*), so this commit fixes it.

mperham:
Changing "sec" to "s" is not "comprehensively self-explanatory" at all. It's been that way for years and you're the first person to have an issue with it.

I'm not psychic, a PR is the start of a conversation and "No description provided" says "it's not worth my time to help the PR reviewer understand my POV and why I want this change".

getsnoopy:
Maybe I overestimated the "obviousness" of the mistake. Here's the explanation:

"sec" is an incorrect abbreviation for the unit *second*; it's an SI (modern metric) unit, so it has a symbol, not an abbreviation, and it is "s".

That's what prompted the change. I figured this was common knowledge (as it's taught in most grade school science classes), which is why I didn't provide a description.

PR #6313, opened 2024-05-29
Do not fail on invalid locale in session hash

mamhoff:
Prior to this commit, Sidekiq would treat the `locale` key in the session as safe: If it's set, it's surely a supported locale. However, in complex applications, the `locale` key in the session might be set by another app to some value that is not what Sidekiq expects.

In this case, Sidekiq fails with a 404 error and no good explanation.

What this commit does is make sure that the value in `session[:locale]` is actually a supported, available locale for Sidekiq. In case this is an unwanted locale, a user can still use the locale selector to change it back, but the Sidekiq admin doesn't suddenly become unreachable.

Another route this issue could be solved is by namespacing the `locale` key in the session to something like `sidekiq_locale`, in order to be safer from other parts of the app setting this value to something odd. I'm opting for this approach though because I believe it to be a less invasive change.

### DIFF
--- a/lib/sidekiq/web/helpers.rb
+++ b/lib/sidekiq/web/helpers.rb
@@ -166,7 +166,7 @@ module Sidekiq
     def locale
       # session[:locale] is set via the locale selector from the footer
       # defined?(session) && session are used to avoid exceptions when running tests
-      return session[:locale].to_s if defined?(session) && session&.[](:locale)
+      return session[:locale].to_s if defined?(session) && session&.[](:locale) && available_locales.include?(session[:locale].to_s)
 
       @locale ||= begin
         matched_locale = user_preferred_languages.map { |preferred|
--- a/test/web_helpers_test.rb
+++ b/test/web_helpers_test.rb
@@ -108,6 +108,18 @@ describe "Web helpers" do
     assert_equal "es", obj.locale
   end
 
+  it "tests user selected locale with invalid locale" do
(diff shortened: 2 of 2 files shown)


tvdeyen (approved):
Works for me! Thanks :)

mperham:
I don't think this is the right fix. Sidekiq should not render a 404 if the locale is corrupt. It should fall back to English.

mamhoff:
That's what this PR does, precisely. Stop the 404 from happening.

Mike Perham ***@***.***> schrieb am Do., 30. Mai 2024, 17:48:

mperham:
Well, I don't think it's the right approach. I don't want that logic in the locale getter. Where is the 404 generated?

mperham:
I refactored a bit and pushed your logic inside the `begin` block so it's not executed on every call to `locale`. Thanks for opening the issue.

tvdeyen:
We updated to Sidekiq 7.2.4 and the bug is still present. In which version has the bugfix been released?

tvdeyen:
A misunderstanding on my side, apologies. 

Any reason why this bug  fix is not released as a patch level release? 

Our web UI is unusable at the moment. And I do not know how to make it work again. Is there a known way to fix this on the app level?

mperham:
Run main:

gem "sidekiq", github: "sidekiq/sidekiq"

I'm hoping to release 7.3.0 in the next week.

PR #6320, opened 2024-06-06
Add `logger` as an explicit dependency

### LINKED ISSUE #6319: Dependency on the `logger` gem which will be removed in Ruby 3.5
Same deal as with all the others that happened in the last few years. https://github.com/ruby/ruby/commit/d7e558e3c48c213d0e8bedca4fb547db55613f7c

https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/logger.rb

Will probably just need to be added to the gemspec

Earlopain:
Sidekiq uses this and it's getting the same treatment like `base64`, `mutex_m`, etc. Starting in Ruby 3.4 a warning will be emitted, and Ruby 3.5 will error

Closes #6319

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -16,6 +16,7 @@ HEAD (7.3.0)
 - Retain CurrentAttributeѕ after inline execution [#6307]
 - Raise default Redis {read,write,connect} timeouts from 1 to 3 seconds
   to minimize ReadTimeoutErrors [#6162]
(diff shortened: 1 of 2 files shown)


amomchilov:
For searchabilty, would you mind adding the exact warning message into your PR description?

```
logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add logger to your Gemfile or gemspec to silence this warning.
```

Earlopain:
Sure, done

PR #6326, opened 2024-06-11
Don't allow to pass nil to Rack::Utils.escape_html

knapo:
Rack 3.1 [changed](https://github.com/rack/rack/pull/2099) how Rack::Utils.escape_html works and now it fails on nil. 

Fixes https://github.com/sidekiq/sidekiq/issues/6325


mperham:
I'd prefer 'text.to_s'

knapo:
Updated.

mperham:
Fixed by 94b998bf

PR #6333, opened 2024-06-17
Eagerly load API to allow usage when route drawing is deferred

### LINKED ISSUE #6332: Sidekiq API uninitialized when route drawing is deferred
Ruby version: ruby 3.3.2
Rails version: main
Sidekiq / Pro / Enterprise version(s): main

Since [rails#52012](https://github.com/rails/rails/pull/52012) the route drawing is being deferred on environments with eager load set to false. The sidekiq API was loaded at this point, so when sidekiq routes weren't mounted we're unable to use the sidekiq API on rails console.

Here is how to reproduce the error:
```shell
rails new minimal --edge --minimal
cd minimal
bundle add sidekiq
sed -i 's/^gem "sidekiq".*/gem "sidekiq", github: "sidekiq/sidekiq", ref: "main"/' Gemfile
bundle
rails c
```
```
Loading development environment (Rails 8.0.0.alpha)
minimal(dev)> Sidekiq::Queue
(minimal):1
`<top (required)>': uninitialized constant Sidekiq::Queue (NameError)

Sidekiq::Queue
```

victorsmoreira:
Now that route drawing will be deferred on environments where eager load is set to false on rails 8, the Sidekiq API wasn't available on rails console in these environments.

Closes #6332

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -27,6 +27,7 @@ begin
 rescue LoadError
 end
 
+require "sidekiq/api"
 require "sidekiq/config"
 require "sidekiq/logger"
 require "sidekiq/client"
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "sidekiq"
-
 require "zlib"
 require "set"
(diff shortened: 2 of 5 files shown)


mperham:
Not the right approach.

Issue #6335, opened 2024-06-19
Unmet dependency: Redis 6.2.0 doesn't exist


badde-media:
Ruby version: ruby 3.3.2 (2024-05-30 revision e5a195edf6) [x86_64-linux] (asdf)
Rails version: Rails 7.1.3.4
Sidekiq / Pro / Enterprise version(s): sidekiq (7.2.4)

Please include your initializer, sidekiq.yml, and any error message with the full backtrace.
I'm trying to get Rails BulletTrain running. That seems to have a sidekiq dep to redis 6.2.0, which doesn't exist. Possibly a typo?
```
worker           | You are connecting to Redis 6.0.16, Sidekiq requires Redis 6.2.0 or greater
worker           | /home/un/.asdf/installs/ruby/3.3.2/lib/ruby/gems/3.3.0/gems/sidekiq-7.2.4/lib/sidekiq/cli.rb:77:in `run'
worker           | /home/un/.asdf/installs/ruby/3.3.2/lib/ruby/gems/3.3.0/gems/sidekiq-7.2.4/bin/sidekiq:31:in `<top (required)>'
worker           | /home/un/.asdf/installs/ruby/3.3.2/bin/sidekiq:25:in `load'
worker           | /home/un/.asdf/installs/ruby/3.3.2/bin/sidekiq:25:in `<top (required)>'
worker           | /home/un/.asdf/installs/ruby/3.3.2/lib/ruby/gems/3.3.0/gems/bundler-2.4.10/lib/bundler/cli/exec.rb:58:in `load'
worker           | /home/un/.asdf/installs/ruby/3.3.2/lib/ruby/gems/3.3.0/gems/bundler-2.4.10/lib/bundler/cli/exec.rb:58:in `kernel_load'
worker           | /home/un/.asdf/installs/ruby/3.3.2/lib/ruby/gems/3.3.0/gems/bundler-2.4.10/lib/bundler/cli/exec.rb:23:in `run'
worker           | /home/un/.asdf/installs/ruby/3.3.2/lib/ruby/gems/3.3.0/gems/bundler-2.4.10/lib/bundler/cli.rb:492:in `exec'
worker           | /home/un/.asdf/installs/ruby/3.3.2/lib/ruby/gems/3.3.0/gems/bundler-2.4.10/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
[...]

PR #6336, opened 2024-06-19
chore: Update Redis version requirement to 5.2.0 or greater

badde-media:
Bugfix for issue #6335

### DIFF
--- a/lib/sidekiq/cli.rb
+++ b/lib/sidekiq/cli.rb
@@ -74,7 +74,7 @@ module Sidekiq # :nodoc:
       # fire startup and start multithreading.
       info = @config.redis_info
       ver = Gem::Version.new(info["redis_version"])
-      raise "You are connecting to Redis #{ver}, Sidekiq requires Redis 6.2.0 or greater" if ver < Gem::Version.new("6.2.0")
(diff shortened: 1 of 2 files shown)


badde-media (commented):
https://github.com/sidekiq/sidekiq/issues/6335

mperham:
Redis server, not the gem.

badde-media:
Well it works after my fix.

badde-media:
At least the app boots up. But log shhows a whole bunch of errors like this one:
```
worker           | 2024-06-19T19:34:42.656Z pid=231024 tid=4ezo ERROR: ERR Error running script (call to f_000a4c2f82df69db1780053377a2c9b0741570e0): @user_script:2: ERR value is not an integer or out of range (redis://localhost:6379)
```
Please correct me if I'm wrong, but doesn't
```
if ver < Gem::Version.new("6.2.0")
```
refer to the version of the Redis gem, which should be at least 6.2.0, but the most recent version of this gem being 5.2.0?

mperham:
> refer to the version of the Redis gem, which should be at least 6.2.0, but the most recent version of this gem being 5.2.0?

Not at all. Sidekiq is telling you you are connected to Redis 6.0.16, so how could Redis version 6 not exist? We refer to the redis gem as redis-rb. Redis is the actual server and the latest version is 7.2.5.

https://github.com/redis/redis/releases/tag/7.2.5

PR #6356, opened 2024-07-04
Add X-Content-Type-Options header

dee-see:
Hello! This is a very minor change but a security scanner flagged this header as missing and even if there's no real security risk I figured it was easy enough to add to get a clean scan result on our application. ;)


mperham:
I'll need more documentation about what this header is and what is the appropriate value for Sidekiq to use.

mperham:
Did some reading. Looks good.

PR #6367, opened 2024-07-16
Add note about required dogstatsd-ruby gem version

nicoleepp:
We ran into an error when trying to upgrade Sidekiq gems to v7 without first upgrading `dogstatsd-ruby` from v4 to v5.

```rb
NoMethodError
undefined method `host' for #<Datadog::Statsd:0x00007f29c1c5bc40 ...
...
sidekiq-pro (7.2.1) lib/sidekiq/pro/metrics.rb in block in location at line 81
```

Adding a note for others that may run into the same issue.


mperham:
Thank you!

PR #6385, opened 2024-08-01
Require component from job logger

nunosilva800:
Since this [change](https://github.com/sidekiq/sidekiq/commit/6677b453982cab276892e7dad65cb320582b7de3), (and [sidekiq v7.3.0](https://github.com/sidekiq/sidekiq/releases/tag/v7.3.0)) the JobLogger depends on `Sidekiq::Component`. 

For apps that only `require "sidekiq/job_logger"`, this can lead to a `NameError`:

```
/home/nuno/.bundle/myapp/ruby-3.3.1/gems/sidekiq-7.3.0/lib/sidekiq/job_logger.rb:5:in `<class:JobLogger>': uninitialized constant Sidekiq::Component (NameError)

    include Sidekiq::Component
                   ^^^^^^^^^^^
        from /home/nuno/.bundle/myapp/ruby-3.3.1/gems/sidekiq-7.3.0/lib/sidekiq/job_logger.rb:4:in `<module:Sidekiq>'
        from /home/nuno/.bundle/myapp/ruby-3.3.1/gems/sidekiq-7.3.0/lib/sidekiq/job_logger.rb:3:in `<top (required)>'
        from /opt/rubies/ruby-3.3.1/lib/ruby/3.3.0/bundled_gems.rb:74:in `require'
```

This PR adds `require "sidekiq/component"` to avoid this error from happening.

### DIFF
--- a/lib/sidekiq/job_logger.rb
+++ b/lib/sidekiq/job_logger.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "sidekiq/component"
+
 module Sidekiq
   class JobLogger
     include Sidekiq::Component


mperham:
This has been submitted at least twice already.  It is by design. We need a code comment to explain.

maael:
[For clarity, here's a link to the explanation on the original PR](https://github.com/sidekiq/sidekiq/pull/6354#issuecomment-2206766817).

> JobLogger is only available in server mode. This error is on purpose as you can't use JobLogger outside of the Sidekiq process.

rafaelfranca:
I think this isn't taking in consideration even the design of Sidekiq. You are using the logger outside the server in the `test/job_logger_test.rb`, and tests on that file fail when are run alone.

```
$ ruby test/job_logger_test.rb
/workspaces/sidekiq/lib/sidekiq/job_logger.rb:5:in `<class:JobLogger>': uninitialized constant Sidekiq::Component (NameError)

    include Sidekiq::Component
                   ^^^^^^^^^^^
Did you mean?  Complex
        from /workspaces/sidekiq/lib/sidekiq/job_logger.rb:4:in `<module:Sidekiq>'
        from /workspaces/sidekiq/lib/sidekiq/job_logger.rb:3:in `<top (required)>'
        from /usr/local/lib/ruby/3.3.0/bundled_gems.rb:74:in `require'
        from /usr/local/lib/ruby/3.3.0/bundled_gems.rb:74:in `block (2 levels) in replace_require'
        from test/job_logger_test.rb:4:in `<main>'
```

Sure, you could require `Sidekiq::Component` in that file as well and fix this problem, but it is pointing to a bigger issue that you can't define and test a custom Sidekiq loggers using the base implementation without knowing to require an internal implementation of that file that can change any time, like we just saw.

Can you please revisit this design decision please?

mperham:
I want to better understand why people are trying to customize logging. One of the benefits of Rails is convention over configuration and this is an example where people are not accepting Sidekiq's conventions. I want to make it hard to change the defaults but at the same time I want the most people to be happy with those defaults.

So why are people changing the logger? Why is this necessary?

rafaelfranca:
In our case we are using a eventted based logger instead of a String based logger. So instead of adding strings to logger we add events in the format of hashes. It is very similar to the `ActiveSupport::Instrumentation` payload.

I can share the code since there isn't anything sensitive on it. We have a logger that looks like this:

```ruby
# frozen_string_literal: true

require "sidekiq/job_logger"
require "sidekiq/job_retry"

  module CanonicalLogging
    module Sidekiq
      class JobLogger < ::Sidekiq::JobLogger
        def call(job, _queue)
          # We need to wrap the entire Sidekiq job processing in a unit of work because we want to just send one
          # event.
          OurLogger.with_unit_of_work do
            job_started_at = Time.now.utc.iso8601
            yield
          rescue ::Sidekiq::JobRetry::Handled => h
            # Sidekiq handles errors (e.g. by enqueueing a retry) and reraises as Sidekiq::JobRetry::Handled
            # (or subclass thereof). We want the original cause, which is most often available.
            error = h.cause || h
            OurLogger.log_append(
              exception_class: error.class.name,
              exception_message: error.message,
            )
            raise
          rescue => error
            OurLogger.log_append(
              exception_class: error.class.name,
              exception_message: error.message,
            )
            raise
          ensure
            job_ended_at = Time.now.utc.iso8601
            OurLogger.safe_log_append(hash_from_sidekiq(job, job_started_at, job_ended_at))
          end
        end

        private

        def hash_from_sidekiq(job, started_at, ended_at)
          job_class = job["wrapped"] || job["class"]

          state = {
            type: "jobs",
            job_class: job_class,
            job_created_at: epoc_to_iso8601(job["created_at"]),
            job_ended_at: ended_at,
            job_id: job["jid"],
            job_queue_name: job["queue"],
            job_retry_policy: job["retry"],
            job_started_at: started_at,
          }

          klass = job_class.safe_constantize
          if klass.respond_to?(:log_arguments?)
            state[:job_arguments] = job["args"] if klass.log_arguments?
          else
            state[:job_arguments] = job["args"]
          end

          # enqueued_at will not be present for scheduled jobs using Worker.perform_at(...)
          state[:job_enqueued_at] = epoc_to_iso8601(job["enqueued_at"]) if job["enqueued_at"]

          # retry_count is present on first retry (attempt 2), initial value is 0
          if job["retry_count"]
            state[:job_first_failed_at] = epoc_to_iso8601(job["failed_at"])
            state[:job_retry_attempts] = job["retry_count"] + 1
          end

          # retried_at is present on second or subsequent retry (attempt 3+)
          if job["retried_at"]
            state[:job_last_retried_at] = epoc_to_iso8601(job["retried_at"])
          end

          state
        end

        def epoc_to_iso8601(epoc)
          return if epoc.nil?

          Time.at(epoc).utc.iso8601
        rescue TypeError
          epoc.to_s
        end
      end
    end
  end
```

I believe if Sidekiq instead of plainly writing to the logger was instead emitting a notification we could, instead of overriding the logger, subscribe to the event and use it to send to our system. The tradeoff is that this would be indirection for Sidekiq that perhaps would not be necessary for most of people.

PR #6389, opened 2024-08-04
Fix the comparison of version from strings to numerics.

valk:
This is a lexicographical comparison: '7.10' < '7.2' which gives `true`, which is incorrect. This is solved using `Gem::Version.new`.


mperham:
I believe it compares "7.1.0" to "7.2" which compares correctly. The code is this way because I don't envision Rails ever having ten minor releases.

PR #6405, opened 2024-08-23
Iterable Jobs: Update cursor after processing iteration

ohthatjames:
Currently the cursor is updated on each iteration before running the
iteration. This means that if the job is interrupted, the cursor will
point to the id of the record that was about to be processed.

When the job is resumed, if the enumerator returns a single item then
everything works fine because the cursor points to the only object we
need. However, for `ActiveRecordEnumerator#batches` the cursor points at
the last record of the batch that hadn't been processed, so when the
enumerator runs we get a new batch starting from that record. That means
we effectively don't process all but the last record from the
interrupted batch.

The [first commit](https://github.com/sidekiq/sidekiq/commit/f7ae8eeef5acf0a12f491dfd58a7a1c6db0b8747) reproduces this behaviour:

```
1) Failure:
Sidekiq::Job::Iterable#test_0019_reschedules batches when sidekiq is
stopping [test/iterable/iterable_test.rb:237]:
--- expected
+++ actual
@@ -1 +1 @@
-[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+[1, 2, 3, 6, 7, 8, 9, 10]

```

The second commit moves the cursor update to the end of the iteration so that
we don't lose most of the interrupted batch. The side-effect of this is
that jobs that enumerate one record at a time will now always rerun the
record associated with the last successful cursor. This should be fine
as the documentation[1] states

```
As always, leverage DB transactions to ensure idempotency and consistent
data.
```

I've had to update some of the existing tests to account for the fact
that resumption of single-record enumerators will now rerun the last
good record. In `calls iteration hooks` we now call the the
around_iteration hook twice more for the two resumptions, and in `can be
resumed` the cursor doesn't increment as often.

I might be missing some nuance as to why this solution doesn't work, but if nothing else hopefully the failing test is useful.

[1] https://github.com/sidekiq/sidekiq/wiki/Iteration


mperham:
> The side-effect of this is that jobs that enumerate one record at a time will now always rerun the record associated with the last successful cursor.

This seems like an unacceptable side effect, it was the reason why the cursor was saved where it was in the first place. Can we handle both cases gracefully?

mperham:
e.g. can the batch cursor save the start_index so it can pull the precise batch window again and skip the elements already processed?

ohthatjames:
> e.g. can the batch cursor save the start_index so it can pull the precise batch window again and skip the elements already processed?

Yup, that sounds like a better solution - I'll amend the PR!

ohthatjames:
I've updated the code to set the cursor to the first element in the batch, and added another test for relations to show the same behaviour.

mperham:
looks like this would restart the entire batch but that seems like a reasonable semantic.

@fatkodima any opinions?

fatkodima:
Lgtm 👍

Issue #6240, opened 2024-03-22
Enterprise 7.2 and "NX: true"


nthx:
Ruby version: 3.0.6
Rails version: 7.0.8.1
Sidekiq / Pro / Enterprise version(s): affected 7.2.2
Sidekiq / Pro / Enterprise version(s): not affected 7.1.2
Redis server version: redis_version: 6.2.13; but 7.+ is also affected

## Description
When starting sidekiq `sidekiq -c 2` it raises an error, but continues to run. The error:
> TypeError: Unsupported command argument type: TrueClass

NOTE: I have modified the error in the following log, to get the full context of what's actually being sent to Redis.

Patch that fixes the issue:
> /site/public/.rvm/gems/ruby-3.0.6/gems/sidekiq-ent-7.1.2/lib/sidekiq-ent/senate.rb:85
      result = redis { |conn|
        #conn.set(@key, identity, ex: TTL, nx: true) == "OK"
        conn.call("SET",@key,identity,"EX",TTL,"NX") == "OK"
      }

I have also checked Redis documentation and "NX" command takes no arguments
https://redis.io/commands/set/

```
@pc:~$ redis-cli -h pc -p 6379  -n 3 
pc:6379[3]> set 'foo' 'value' EX 1 NX 'true'
(error) ERR syntax error
pc:6379[3]> set 'foo' 'value' EX 1 NX
OK
```

The solution for me is NOT to upgrade to the latest 7.2, but stay at 7.1.* of `sidekiq` gem

Let me know if you need more information.

Full startup log:
```
INFO: Booted Rails 7.0.8.1 application in development environment 
INFO: Running in ruby 3.0.6p216 (2023-03-30 revision 23a532679b) [x86_64-linux]
INFO: Sidekiq Pro 7.1.5 / Sidekiq Enterprise 7.1.2, commercially licensed.
INFO: Sidekiq 7.2.2 connecting to Redis with options {:size=>10, :pool_name=>"internal"
INFO: Sidekiq 7.2.2 connecting to Redis with options {:size=>3, :pool_name=>"default"
INFO: Starting processing, hit Ctrl-C to stop
WARN: {"context":"senate"}
[...]

PR #6419, opened 2024-09-06
Replace usage of OpenStruct with Struct

davidrunger:
Ruby 3.3.5 warns when requiring `ostruct` from the standard library (as opposed to requiring it as a gem). Such a warning can be seen in the Ruby 3.3 build in CI [here][1], since `test/sidekiqmon_test.rb` has `require "ostruct"`, but `ostruct` is not specified in the Gemfile or gemspec as a gem dependency.

[1]: https://github.com/sidekiq/sidekiq/actions/runs/10710835954/job/29698382673#step:6:7

This is the warning:

> /home/runner/work/sidekiq/sidekiq/test/sidekiqmon_test.rb:4: warning: ostruct was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
>
> You can add ostruct to your Gemfile or gemspec to silence this warning.

In order to get rid of that warning when running tests, this change removes all usage of `OpenStruct` from Sidekiq. There were only a few such usages, all in `test/sidekiqmon_test.rb`. This change replaces those `OpenStruct` usages with `Struct`, and removes the `require "ostruct"` from the top of that file.

#6420 is an alternative possible solution. If that PR is merged, then this PR should be closed without merging.

# Screenshots

_[Running tests on Ruby **3.3.5**]_

## Before

image

## After

image

### DIFF
--- a/test/sidekiqmon_test.rb
+++ b/test/sidekiqmon_test.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require_relative "helper"
-require "ostruct"
 require "sidekiq/monitor"
 
 def capture_stdout
@@ -42,7 +41,7 @@ describe Sidekiq::Monitor do
       end
 
       it "displays the correct output" do
-        mock_stats = OpenStruct.new(
+        stats_attributes = {
           processed: 420710,
           failed: 12,
           workers_size: 34,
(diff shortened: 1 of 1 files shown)


mperham:
Why did you submit #6240 also? Are these two mutually exclusive?

davidrunger:
@mperham : Sorry, yes; the two PRs are mutually exclusive. (I had buried a mention of this in an edit that I made to my PR description; maybe you saw the PR description without that edit; sorry again!)

> https://github.com/sidekiq/sidekiq/pull/6420 is an alternative possible solution. If that PR is merged, then this PR should be closed without merging.

PR #6430, opened 2024-09-15
Add Active Job adapter

### LINKED ISSUE #6429: Implement active job adapter
Discussion - https://github.com/rails/rails/issues/52929

I can help implementing this, if agreed.

fatkodima:
Closes #6429.

### DIFF
--- a/lib/active_job/queue_adapters/sidekiq_adapter.rb
+++ b/lib/active_job/queue_adapters/sidekiq_adapter.rb
@@ -5,7 +5,7 @@ module Sidekiq
   module ActiveJob
     # @api private
     class Wrapper
-      include Sidekiq::Worker
+      include Sidekiq::Job
 
       def perform(job_data)
         ::ActiveJob::Base.execute(job_data.merge("provider_job_id" => jid))
--- a/lib/active_job/queue_adapters/sidekiq_adapter.rb
+++ b/lib/active_job/queue_adapters/sidekiq_adapter.rb
@@ -1,0 +1,76 @@
+# frozen_string_literal: true
+
+module ActiveJob
+  module QueueAdapters
+    # Sidekiq adapter for Active Job
+    #
+    # To use Sidekiq set the queue_adapter config to +:sidekiq+.
+    #
+    #   Rails.application.config.active_job.queue_adapter = :sidekiq
+    class SidekiqAdapter
+      # Defines whether enqueuing should happen implicitly to after commit when called
+      # from inside a transaction.
+      # @api private
+      def enqueue_after_transaction_commit?
+        true
+      end
+
+      # @api private
+      def enqueue(job)
+        job.provider_job_id = JobWrapper.set(
+          wrapped: job.class,
+          queue: job.queue_name
+        ).perform_async(job.serialize)
+      end
+
+      # @api private
+      def enqueue_at(job, timestamp)
+        job.provider_job_id = JobWrapper.set(
+          wrapped: job.class,
+          queue: job.queue_name
+        ).perform_at(timestamp, job.serialize)
+      end
+
+      # @api private
+      def enqueue_all(jobs)
+        enqueued_count = 0
+        jobs.group_by(&:class).each do |job_class, same_class_jobs|
+          same_class_jobs.group_by(&:queue_name).each do |queue, same_class_and_queue_jobs|
+            immediate_jobs, scheduled_jobs = same_class_and_queue_jobs.partition { |job| job.scheduled_at.nil? }
+
+            if immediate_jobs.any?
+              jids = Sidekiq::Client.push_bulk(
+                "class" => JobWrapper,
+                "wrapped" => job_class,
+                "queue" => queue,
+                "args" => immediate_jobs.map { |job| [job.serialize] }
+              )
+              enqueued_count += jids.compact.size
+            end
+
+            if scheduled_jobs.any?
+              jids = Sidekiq::Client.push_bulk(
+                "class" => JobWrapper,
+                "wrapped" => job_class,
+                "queue" => queue,
+                "args" => scheduled_jobs.map { |job| [job.serialize] },
+                "at" => scheduled_jobs.map { |job| job.scheduled_at&.to_f }
+              )
+              enqueued_count += jids.compact.size
+            end
+          end
+        end
+        enqueued_count
+      end
+
+      # @api private
+      class JobWrapper
+        include Sidekiq::Worker
+
+        def perform(job_data)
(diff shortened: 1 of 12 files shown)


lib/sidekiq/rails.rb
```diff
@@ -3,6 +3,12 @@
 require "sidekiq/job"
 require "rails"
 
+begin
+  require "active_job"
+  require "active_job/queue_adapters/sidekiq_adapter"
```
Earlopain, line 8: This will result in method redefinition warnings if running with a rails version that still ships the adapter, right? Would it make sense to add `gem "active_job", ">= 8.0.0"` to prevent this?

fatkodima: Rails implicitly `autoload`s this class (https://github.com/rails/rails/blob/2f55eba02511ba3f79e4771305f14c85b1f7caf2/activejob/lib/active_job/queue_adapters.rb#L123) while we explicitly `require` it, so the one from rails will not be loaded and the redefinition should not happen?

Earlopain: I didn't realize it works that way. I tested it out with your branch and it indeed seems fine.

mperham:
The `remove_const` bit is subtle but necessary when using older versions of Rails.

Do we want to re-home the class? I don't particularly care for the long name, especially since that class name is serialized into every single job payload. `ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper` is entirely too long, but it's [trivial to create a class alias](https://github.com/sidekiq/sidekiq/wiki/FAQ#how-do-i-safely-rename-a-worker).

```ruby
> "ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper".size
=> 52
> "Sidekiq::ActiveJob::Wrapper".size
=> 27
```

fatkodima:
Updated to explicitly remove the existing adapter and use shorter wrapper job class name.

lib/active_job/queue_adapters/sidekiq_adapter.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+# Define a new class whose name is shorter than from Active Job.
+module Sidekiq
+  module ActiveJob
+    # @api private
+    class Wrapper
+      include Sidekiq::Worker
```
mperham, line 8: S::Job

lib/active_job/wrapper.rb
mperham, line 1: What's this?

fatkodima: oops, deleted now

mperham:
One further thing: there are a number of places where we check for the AJ Wrapper class name. We'll need to add the new name to that logic. See e.g. sidekiq/api.

fatkodima:
Ah, right. There are also a few references to the old class name in tests, but I think leaving them unchanged is fine?

mperham:
I think it's mostly fine but I would update at least one api test to use the new name, just to test that both names are working as expected.

If the user is running Rails 7.0, will this use our adapter or AJ's internal adapter? Do we need to maintain compatibility with the old name as long as we support Rails 7.x? Or does remove_const solve that issue?

fatkodima:
Added a test to api_test.rb.

> If the user is running Rails 7.0, will this use our adapter or AJ's internal adapter?

It will always use our adapter. We explicitly disable the existing adapter, if any, by `remove_const`.

test/api_test.rb
```diff
@@ -32,6 +32,10 @@ class JobWithTags
...
+  ],
+  "8.x" => [
+    '{"class":"Sidekiq::ActiveJob::Wrapper","wrapped":"ApiAjJob","queue":"default","args":[{"job_class":"ApiAjJob","job_id":"37649eb0-c437-4e00-8a29-85cc12da1440","provider_job_id":null,"queue_name":"default","priority":null,"arguments":[1,2,3],"executions":0,"exception_executions":{},"locale":"en","timezone":null,"enqueued_at":"2024-09-19T16:39:26.609737000Z","scheduled_at":null}],"retry":true,"jid":"ec42f101ed1b16f27c6a6188","created_at":1726763966.610073,"enqueued_at":1726763966.6101718}',
+    '{"class":"Sidekiq::ActiveJob::Wrapper","wrapped":"ActionMailer::MailDeliveryJob","queue":"mailers","args":[{"job_class":"ActionMailer::MailDeliveryJob","job_id":"d6573e12-dd78-454a-83d5-67df94934c82","provider_job_id":null,"queue_name":"mailers","priority":null,"arguments":["ApiMailer","test_email","deliver_now",{"args":[1,2,3],"_aj_ruby2_keywords":["args"]}],"executions":0,"exception_executions":{},"locale":"en","timezone":null,"enqueued_at":"2024-09-19T16:45:38.673195000Z","scheduled_at":null}],"retry":true,"jid":"2cd7874f7651115f453d6315","created_at":1726764338.673335,"enqueued_at":1726764338.673404}'
```
zzak, line 38: @mperham Is the new class name the one that gets serialized? Won't that explode if you have a rolling deploy (old worker tries to deserialize the new const) or a rollback?

mperham: We don’t support old and new versions running at the same time. It would break a rollback though.

zzak: I would be surprised to find a change that would break in a rollback to make it into a patch-level release without proper warning. Why not keep the old name and deprecate it?

mperham: What we can do is stick with the existing name and add the new name as an existing alias. Then switch the default name in 8.0 so both old and new names would be supported in 7.x and 8.0.

lib/active_job/queue_adapters/sidekiq_adapter.rb
```diff
@@ -0,0 +1,88 @@
...
+    #
+    # To use Sidekiq set the queue_adapter config to +:sidekiq+.
+    #
+    #   Rails.application.config.active_job.queue_adapter = :sidekiq
+    class SidekiqAdapter
+      # Defines whether enqueuing should happen implicitly to after commit when called
+      # from inside a transaction.
+      # @api private
+      def enqueue_after_transaction_commit?
+        true
```
palexvs, line 32: @mperham does it mean we don't need `Sidekiq.transactional_push! (after_commit_everywhere)` in Rails 7.2 and newer? Should https://github.com/sidekiq/sidekiq/wiki/Advanced-Options#transactional-push be updated?

mperham: I'm assuming that the Rails support only applies to Active Jobs. Sidekiq::Jobs still need after_commit_everywhere because Sidekiq has no required dependencies on Rails, SQL databases or transactions.

PR #6434, opened 2024-09-20
Stick with old ActiveJob adapter name

zzak:
Follow up to #6430

### DIFF
--- a/lib/active_job/queue_adapters/sidekiq_adapter.rb
+++ b/lib/active_job/queue_adapters/sidekiq_adapter.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-# Define a new class whose name is shorter than from Active Job.
 module Sidekiq
   module ActiveJob
     # @api private
@@ -34,7 +33,7 @@ module ActiveJob
 
       # @api private
       def enqueue(job)
-        job.provider_job_id = Sidekiq::ActiveJob::Wrapper.set(
+        job.provider_job_id = JobWrapper.set(
           wrapped: job.class,
           queue: job.queue_name
         ).perform_async(job.serialize)
@@ -42,7 +41,7 @@ module ActiveJob
 
       # @api private
       def enqueue_at(job, timestamp)
-        job.provider_job_id = Sidekiq::ActiveJob::Wrapper.set(
+        job.provider_job_id = JobWrapper.set(
           wrapped: job.class,
           queue: job.queue_name
         ).perform_at(timestamp, job.serialize)
@@ -57,7 +56,7 @@ module ActiveJob
 
             if immediate_jobs.any?
               jids = Sidekiq::Client.push_bulk(
-                "class" => Sidekiq::ActiveJob::Wrapper,
+                "class" => JobWrapper,
                 "wrapped" => job_class,
                 "queue" => queue,
                 "args" => immediate_jobs.map { |job| [job.serialize] }
@@ -67,7 +66,7 @@ module ActiveJob
 
             if scheduled_jobs.any?
               jids = Sidekiq::Client.push_bulk(
-                "class" => Sidekiq::ActiveJob::Wrapper,
+                "class" => JobWrapper,
                 "wrapped" => job_class,
                 "queue" => queue,
                 "args" => scheduled_jobs.map { |job| [job.serialize] },
@@ -82,7 +81,8 @@ module ActiveJob
 
       # Defines a class alias for backwards compatibility with enqueued Active Job jobs.
       # @api private
-      JobWrapper = Sidekiq::ActiveJob::Wrapper
+      class JobWrapper < Sidekiq::ActiveJob::Wrapper
+      end
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/active_job/queue_adapters/sidekiq_adapter.rb
```diff
@@ -1,19 +1,5 @@
 # frozen_string_literal: true
 
-# Define a new class whose name is shorter than from Active Job.
```
fatkodima, line 3: To make the PR diff shorter, I would left this class as is and just change a few lines down in the class `"class" => Sidekiq::ActiveJob::Wrapper,` to use the old name (note that this change is needed in any case, because we still enqueue the new class name).

zzak: Not sure I follow, but if we want to alias the constant it has to be defined first, right?

fatkodima: I mean the code change for this PR will be only something like 
```diff
diff --git a/lib/active_job/queue_adapters/sidekiq_adapter.rb b/lib/active_job/queue_adapters/sidekiq_adapter.rb
index 2bfa91d6..ad14cb9e 100644
--- a/lib/active_job/queue_adapters/sidekiq_adapter.rb
+++ b/lib/active_job/queue_adapters/sidekiq_adapter.rb
@@ -34,7 +34,7 @@ module ActiveJob

       # @api private
       def enqueue(job)
-        job.provider_job_id = Sidekiq::ActiveJob::Wrapper.set(
+        job.provider_job_id = JobWrapper.set(
           wrapped: job.class,
           queue: job.queue_name
         ).perform_async(job.serialize)
@@ -42,7 +42,7 @@ module ActiveJob

       # @api private
       def enqueue_at(job, timestamp)
-        job.provider_job_id = Sidekiq::ActiveJob::Wrapper.set(
+        job.provider_job_id = JobWrapper.set(
           wrapped: job.class,
           queue: job.queue_name
         ).perform_at(timestamp, job.serialize)
@@ -57,7 +57,7 @@ module ActiveJob

             if immediate_jobs.any?
               jids = Sidekiq::Client.push_bulk(
-                "class" => Sidekiq::ActiveJob::Wrapper,
+                "class" => JobWrapper,
                 "wrapped" => job_class,
                 "queue" => queue,
                 "args" => immediate_jobs.map { |job| [job.serialize] }
@@ -67,7 +67,7 @@ module ActiveJob

             if scheduled_jobs.any?
               jids = Sidekiq::Client.push_bulk(
-                "class" => Sidekiq::ActiveJob::Wrapper,
+                "class" => JobWrapper,
                 "wrapped" => job_class,
                 "queue" => queue,
                 "args" => scheduled_jobs.map { |job| [job.serialize] },
```

And then in the sidekiq 8.0 we will remove that `JobWrapper = Sidekiq::ActiveJob::Wrapper` alias and switch back to enqueuing shorter job names via `"class" => Sidekiq::ActiveJob::Wrapper`.

zzak: Oh yeah, good catch. Thought you meant the diff of moving this module to the bottom. Thanks for the review!

lib/active_job/queue_adapters/sidekiq_adapter.rb
```diff
@@ -82,7 +68,21 @@ def enqueue_all(jobs)
 
       # Defines a class alias for backwards compatibility with enqueued Active Job jobs.
       # @api private
-      JobWrapper = Sidekiq::ActiveJob::Wrapper
+      class JobWrapper
```
fatkodima, line 71: Would also let this class to be where it is currently, because when we will remove the alias in 8.0, we will need to move it back again where it currently was before this PR.

So, literally only the changes suggested in https://github.com/sidekiq/sidekiq/pull/6434/files#r1768226056 are needed for this file, I think

fatkodima (approved):

zzak:
@fatkodima This is still the same problem I wanted to avoid for rollbacks:

```
2024-09-20T09:21:28.438Z pid=549985 tid=bfep WARN: {"context":"Job raised exception","job":{"retry":true,"queue":"default","wrapped":"AddJob","args":[{"job_class":"AddJob","job_id":"24c65e1a-0b82-489b-bdb7-50b348e2dc5e","provider_job_id":null,"queue_name":"default","priority":null,"arguments":[20,39,"from patched"],"executions":0,"exception_executions":{},"locale":"en","timezone":"UTC","enqueued_at":"2024-09-20T09:18:58.461103550Z","scheduled_at":null}],"class":"Sidekiq::ActiveJob::Wrapper","jid":"b104661be32913e586b41682","created_at":1726823938.4613,"enqueued_at":1726823938.4613526}}
2024-09-20T09:21:28.438Z pid=549985 tid=bfep WARN: NameError: uninitialized constant Sidekiq::ActiveJob
```

zzak:
> Can you describe the steps that led to this situation?

I just have a dummy `rails new` app with a single job.

`gem "sidekiq", path: "~/code/sidekiq"`

Then enqueue a job.

Now remove the `path` and use the gem.

If I then start sidekiq to process jobs I will get that error.

zzak:
@fatkodima That works, thanks for helping figure this out!

fatkodima (approved):
Thanks for working on this! ❤️

PR #6437, opened 2024-09-22
Add missing uk translations

fatkodima:


mperham:
For a second I was expecting British English. 😂

fatkodima:
I always thought that 'uk' is United Kingdom, because I always saw 'ua' as a shorthand for Ukraine. But then learned that actually 'uk' is used for Ukraine 😃

PR #4103, opened 2019-02-22
add back the dropdown css

urkle:
@mperham here is the PR that add only the .caret and .dropdown css pieces to the bootstrap.

(new classes added)
```css
.caret {
    display: inline-block;
    width: 0;
    height: 0;
    margin-left: 2px;
    vertical-align: middle;
    border-top: 4px dashed;
    border-top: 4px solid \9;
    border-right: 4px solid transparent;
    border-left: 4px solid transparent
}

.dropup, .dropdown {
    position: relative
}

.dropdown-toggle:focus {
    outline: 0
}

.dropdown-menu {
    position: absolute;
    top: 100%;
    left: 0;
    z-index: 1000;
    display: none;
    float: left;
    min-width: 160px;
    padding: 5px 0;
    margin: 2px 0 0;
    font-size: 14px;
    text-align: left;
    list-style: none;
    background-color: #fff;
    -webkit-background-clip: padding-box;
    background-clip: padding-box;
    border: 1px solid #ccc;
    border: 1px solid rgba(0, 0, 0, 0.15);
    border-radius: 4px;
    -webkit-box-shadow: 0 6px 12px rgba(0, 0, 0, 0.175);
    box-shadow: 0 6px 12px rgba(0, 0, 0, 0.175)
}

.dropdown-menu.pull-right {
    right: 0;
    left: auto
}

.dropdown-menu .divider {
    height: 1px;
    margin: 9px 0;
    overflow: hidden;
    background-color: #e5e5e5
}

.dropdown-menu > li > a {
    display: block;
    padding: 3px 20px;
    clear: both;
    font-weight: 400;
    line-height: 1.42857143;
    color: #333;
    white-space: nowrap
}

.dropdown-menu > li > a:hover, .dropdown-menu > li > a:focus {
    color: #262626;
    text-decoration: none;
    background-color: #f5f5f5
}

.dropdown-menu > .active > a, .dropdown-menu > .active > a:hover, .dropdown-menu > .active > a:focus {
    color: #fff;
    text-decoration: none;
    background-color: #337ab7;
    outline: 0
}

.dropdown-menu > .disabled > a, .dropdown-menu > .disabled > a:hover, .dropdown-menu > .disabled > a:focus {
    color: #777
}

.dropdown-menu > .disabled > a:hover, .dropdown-menu > .disabled > a:focus {
    text-decoration: none;
    cursor: not-allowed;
    background-color: transparent;
    background-image: none;
[...]

PR #6451, opened 2024-10-07
fix(web): replace Bootstrap 3.3.7 with 5.3.3

tnir:
## Background

While Sidekiq Web UI has used Boostrap 3.3.7 since #3117 (2016; 8 years ago), 5 years passed after the EOL of Bootstrap 3.

> Bootstrap 3 reached end of life July 24, 2019, followed by Bootstrap 4 on January 1, 2023. As such, neither version receives new features, updates, or security fixes. 

https://getbootstrap.com/docs/4.6/end-of-life/

## Changes

- Replaces Bootstrap 3.3.7 with 5.3.3.

## Refs.

- #3117
- #4103
- Bootstrap's migration guides:
   - https://getbootstrap.com/docs/4.6/migration/
   - https://getbootstrap.com/docs/5.3/migration/


mperham:
Testing:

- [ ] RTL languages
- [ ] Dark/light mode
- [ ] Pro/Ent support
- [ ] Live Polling

mperham:
Screenshot 2024-10-07 at 9 40 40 AM

mperham:
RTL looks good enough. Dark mode ❌.

Screenshot 2024-10-07 at 10 21 22 AM

mperham:
Minor issues with Pro/Ent which I can fix, e.g. rate limiters filtering widget.

mperham:
Thank you for doing this work. I attempted this upgrade last Spring and it got overwhelming as I'm not good with CSS/styling so this is a huge lift for me.

web/views/_poll_link.erb
```diff
@@ -1,4 +1,4 @@
 <% if pollable? %>
     <a class="live-poll-start live-poll btn btn-primary"><%= t('LivePoll') %></a>
-    <a class="live-poll-stop live-poll btn btn-primary active"><%= t('StopPolling') %></a>
+    <a class="live-poll-stop live-poll btn btn-primary active" style="display: none"><%= t('StopPolling') %></a>
```
markets, line 3: @tnir @mperham I think we should remove inline styles: #6292

markets:
@mperham I understand this change, as Sidekiq needs to move forward also regarding its front-end dependencies, but my question is: How this can be handled by plugins with UI? I'm the maintainer of [`sidekiq-cron`](https://github.com/sidekiq-cron/sidekiq-cron) (one of the most used Sidekiq plugins) and for the views we are just reusing current Bootstrap classes, example: https://github.com/sidekiq-cron/sidekiq-cron/blob/master/lib/sidekiq/cron/views/cron.erb. So after this Bootstrap update, it will really hard to maintain compatibility with different Sidekiq versions.

mperham:
@markets I appreciate your compatibility concerns. We'll get this into Sidekiq 8.0 so it is part of a major version change. If you want to maintain compatibility with both 7 and 8 in the same package, that's up to you. You only have two pages so you could provide two templates, one for 7 and one for 8, and render based on Sidekiq::MAJOR.

markets:
Yes, the only reasonable way I can imagine right now is to provide two templates. Out of curiosity, do you have any ETA for the v8 release? Depending on the changes it will be probably better to jump a major version on our side too. We already implemented support for the new UI assets API (https://github.com/sidekiq-cron/sidekiq-cron/pull/480), but that was easy to make it work with different Sidekiq versions.

Thanks Mike!

mperham:
> do you have any ETA for the v8 release?

Not in 2024. 2025ish. 😂

mperham:
There's still many noticeable regressions in the look and feel from stock. Here's stock:

stock

and here's this PR:

pr

I hate to nitpick contributions but can you go through the pages Stock vs PR and clean up as many differences as possible? Part of the reason why Sidekiq is so popular is that we care about this level of detail.

mperham:
The two major issues in those screenshots are the upper navbar tab highlighting and the smaller action buttons in the Processes table. Hopefully easy fixes.

mperham:
See #6549 and #6551

PR #6475, opened 2024-10-24
Dup current attributes on saving

fatkodima:
See https://github.com/rails/rails/pull/53440 for the problem description.

### DIFF
--- a/lib/sidekiq/middleware/current_attributes.rb
+++ b/lib/sidekiq/middleware/current_attributes.rb
@@ -33,10 +33,25 @@ module Sidekiq
             attrs = strklass.constantize.attributes
             # Retries can push the job N times, we don't
             # want retries to reset cattr. #5692, #5090
-            job[key] = attrs if attrs.any?
+            if attrs.any?
+              # Older rails has a bug that `CurrentAttributes#attributes` always returns
+              # the same hash instance. We need to dup it to avoid being accidentally mutated.
+              job[key] = if returns_same_object?
+                attrs.dup
+              else
+                attrs
+              end
+            end
           end
         end
         yield
+      end
+
+      private
+
+      def returns_same_object?
+        ActiveSupport::VERSION::MAJOR < 8 ||
+          (ActiveSupport::VERSION::MAJOR == 8 && ActiveSupport::VERSION::MINOR == 0)
       end
     end
 


mperham:
I think it's reasonable to `dup` as we really want to serialize this data. I am concerned that if we are `push_bulk`ing 10,000 jobs, we'll create 10,000 copies of the Hash but I guess that's unavoidable. What if we serialize the jobs at the time of `push` failure and send the serialized copy when recovering?

fatkodima:
The current approach looks the simplest to me. 

> What if we serialize the jobs at the time of push failure and send the serialized copy when recovering?

This needs to be implemented in the sidekiq-pro, right?

mperham:
Yes, but I'm ok with that. This is a small but subtle improvement to reliability.

fatkodima:
Given that the rails PR is about to be merged (https://github.com/rails/rails/pull/53440#issuecomment-2436384115) and this will now return a copy on each call, wdyt about making a conditional check on rails version and dup if needed?

mperham:
I just gave you access since you've been a great collaborator but I want to be clear that I don't expect free labor from you. I'm happy to pay for PRs to Pro/Ent, I'm not sure how to do this but I'm open to it.

mperham:
If they are going to dup, let's just leave things as is. I'd say upgrade Rails to get the new behavior.

fatkodima:
Thanks, accepted. No worries, I don't need payments for the great OSS projects I use 😄.

I would still prefer to maybe have this PR given how simple it is, because rails PR will not be backported and will only be in the yet to be released rails version. Most people are not on the latest versions and can have this issue unnoticed. We are currently on latest rails 7.2 and I don't know when we will be on 8.0, but this already caused an issue for us for the job to ran on the incorrect shard (shard was configured by `Current`). We can avoid the potential performance penalty by adding a check for rails version.

mperham:
I guess I will have Sidekiq Pro users on Rails < 8 for many years to come so it's best to fix this now.

fatkodima:
So should I reopen and finish this PR (preferred) or create a new in the sidekiq-pro?

mperham:
I'm ok using this PR.

fatkodima:
Updated. I was incorrect - it will be rails 8.1, which will be released in half a year from now or even more.

mperham:
And in rethinking about this, we really do want to serialize this data via `dup` when the middleware is run, because the payload push can be delayed and CurrentAttributes mutated during that time. Pro's reliable push delays it occasionally but transactional push delays it *every time*.

This could be a long-time source of mysterious bugs for Sidekiq users. I see this as a pretty high priority fix for anyone using transactional_push.

mperham:
Oh wait, I'm wrong. The middleware doesn't run until the transaction commits anyways so it's not an issue with transactional_push.

PR #6477, opened 2024-10-27
Fix circular require for ActiveJob integration

fatkodima:
Fixes https://github.com/rails/rails/issues/53460 (cc @zzak)

We have a circular require error
https://github.com/rails/rails/blob/e6429269fd5a8a7d728557f4e7d7f82c0ad1478c/activejob/lib/active_job/queue_adapters/sidekiq_adapter.rb#L4
https://github.com/sidekiq/sidekiq/blob/bd3604909687bb0a0f4e71e087eb5a75e69fdf83/lib/sidekiq/rails.rb#L8

### DIFF
--- a/lib/sidekiq/rails.rb
+++ b/lib/sidekiq/rails.rb
@@ -3,12 +3,6 @@
 require "sidekiq/job"
 require "rails"
 
-begin
-  require "active_job"
-  require "active_job/queue_adapters/sidekiq_adapter"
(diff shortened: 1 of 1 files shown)


zzak:
@fatkodima Thanks for fixing, I'm not able to reproduce this outside of Rails tests, but I think that's probably fine. Used this same technique for the Resque adapter as well. :+1:

PR #6485, opened 2024-10-31
Production job profiling with Vernier

mperham:
Work in progress. Needs web testing and integration with the Firefox viewer.

See https://github.com/sidekiq/sidekiq/wiki/Profiling for usage.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -2,6 +2,13 @@
 
 [Sidekiq Changes](https://github.com/sidekiq/sidekiq/blob/main/Changes.md) | [Sidekiq Pro Changes](https://github.com/sidekiq/sidekiq/blob/main/Pro-Changes.md) | [Sidekiq Enterprise Changes](https://github.com/sidekiq/sidekiq/blob/main/Ent-Changes.md)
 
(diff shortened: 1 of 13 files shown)


mperham:
How to get a profile:

0. Start Redis. In three terminal windows in `myapp/`:
1. 
```
bundle
bundle exec sidekiq -v
```
2.
```
bundle exec rails c
> HardJob.set(profile: "mike").perform_async(123, "mike", nil)
```
3.
```
bundle exec rails s
open http://localhost:3000/sidekiq/profiles
```

Is there a way to script the terminal setup, sort of like `foreman`, so the terminal input in (2) is usable?

tisba:
It might be worth mentioning in https://github.com/sidekiq/sidekiq/wiki/Profiling#profile-data-and-security that `Sidekiq::Web::PROFILE_OPTIONS[:view_url]` can be used to prevent accidental data exposure. Especially big organizations are rather paranoid when it comes to this stuff.

mperham:
Thanks for the feedback. I've incorporated both suggestions.

PR #6494, opened 2024-11-08
Correctly remove the buildin rails adapter

Earlopain:
Followup #6474. cc @leoarnold 

`defined?` for a string literal is always truthy, same as for symbols. `defined?(ActiveJob::QueueAdapters::SidekiqAdapter)` would work as an alternative.

```rb
module ActiveJob
  module QueueAdapters
    # Adapter removed from rails
  end
end

module ActiveJob
  module QueueAdapters
    remove_const(:SidekiqAdapter) if defined?("::#{name}::SidekiqAdapter")
  end
end
```

`test.rb:10:in 'remove_const': constant ActiveJob::QueueAdapters::SidekiqAdapter not defined (NameError)`

### DIFF
--- a/lib/active_job/queue_adapters/sidekiq_adapter.rb
+++ b/lib/active_job/queue_adapters/sidekiq_adapter.rb
@@ -16,7 +16,7 @@ end
 module ActiveJob
   module QueueAdapters
     # Explicitly remove the implementation existing in older rails'.
-    remove_const(:SidekiqAdapter) if defined?("::#{name}::SidekiqAdapter")
+    remove_const(:SidekiqAdapter) if const_defined?(:SidekiqAdapter)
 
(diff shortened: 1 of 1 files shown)


mperham:
What if ActiveJob::SidekiqAdapter is defined? Will it remove it?

Earlopain:
> What if ActiveJob::SidekiqAdapter is defined? Will it try to remove it and raise?

Doesn't look like it:
```rb
module ActiveJob
  module SidekiqAdapter
  end
end

module ActiveJob
  module QueueAdapters
    remove_const(:SidekiqAdapter) if const_defined?(:SidekiqAdapter)
  end
end

# Survived the `remove_const` and the other one still is there as well
pp ActiveJob::SidekiqAdapter # => ActiveJob::SidekiqAdapter
```

PR #6497, opened 2024-11-09
Sidekiq Web: Apply filters on GET requests

franzliedke:
This tackles a small paper cut I've been feeling on Sidekiq's useful Web UI: Applying filters, e.g. to the "Metrics" tab, changes the URL, which breaks reloads. Also, it's not possible to share a permalink to a pre-filtered list.

On the web, filtering a list is usually done with GET requests, using query string parameters. Such an operation does not have side effects, so it can be considered [safe](https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP).

---

Previous, related changes:
- 1b9f7db4c5d0dc41b6eac2682710715eed5aadce
- #6440

---

To test the waters, I only applied this change to the "Metrics" tab. If you like the idea, I will happily and quickly expand the scope to all other tabs with filtering.

### DIFF
--- a/lib/sidekiq/web/application.rb
+++ b/lib/sidekiq/web/application.rb
@@ -67,12 +67,16 @@ module Sidekiq
     end
 
     get "/metrics" do
+      x = params[:substr]
+      class_filter = (x.nil? || x == "") ? nil : Regexp.new(Regexp.escape(x), Regexp::IGNORECASE)
+
       q = Sidekiq::Metrics::Query.new
       @period = h((params[:period] || "")[0..1])
       @periods = METRICS_PERIODS
       minutes = @periods.fetch(@period, @periods.values.first)
-      @query_result = q.top_jobs(minutes: minutes)
-      erb(:metrics)
+      @query_result = q.top_jobs(minutes: minutes, class_filter: class_filter)
+
+      erb :metrics
     end
 
     get "/metrics/:name" do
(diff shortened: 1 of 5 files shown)


web/views/metrics.erb
```diff
@@ -9,8 +9,7 @@
   </div>
 
   <div>
-    <form id="metrics-form" class="form-inline" action="<%= root_path %>filter/metrics" method="post">
-      <%= csrf_tag %>
```
franzliedke, line 13: CSRF checks are skipped on GET requests anyway, so the tag is not needed.

Was there any explicit intention behind filtering only with POST?
Is there any risk (e.g. filter operations being super heavy on Redis) that I am missing?

mperham:
I’m ok with all of this. Make sure you sanitize all inputs.

web/views/metrics.erb
mperham: I don’t recall any specific reason. It’s likely just copied from the retry/schedule filtering.

lib/sidekiq/web/application.rb
```diff
@@ -153,9 +157,15 @@ def self.set(key, val)
...
-      @dead = @dead.map { |msg, score| Sidekiq::SortedEntry.new(nil, score, msg) }
+      x = params[:substr]
+
+      if x && x != ""
+        @dead = search(Sidekiq::DeadSet.new, x)
+      else
+        @count = (params["count"] || 25).to_i
+        (@current_page, @total_size, @dead) = page("dead", params["page"], @count, reverse: true)
+        @dead = @dead.map { |msg, score| Sidekiq::SortedEntry.new(nil, score, msg) }
+      end
```
franzliedke, line 168: This logic looks a little bit inconsistent, but I went for the least invasive change, so it should work exactly like before.

If I tested things correctly, these three filter routes boil down to filtering just the job ID, so usually there's just one result. I guess that's why there was no "real" pagination logic here? 🤔

franzliedke:
@mperham Alright, I adjusted the remaining routes. Thanks for the quick feedback! 🙌🏼

lib/sidekiq/web/application.rb
mperham: Pagination implies more state to remember across page loads which I wasn’t willing to build. If there are 1000 results, they’ll get a table with 1000 rows.

franzliedke: I know very little about direct Redis interaction, but [`ZSCAN`](https://redis.io/docs/latest/commands/zscan/) seems to support passing a cursor argument, so it should be possible, right? (Possible follow-up PR material, hence the question.)

mperham: Since the set is changing all the time, I didn’t want to implement cursor based filtering.

Issue #6516, opened 2024-11-25
Web UI does not take advantage of full width of screen


stimms:
Ruby version: 3.2.6
Rails version: 7.0.8.6
Sidekiq / Pro / Enterprise version(s): 6.5.12/5.5.8/2.5.3

The web UI doesn't take advantage of a larger screen (in excess of 1200px side) as a result columns like Arguments are crowded and unreadable. 

![image](https://github.com/user-attachments/assets/6b05c19e-4f74-44fe-a356-1c8af742b53a)

mperham:
Is the Dashboard 100% width "safe", so to speak? Do the graphs look ok?

stimms:
They looked fine to me! Sorry my screenshots are bit weird I'm trying not
to leak too much about our volume.

[image: image.png]

mperham:
I don't think attachments come through email replies, I don't see the image on the web.

stimms:
huh, computers...

![image](https://github.com/user-attachments/assets/8f99f95e-b31a-4770-959e-6a318c35090b)

mperham:
This is fixed by the 8.0 UI restyle in #6551

PR #6517, opened 2024-11-25
Allow full width for web console

stimms:
Addresses #6516 

Before 
image

After 
image


mperham:
I'm not convinced this is a good idea, the UI is not meant to scale to infinite horizontal width.

stimms:
I don't see any real drawback to allowing width to be full screen. If it is too wide for folks it's much easier to shrink the window than to have to fiddle with CSS in an inspector to get the container wider. 

Is there a middle ground we can settle on? It is frustrating to have 5000 pixels of monitor width and still be really struggling to see what is in the arguments column while a thousand pixels of gutter space go unused on each side. Perhaps we could agree to another couple of width breakpoints at say ≥2560px and ≥3840px.

mperham:
No one runs their browser full width at those resolutions, be real. If you need the arguments, click on the row and view the details page which has a lot more space for display. Sidekiq even explicitly recommends keeping arguments small. But I’m not going to fundamentally change the UI layout for this.

PR #6523, opened 2024-11-26
Add emoji flag to locale dropdown

### LINKED ISSUE #6520: Add emoji flag to locale dropdown
Add the corresponding emoji to the locale code, will make it easier for all to find their language.

fatkodima:
Closes #6520.

Screenshot 2024-11-26 at 15 08 16

.
.
Links that were useful:
* [List of locales](https://help.sap.com/docs/SAP_BUSINESSOBJECTS_BUSINESS_INTELLIGENCE_PLATFORM/09382741061c40a989fae01e61d54202/46758c5e6e041014910aba7db0e91070.html)
* [List of country codes](https://en.wikipedia.org/wiki/ISO_3166-1)
* https://dev.to/ionbazan/turn-a-country-code-into-an-emoji-flag-us--360a

### DIFF
--- a/lib/sidekiq/web/helpers.rb
+++ b/lib/sidekiq/web/helpers.rb
@@ -95,6 +95,10 @@ module Sidekiq
       locale_files.select { |file| file =~ /\/#{lang}\.yml$/ }
     end
 
+    def language_name(locale)
+      strings(locale).fetch("LanguageName", locale)
+    end
+
     def search(jobset, substr)
       resultset = jobset.scan(substr).to_a
       @current_page = 1
--- a/test/web_helpers_test.rb
+++ b/test/web_helpers_test.rb
@@ -110,6 +110,13 @@ describe "Web helpers" do
     assert_equal "en", obj.locale
   end
 
+  it "returns language name from locale" do
+    obj = Helpers.new
+    assert_equal "English", obj.language_name("en")
+    assert_equal "Português", obj.language_name("pt")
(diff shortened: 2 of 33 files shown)


mperham:
Should English use the flag of England? 🤔 I can see this becoming problematic since language != country. Which flag should Arabic use?

Ultimately I'm trying to make it easier for the user to locate their native language in the dropdown.
Maybe we should show the native name of each language: Español rather than Spanish?

fatkodima:
> Should English use the flag of England?

I don't know, I saw both USA and England flags be used for English in my life 🤷 For example, amazon https://www.amazon.com uses USA flag.

> Which flag should Arabic use?

Actually I see in this PR that it is rendered as a `AR`gentina. Maybe look at what people use for Arabic as a flag and use it?

> Maybe we should show the native name of each language: Español rather than Spanish?

We currently show only 2 letter language abbreviations, how these are specified by the ISO. But we can implement something like in the image 
Screenshot 2024-11-26 at 19 58 49

mperham:
I think the flags are a mistake, as they don't map cleanly. Reading a few web pages on language picker design, I think V.02 here is the right approach, as does the author:

https://medium.com/design-bootcamp/how-to-build-a-thoughtful-website-language-selector-d30ae2f1910f

fatkodima:
Current look
Screenshot 2024-11-26 at 21 19 55

https://gist.github.com/SimonEast/952095c9c8e4b103b47604d47c6012bd was useful to get language names in the languages themselves.

PR #6532, opened 2024-11-27
Sidekiq::Web internal refactoring

mperham:
Re-implement Sidekiq::Web to remove some gross hacks and more questionable bits. 

## Configuration

This refactoring brings some DRY sanity to Sidekiq::Web configuration.

### Before

```ruby
Sidekiq::Web.app_url = "https://myapp.com"
Sidekiq::Web.use Some::Rack::Middleware
Sidekiq::Web.register Some::Extension
run Sidekiq::Web
```

### After

```ruby
Sidekiq::Web.configure do |config|
  config.app_url = "https://myapp.com"
  config.use Some::Rack::Middleware
  config.register Some::Extension
end

run Sidekiq::Web
```

`Sidekiq::Web.configure` has been backported to 7.3.7 so extensions can use the new pattern in Sidekiq 7 and 8.

## Parameter Paranoia

Previously Sidekiq::Web intermixed route parameters (e.g. "/queues/default") and URL parameters (e.g. "/?foo=bar&count=1&page=3") into one big bag of parameters. A paranoid engineer can see opportunities for namespace clashes and unexpected overrides, possibly leading to security issues. Now these two sources of user input are distinct and separate:

```ruby
app.get "/queues/:name" do
  route_params(:name) # => "default"
  url_params("foo") # => "bar"
  url_params("page") # => "3"
end
```

Route parameters are declared as Symbols and lookup uses Symbols. URL parameters use Strings only. Sidekiq does not use indifferent Hashes internally. The `route_params` and `url_params` methods have been backported to 7.3.7 so Web UI extensions can support both 7.x and 8.x.

## Notes

There are some smaller compatibility breaks:

* Removed a lot of dumb public constants like `X_CASCADE = "x-cascade"`. I doubt many people are referencing our internal constants, public or not.
* Removed `default_tabs` and `custom_tabs`, in favor of just `tabs`.
* Removed the `before` and `after` functionality which I'm the only user that I know of.
* Removed old methods like `settings`, `set` and `enable` on Sidekiq::Web which were Sinatra compatibility stubs.


badosu:
Very nice, much cleaner! :-)

PR #6539, opened 2024-12-06
Make queue latency always float even if there's no job entry

sato-s:
According to the documentation, `#latency` returns Float.

> Returns:
> (Float) — in seconds
> https://www.rubydoc.info/github/mperham/sidekiq/Sidekiq%2FQueue:latency

But, it returns integer 0 when there's no job entry.  
This behavior is problematic when you want to send `#latency` to somewhere that require type definition.
I have faced this problem when I tried to send latency to [Google Cloud Metrics](https://cloud.google.com/monitoring/api/metrics_gcp)

This PR makes `#latency` to send float 0.0 even if there's no job entry.

### DIFF
--- a/test/api_test.rb
+++ b/test/api_test.rb
@@ -51,6 +51,7 @@ describe "API" do
       assert_equal 0, s.failed
       assert_equal 0, s.enqueued
       assert_equal 0, s.default_queue_latency
+      assert_kind_of Float, s.default_queue_latency
       assert_equal 0, s.workers_size
     end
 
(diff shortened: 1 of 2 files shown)


mperham:
Seems fine. 👍🏻

test/api_test.rb
```diff
@@ -51,6 +51,7 @@ class JobWithTags
       assert_equal 0, s.failed
       assert_equal 0, s.enqueued
       assert_equal 0, s.default_queue_latency
+      assert_kind_of Float, s.default_queue_latency
```
sato-s, line 54: `assert_equal` doesn't distinguish float 0.0 and integer 0, so I have added class assertion

PR #6542, opened 2024-12-11
Enhance Language Code Consistency for Internationalization

mailbaoer:
This pull request aims to standardize and enhance the compatibility of language codes across our application. By changing the following language codes to their more widely accepted forms like `rails-i18n` gem:

`pt-br` to `pt-BR`
`zh-cn` to `zh-CN`
`zh-tw` to `zh-TW`
We ensure that our internationalization efforts are more uniform and consistent, which is particularly important when integrating with tools like sidekiq-cron. This change prevents the confusion that can arise from having two seemingly different codes (`zh-cn` and `zh-CN`) both representing Simplified Chinese. By adopting a single, standardized code for each language, we streamline our localization process and improve the user experience for international users.

### DIFF
--- a/lib/sidekiq/web/helpers.rb
+++ b/lib/sidekiq/web/helpers.rb
@@ -154,7 +154,7 @@ module Sidekiq
     # See https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.4
     def user_preferred_languages
       languages = env["HTTP_ACCEPT_LANGUAGE"]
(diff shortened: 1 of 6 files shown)


mailbaoer:
When using both Sidekiq and Sidekiq-Cron, both 'zh-cn' and 'zh-CN' will appear simultaneously.
image

mperham:
This broke the build. Please submit a followup or I will need to revert.

PR #6543, opened 2024-12-11
Change Sidekiq's default thread priority to -1

mperham:
Ruby's default 100ms timeslice can cause significant delays in threading if the user has configured high concurrency with CPU heavy jobs (e.g. the old Sidekiq default of 25 threads). Worst case, each of those 25 threads can wait 2.5 seconds before they get back on the CPU. Now imagine an IO timeout set for 1 second -- it's quite possible to see random TimeoutErrors due to timeslice starvation.

This PR adjusts the default Thread.priority for all Sidekiq threads to be -1, or a 50ms timeslice. -2 would use a 25ms slice.

The user can adjust the priority like so:

```ruby
  # I don't recommend setting any values besides -2, -1 and 0.
  # 0 is Ruby's default, -1 is Sidekiq's default.
  cfg.thread_priority = 0
end
```

See also https://github.com/bensheldon/good_job/issues/1554

### DIFF
--- a/lib/sidekiq/capsule.rb
+++ b/lib/sidekiq/capsule.rb
@@ -27,7 +27,7 @@ module Sidekiq
     attr_reader :mode
     attr_reader :weights
 
-    def_delegators :@config, :[], :[]=, :fetch, :key?, :has_key?, :merge!, :dig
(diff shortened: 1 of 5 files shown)


mperham:
I'm being relatively conservative here. -2 might be a better choice long-term but I want to test the waters first.

PR #6549, opened 2024-12-14
Reimplement UI on Pico CSS

mperham:
This is a prototype of Sidekiq's Web UI using [Pico CSS](https://picocss.com) instead of Bootstrap.

## Goals

1. Reduce as much of our custom CSS and class usage as possible.
2. Reduce as much of our custom JavaScript as possible.

## Notes

- There's lots of bugs, this is still a work in progress.
- Everything is BIG. Too big. Needs to be scaled down.
- I have not tested light mode.
- RTL appears to work ok.
- I've removed a few minor things where its value did not justify its codesize, e.g. the dashboard pulse.

Let me know what you think!

## Gallery

Screenshot 2024-12-14 at 5 27 27 AM
Screenshot 2024-12-14 at 5 27 46 AM
Screenshot 2024-12-14 at 5 27 58 AM
Screenshot 2024-12-14 at 5 28 07 AM
Screenshot 2024-12-14 at 5 28 20 AM

PR #6551, opened 2024-12-16
Reimplement UI with vanilla CSS

antoinem:
Another prototype for the web UI (cf #6549) but with vanilla css.

There is still some tuning and cleaning to do as we can for instance remove a lot of useless CSS classes in the views and have something similar to Pico, but with less (we don't need everything in Pico) and tailor-made for Sidekiq.

What do you think?

Screenshot 2024-12-16 at 09 24 44
Screenshot 2024-12-16 at 09 24 00
Screenshot 2024-12-16 at 09 21 56
Screenshot 2024-12-16 at 09 31 39


antoinem:
Thank you!

Yes, I would be happy to continue improving and supporting it in the future.

We can get to something very close to Pico, classless or few CSS classes, where you can write `<table>` to have all the styles instead of what you had before `<table class="table table-bordered table-striped table-hover">`…
We can split the CSS file to make it more readable: "design_tokens, layout, typography, header, table, rtl, …" or add some comments, but right now, with everything, it's just 12kb, and we probably can simplify a few elements (nav…).
I think it's not too complex, even if you don't know a lot of CSS. You will find contributors that can update it.

Pico is nice, but it covers much more with many parts you don't need (typography, forms, several color schemes, etc.). Also, you'll probably get to a point where you have to dive into it or write some custom CSS to overwrite their choices. If there are some elements of styles you prefer in Pico, we can get some inspiration and borrow them.

mperham:
I'd prefer a minimum of files and no build. A single `style.css` is perfect, you can add comments to separate sections within.

I used Pico's Card component for the stats boxes, that's a nice abstraction versus styled divs.

https://picocss.com/docs/card

mperham:
Background research on `oklch`:

https://evilmartians.com/chronicles/oklch-in-css-why-quit-rgb-hsl
https://github.com/microsoft/vscode/issues/165207

mperham:
Make sure the UI incorporates the dark red Sidekiq brand color. The primary color is way too light right now.

logo

mperham:
^^^^^^ Please don't touch anything in `examples/`. You're missing the greater view: that example is meant to represent 3rd party code and how it can integrate their own CSS.

antoinem:
Screenshot 2024-12-19 at 20 34 47

Yes, I need to recalculate all the schemes from your official brand color. It will work well for the light theme.
The problem is: for the dark theme, even on pure black the official color doesn't meet the accessibility requirements for contrast. That's why I was using a lighter red in dark mode.
Do you prefer to use the official color like that anyway? Or may I adjust it in dark mode to meet the requirements?

mperham:
Definitely lighten the color in dark mode. Readability is more important; we just need to find a good balance.

antoinem:
Screenshot 2024-12-20 at 12 27 53

Screenshot 2024-12-20 at 12 26 25

Better?

I won't be around next week. Happy holidays!

mperham:
Looks great. Whats the status? What’s left?

mperham:
A few quick items. I'm looking at this in dark mode, may or may not apply in light mode.

- Make the brand name bigger in the header and maybe bold it in the footer?
- Warn button is a little too close to brown, a little more yellow/orange?
- The language selector isn't visible in dark mode. The user can't tell it is changeable, maybe the border should be lighter?
- Args are monospaced in Retries but not in Scheduled. Make sure this is consistent everywhere?
- Add a little bit of padding between number and caption in the Busy Status bar, it's just a little tight.
- Buttons on the job details page need padding with the table above.
- The period selector on the metrics_for_job page could be more visible.

But overall, this looks amazing. Really nice work.

mperham:
I'm testing with Sidekiq Enterprise, which found a few more issues:

* When the live poll button is gone (on the Dashboard), the nav has no top padding. It also floats over the content when it should scroll with the rest of the content.
* On Busy in dark mode, the "all" buttons use white text but the buttons in the Processes table use black text.
* The "reliable" and "leader" tags have no color/border.
* Footer also appears to float over content. I'd prefer them to be fixed in the page.
Screenshot 2024-12-20 at 11 22 35 AM

antoinem:
Up.
Not sure to understand this:
> Footer also appears to float over content. I'd prefer them to be fixed in the page.

mperham:
Scroll all the way down and then back up. The footer will cover some content.

mperham:
Screenshot 2024-12-20 at 12 56 33 PM
Screenshot 2024-12-20 at 12 56 51 PM

The retry buttons are a usability question. Should we move the All buttons to the right side? Will people confuse them by clicking multiple checkboxes and clicking "Delete All" (when the correct thing is to click "Delete")? Removing the checkboxes and putting three buttons in every table row seems bad too but that's what we do in Busy.

antoinem:
Screenshot 2024-12-20 at 22 33 34

I have sorted them to be in the same order and same colors (retries all like retry now).
Maybe there should only be one of the two rows of buttons visible at a time?
All buttons when nothing or all checkboxes are checked. The others otherwise.
"Retry now" or "Delete" doesn't work if nothing is selected.

mperham:
Good catch on the sort order.

In principle I like the idea of reducing the button count but it feels a little suspicious. There's several states:

- Check one box. Click Delete to delete 1.
- Check the top check box to select all on page. Click Delete to delete 25.
- Check none. Click Delete All to delete all Retries.

So we'd use JS to swap the buttons if anything is checked? I think as long as it's clear that the buttons have changed, that should work. I would toggle the left or right side buttons visible rather than swapping them in place. What do you think?

antoinem:
I would try swapping them in place. For the user it is the same actions. It's probably weird if the buttons disappear and appear somewhere else after you check or uncheck a checkbox.

mperham:
Do what you think is best and we'll see how it "feels" in use.

antoinem:
Done. I think it's better. You?

mperham:
This looks really good.

* I think the filter box on Retries, Scheduled, etc needs to be a bit more visible, maybe a lighter background color for form controls?
* Checking the top checkbox doesn't switch to Retry Now. It shouldn't say Retry All because it only retries the selected items on this page.

Screenshot 2024-12-31 at 9 51 56 AM

mperham:
I'm thinking the button swap is too tricky for the UI. I'm ok with removing the "* All" functionality. It's not something that should happen normally and people can use a one liner API call if necessary.

```ruby
require 'sidekiq/api'
Sidekiq::RetrySet.new.retry_all
Sidekiq::RetrySet.new.kill_all
```

mperham:
Huge congratulations, this is a major upgrade for Sidekiq and you've been fantastic to work with. Thank you so much for contributing your time and expertise to improve Sidekiq for everyone. 😎

antoinem:
Thank you, and thank you for Sidekiq! Happy to be able to give back. ❤️

mperham:
Hey @antoinem, since you're the current Sidekiq UI CSS expert, do you think the UI could be converted quickly from fixed-width to full-width, like #6686 asks?

antoinem:
Hey @mperham,
I'm not sure to understand the issue.
You mean having all the UI full width without limit? Then you can simply remove `max-width: 1440px;`  in style.css:68
or maybe just raise the value as I'm not sure it won't have downsides on some pages if it's too large (very long lines hard to read, big movements needed to reach an action, …).
If it's in a specific context only, there might be a better solution than increasing all the app max width.

PR #6552, opened 2024-12-16
push bulk with dynamic queues

### LINKED ISSUE #6546: Dealing with perform_bulk and dynamic queues set from a client middleware.
By design, `perform_bulk` will push all the jobs in a single queue since its based on a single worker. In my case, I've a `client_middleware` that will update a job's queue dynamically based on its args. This works perfectly when calling `perform_async` but when using `perform_bulk` the queue is [set](https://github.com/sidekiq/sidekiq/blob/7239276d85089e9a19c2642584bf6fa5a53224dc/lib/sidekiq/client.rb#L272) like this `payload['queue'].first`.

I wonder if it could be possible to partition queues from the payload and push jobs on different queues instead of relying on the first one.

ericproulx:
This PR changes the way `atomic_push` chooses the queue to send payloads. Instead of taking the first one, it regroups payloads by queue and dispatch them. Fix #6546

### DIFF
--- a/test/client_test.rb
+++ b/test/client_test.rb
@@ -58,6 +58,13 @@ class MiddlewareArguments
   end
 end
 
+class MiddlewareDynamicQueue
+  def call(worker_class, job, queue, redis)
+    job["queue"] = job["args"].first.odd? ? "odd_queue" : "even_queue"
+    yield
+  end
+end
+
 class DJob < BaseJob
 end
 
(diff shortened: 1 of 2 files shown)


test/client_test.rb
```diff
@@ -479,6 +486,41 @@ def call(worker_klass, msg, q, r)
...
       assert_match(/[0-9a-f]{12}/, result[1])
     end
+
+    it "push bulk sends worker in correct queue" do
+      aggregator = Hash.new do |hash, key|
+        hash[key] = Hash.new do |inner_hash, inner_key|
+          inner_hash[inner_key] = []
+        end
+      end
+      pool = Sidekiq::RedisConnection.create(size: 1)
```
mperham, line 496: Don’t do this mocking. Test the contents of Redis direct with the API.

mperham (commented):
Nice work.

PR #6559, opened 2024-12-20
Use String#b to create a mutable binary string

byroot:
I just saw: https://bugs.ruby-lang.org/issues/20966

`String#b` is the most convenient method to create a mutable binary string.

I also replaced encoding strings by the actual constants.


mperham:
Huh, never heard of it.  Seems like String#binary would have been a much clearer name.

PR #6564, opened 2024-12-24
Make `created_at`/`enqueued_at` storing integer milliseconds

fatkodima:
Discussion https://github.com/sidekiq/sidekiq/discussions/6562.

With this PR, sidekiq will now store created_at/enqueued_at timestamps as integer milliseconds since the epoch. 
It will also currently accept timestamps in the old format (fractional seconds since the epoch) for existing jobs.


mperham:
How do the benchmarks look?

fatkodima:
Time benefit is the same as if using integers, but memory benefit is 2x worse than if using integers (for integers we will save 180mb in redis, but for milliseconds now 90mb).

mperham:
I switched the implementation to use:

```ruby
::Process.clock_gettime(::Process::CLOCK_REALTIME, :millisecond)
```

as it benchmarked as over 2x faster than `(Time.now.to_f * 1000).floor`.

PR #6587, opened 2025-01-24
Add ARIA labels and other accessibility fixes

KNejad:
I work at GitLab and we have been looking into some accessibility improvements. As we use the Sidekiq Web UI, we had a look at the Sidekiq repo as well, and found a few improvements that we can make.

In this PR I've added a few ARIA labels and a missing table header.

### DIFF
--- a/web/assets/stylesheets/style.css
+++ b/web/assets/stylesheets/style.css
@@ -35,20 +35,6 @@
   outline: 1px solid oklch(from var(--color-primary) l c h / 50%);
 }
 
-.sr-only {
-  border: 0 !important;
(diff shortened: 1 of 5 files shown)


mperham:
Yay! I'm always happy to get accessibility improvements since that's not an area I know much about.

mperham:
Do you want to backport this to 7.x or are you ok with it targeting Sidekiq 8.0+?

KNejad:
Thanks @mperham! Certainly, we actually use Sidekiq 7 so I'll backport it on Monday :slightly_smiling_face:

mperham:
Sounds good, it's the `7-x` branch.

PR #6589, opened 2025-01-25
Fix iterable jobs cancellation

fatkodima:
Currently, cancelling an iterable job works only when the job was interrupted and it starts again. In this case the cancellation flag is checked correctly.

When the job is running, the flag's value is checked incorrectly, because it is set to a timestamp in the codebase.
https://github.com/sidekiq/sidekiq/blob/6a360285c146628d6332bf4a68d85c419330c27d/lib/sidekiq/client.rb#L68

I haven't added tests, because looks like it will be particularly hard to test this case (the test case should at least use a new interrupting thread).


mperham:
Can you cherry pick that onto 7-x too? Don't worry about a PR.

fatkodima:
Cherry-picked into 7-x.

PR #6598, opened 2025-02-03
Accessibility polling updates

KNejad:
This adds another accessibility improvement, following on from my previous PR.

When there is an update in the dashboard, if the user is using a screen reader, they should be informed about the update.

This PR makes it so the screen reader will say "Latest poll: Processed: 16. Failed: 0." each update if there is a change (the polling frequency is the same as for the graph, and can be updated the same way).

This works thanks to the `role="status"` attribute. The text will only be read when the user is idle. It will not interrupt other activity.

I added back the `sr-only` CSS class that I removed in the previous PR, since it is now needed for the screen reader update span.

As this was a more involved change than the previous change I made, I asked for a review from a Senior Frontend co-worker, before submitting here, to confirm the value of these changes and check the code quality: https://github.com/KNejad/sidekiq/pull/1

Once this is merged, I'll backport both the previous changes I made as well as this to the `7.x` branch.

### DIFF
--- a/web/assets/javascripts/dashboard-charts.js
+++ b/web/assets/javascripts/dashboard-charts.js
@@ -83,6 +83,8 @@ class RealtimeChart extends DashboardChart {
     this.chart.data.datasets[1].data.push(failed);
     this.chart.update();
 
+    updateScreenReaderDashboardValues(processed, failed);
+
     updateStatsSummary(this.stats.sidekiq);
     updateRedisStats(this.stats.redis);
     updateFooterUTCTime(this.stats.server_utc_time);
--- a/web/assets/javascripts/dashboard.js
+++ b/web/assets/javascripts/dashboard.js
@@ -36,6 +36,12 @@ var ready = (callback) => {
   else document.addEventListener("DOMContentLoaded", callback);
 }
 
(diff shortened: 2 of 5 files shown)


mperham:
I'm not convinced this is useful and worth the effort. Is there a developer who actually wants this? Why not refresh the page when you want an update? Polling updates are easy when you are updating the screen but speaking the changes every N seconds seems overly assertive. The developer could be polling for other reasons.

KNejad:
Thanks for taking a look @mperham! To be honest when I started working on this, I was also concerned it would be quite intrusive to users. However, thinking about it more, a big reason for the dashboard page is so that users can quickly see what's going on and whether there is an increase in failed or processed jobs. As sighted users we can easily check the graph to see whenever we need to, however, a visually impaired person currently doesn't have a way to know the latest update about job statuses. This allows them to use the dashboard in the same way sighted people can.

I discussed this with a couple co-workers who have more experience with accessibility, and they agreed this would be an appropriate solution.

Just for clarity: This PR does not cause any additional polling. It just hooks into the existing polls. It also will have no affect on sighted users. Only people with a screen reader will hear the audible announcement about changes.

PR #6606, opened 2025-02-11
Expose current cursor inside iterable jobs

fatkodima:
I need to have access to the current cursor inside the iterable job, to store it externally (basically a db table row with some metadata about the iterable job). Currently, can access it only via an instance variable.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -19,6 +19,7 @@ HEAD / main
 - Freshen up `Sidekiq::Web` to simplify the code and improve security [#6532]
   The CSS has been rewritten from scratch to remove the Bootstrap framework.
 - Add `on_cancel` callback for iterable jobs [#6607]
+- Add `cursor` reader to get the current cursor inside iterable jobs [#6606]
 - Default error logging has been modified to use Ruby's `Exception#detailed_message` and `#full_message` APIs.
 - CI now runs against Redis, Dragonfly and Valkey.
 - Job tags now allow custom CSS display [#6595]
--- a/lib/sidekiq/job/iterable.rb
+++ b/lib/sidekiq/job/iterable.rb
@@ -64,6 +64,10 @@ module Sidekiq
         @_cancelled
       end
 
+      def cursor
+        @_cursor.freeze
+      end
+
       # A hook to override that will be called when the job starts iterating.
       #
       # It is called only once, for the first time.


mperham:
I’m concerned about encapsulation. We don’t expose the job payload to jobs for the same reason. Can we expose it in a safer manner, e.g. a callback?

fatkodima:
1.
>  Can we expose it in a safer manner, e.g. a callback?

Can you provide an example?

2. While we are here, I also need a callback for when the job is cancelled (similar to `on_start`/`on_complete` etc). Ok if I will add?

mperham:
Something like:

```ruby
on_step(args, cursor:)
on_cancel(args)
```

fatkodima:
> Can we expose it in a safer manner, e.g. a callback?

These will be in the 99% of cases simple integers. By "safer" do you mean for people to not be able to modify it? If yes, will freezing it help?

mperham:
Yeah, I’m worried about people mutating the cursor. Dup’ing it would also work but freeze is probably better to avoid an unwanted allocation.

fatkodima:
Added freezing. A simple reader looks simpler to me than the callback. And with a custom callback I will still need to save in another instance variable the current cursor's value.

I will implement that `on_cancel` cursor in a separate PR.

mperham:
Can you show me a more complete use case so I can judge if this is the right choice? For instance, this doesn’t feel right for unit testing the job code. I’d prefer a callback for that.

fatkodima:
A sketch of the job
```ruby
class MigrationsJob
  include Sidekiq::IterableJob

  def initialize
    super

    @ticker = Ticker.new(1) do |ticks, duration|
      @migration.persist_progress(@_cursor, ticks, duration)  # <--- here @_cursor is used
      @migration.reload
    end
  end

  def on_start
    @migration.start
  end

  def on_stop
    @ticker.persist
    @migration.stop
  end

  def build_enumerator(migration_id, cursor:)
    @migration = BackgroundMigrations::Migration.find(migration_id)
    # ... more code
  end

  def each_iteration(item, *args)
    # ... more code
    @ticker.tick
  end
end
```

Users want to run background data migrations. The row in the relevant db table is created to store the metadata about what data to migrate, tracking the progress etc. The sidekiq job can disappear, be stopped etc, but I need to track the progress of the data migration, also display a dashboard of existing background migrations. So I store the current cursor once in a while inside the job.

mperham:
I don't think that's a good design. The code is accessing private state from the job with zero thread safety (I assume ticker is a separate thread?). And direct access to the cursor within the job makes unit testing more complex because someone has to provide a valid cursor to the job code. If you're performing database work, why not rely on ACID semantics and SQL queries to get your current status?

fatkodima:
Ticker is a separate class, no threading. Just an example of usage of the current cursor.

> If you're performing database work, why not rely on ACID semantics and SQL queries to get your current status?

Please, explain more. Assume the case where the user wants to backfill a 1 billion rows table. The current design is that the user provides an ActiveRecord::Relation to iterate over and a method to call with each subrelation of records. We need to be able to pause the job, cancel, get the % of progress, call some callbacks on completion/cancellation/etc.

mperham:
I'll accept this as is. Would you update Changes.md?

fatkodima:
Updated changelog.

Issue #6610, opened 2025-02-13
Expand metrics to 24 hours (up from 8 hours)


MarkSiemers-msr:
The metrics available both in the UI and programmatic API are great.

Would you consider expanding the time-frame to allow for 24 hours of metrics. Perhaps have it configurable if some customers don't want to use the space for that many metrics.

I understand it is a slippery slope and this isn't meant to replace something like DataDog. Increasing to 24 hours allows viewing of daily trends whereas an 8 hour window doesn't give a complete picture.

Having the options for viewing set at:
```
1h
2h
4h
8h
16h
24h
```

Of course, if you wanted to stick with powers of 2, then a 32 hour max would also work.

mperham:
I don't have a strong opinion on the time limit. My main concern is that the UI scales to the data density.

mperham:
The issue is that the longer we store, the more data points we'd need to store and query. 8 hrs requires (8*60) or 480 data points. Generally for longer terms we want to rollup or aggregate the data into larger buckets. Right now we only store minute buckets but we already have the start of "MID_TERM" rollups in `sidekiq/metrics/tracking.rb`. I'd adjust this to make the buckets every ten minutes (the bucket key would just drop the last digit from the minutes). So storing 24 hours of data would store 144 data points instead of 1440 points. 144 points should make for a reasonable data density and query size.

In summary, we don't aggregate the data today. We'd have to add bucket aggregation and adjust the query implementation to use the aggregate buckets.

mperham:
Done, now available on main.

mperham:
This will be part of Sidekiq 8.0 coming next month.

MarkSiemers-msr:
Thank you @mperham - I just used this change the other day, and wanted to provide an example here where it is very useful.

Changes were made to one of our workers recently (SQL optimizations, and a few other things).

Viewing the history over 8 hours, you can't really tell if there was a meaningful effect:

<img width="1430" height="509" alt="Image" src="https://github.com/user-attachments/assets/5383a8b5-04a9-4edb-b838-934557bb81e6" />

But over 72 hours, it is clear:

<img width="1444" height="560" alt="Image" src="https://github.com/user-attachments/assets/36354b33-9d04-46d3-9298-c953e083972e" />

For anyone extra curious on why there are still spikes in the 8 hour period. We occasionally perform the task not just on "current" records, but also historical. When that happens an increase in processing time is expected.

[...]

PR #6614, opened 2025-02-17
Multi-day metrics storage

mperham:
In #6610 we have a request to store metrics longer term than 8 hours. Right now showing an eight hour graph requires (8*60) = 480 data points.

If we take 480 as our upper limit and think about aggregate buckets, today a minute bucket for 2025-02-17 12:34 is named `20250217|12:34` in Redis and gives us 1440 buckets per day. We could easily create 10 minute buckets like `20250217|12:3` which would give 144 buckets per day, or 432 buckets for three days.

Storing 480 hourly buckets `20250217|12` would allow us to show 20 days worth of data but I'm not willing to accept that yet.

I'm proposing:

- We create a new set of rollup buckets which aggregate over 10 minutes.
- The TTL for this data should be 72 hours.
- Add two new options to /metrics: 24 hours and 72 hours.
- This will NOT affect the histogram/bubble page, which should remain as is.
- Update the API to add new `hours` option for metrics query.
- (unrelated) chopping off the century ("20") in the bucket name, it doesn't seem necessary. Since we only store 8 hours of data today, I'm not concerned with backwards compatibility.


MarkSiemers-msr:
@mperham - This looks good, I appreciate you considering the suggestion and implementing a proposed change so quickly.

Feedback on a couple proposal items:
> Add two new options to /metrics: 24 hours and 72 hours.

Would it be easy to also support 48 hours, meaning three new options: 24, 48, 72 ?

> This will NOT affect the histogram/bubble page, which should remain as is.
This makes sense, though it might make the transition from the line-graph page to the bubble page tricky, as the link may provide an unsupported parameter.
For example
From the page: https://example.com/admin/sidekiq/metrics?substr=&period=24h
A link could be: https://example.com/admin/sidekiq/metrics/ReallyImportantWorker?period=24h

It looks like the current behavior when an unsupported param is passed is to default to 1h. I think that would be acceptable, but perhaps consider - if the param is larger than the max supported value, default to the max supported value (8 hours).

Thanks again for all your work. We rely heavily on Sidekiq and Sidekiq Pro on our team.

lib/sidekiq/metrics/query.rb
```diff
@@ -21,17 +21,29 @@ def initialize(pool: nil, now: Time.now)
...
         result = Result.new
-
+        minutes = 60 unless minutes || hours
+        count = hours ? hours * 6 : minutes
         time = @time
+        minutely = ->(time) { "j|#{time.strftime("%y%m%d")}|#{time.hour}:#{time.min}" }
+        tenly = ->(time) do
+          m = time.min
+          mins = (m < 10) ? "0" : m.to_s[0]
+          "j|#{time.strftime("%y%m%d")}|#{time.hour}:#{mins}"
```
MarkSiemers-msr, line 35: It looks like the `minutely` and `tenly` procs are repeated here and in `for_job`

Would it make sense to move them to methods?
```ruby
def minutely_key(time)
  "j|#{time.strftime("%y%m%d")}|#{time.hour}:#{time.min}"
end

def ten_minutely_key(time)
  m = time.min
  mins = (m < 10) ? "0" : m.to_s[0]
  "j|#{time.strftime("%y%m%d")}|#{time.hour}:#{mins}"
end
```
And then use:
```ruby
key_method = minutes ? :minutely_key : :ten_minutely_key
# ...
count.times do |idx|
  key = send(key_method, time)
  # ...
end
```

MarkSiemers-msr: If you don't mind indexing to the 14th character of a string, the tenly could be:
```ruby
def ten_minutely_key(time)
  "j|#{time.strftime("%y%m%d")}|#{time.hour}:#{time.to_s[14]}"
end
```

lib/sidekiq/metrics/query.rb
```diff
@@ -21,17 +21,29 @@ def initialize(pool: nil, now: Time.now)
...
         redis_results = @pool.with do |conn|
           conn.pipelined do |pipe|
-            minutes.times do |idx|
-              key = "j|#{time.strftime("%Y%m%d")}|#{time.hour}:#{time.min}"
+            count.times do |idx|
+              key = keyproc.call(time)
               pipe.hgetall key
               result.prepend_bucket time
-              time -= 60
+              time -= advance
```
MarkSiemers-msr, line 46: It's just a local variable, but `advance` is a little counter-intuitive here. Would you be opposed to `decrement` or `decr`

mperham:
I'm not really looking for PR feedback just yet. It's a WIP as the comment message notes.

mperham:
Backend is done. Might need help with the UI, we'll see tomorrow.

mperham:
I think I got most of the UI working. The JS labels/tooltips are a little rough but I'm not familar enough with JS and @adamlogic's code to fix those.

Screenshot 2025-02-20 at 9 05 23 AM

lib/sidekiq/metrics/query.rb
```diff
@@ -72,49 +102,62 @@ def for_job(klass, minutes: 60)
...
         def prepend_bucket(time)
-          buckets.unshift time.strftime("%H:%M")
+          buckets.unshift bkt_time_s(time)
           self.ends_at ||= time
           self.starts_at = time
         end
+
+        def bkt_time_s(time)
+          # hourly buckets should be rounded to ten ("8:40", not "8:43")
+          # and include day
```
adamlogic, line 131: This string is used for both the x-axis values and the tooltips. I agree that "22 11:30" is weird and barely readable in the UI.

CleanShot 2025-02-22 at 11 29 33@2x


I'd thought maybe we ditch the data altogether and only show the time—it's three days of data at most, pretty easy to figure out which day you're on—but we can't do that. These strings are used as the keys for the datapoints in the chart, and we need to disambiguate "11:30" today versus "11:30" yesterday.

CleanShot 2025-02-22 at 11 28 12@2x

So we need buckets that include the date and are human readable. It also needs to be universally recognizable ("2/7 means different things depending where you live). I feel like that leaves a few options:

1. `Feb 22 11:30`
2. `22 Feb 11:30`
3. `2025-02-07 11:30`

All of them make for a pretty noisy x-axis, but I can't think of a way around it. I personally think option 2 the cleanest, but the tooltip is still weird.

CleanShot 2025-02-22 at 11 53 18@2x

We can clean that up by getting rid of the range in the tooltip.

CleanShot 2025-02-22 at 12 05 16@2x

See #6620 for this implementation

adamlogic:
Regarding the issue with the last bucket in the chart, it's caused by the "marks" not being rounded to the same time buckets as the other data points in the chart.

Before:

CleanShot 2025-02-22 at 12 45 22@2x

After:

CleanShot 2025-02-22 at 12 48 32@2x

This fix is also in #6620.

adamlogic:
As a side note, part of what makes these charts such a pain is because we're manually managing the buckets and labels. Typically with time-series charts you'll leave that to the chart itself—the data includes a timestamp, and you tell the chart how to lay them out across a time axis.

Chart.js has [good support for time axes](https://www.chartjs.org/docs/latest/axes/cartesian/time.html), but it requires pulling in [one of these date libraries](https://github.com/chartjs/awesome#adapters), and I know we're really trying to avoid dependencies here.

mperham:
If you want to refactor to make things easier to write and maintain, let's give it a shot. Are these date libraries easy to vendor?

mperham:
@adamlogic Should we see grid marks here? 

Screenshot 2025-02-24 at 8 28 26 AM

I noticed the official screenshot linked below shows an obvious grid.

https://github.com/sidekiq/sidekiq/wiki/Metrics#metrics-home

mperham:
Do the metrics graphs auto-refresh every minute? If so, we should add a footnote that the graphs are live and don't need a page refresh.

mperham:
Do you see any need to upgrade to Chart.js 4.x? https://www.chartjs.org/docs/latest/migration/v4-migration.html

adamlogic:
> Should we see grid marks here?

Yes. They do show up in light mode, but not in dark mode. It looks like that broke in #6551. I can see the gridlines as expected if I revert this diff:

CleanShot 2025-02-25 at 16 30 30@2x

According to the chart.js docs, `oklch` format is [not listed as supported](https://www.chartjs.org/docs/latest/general/colors.html#color-formats). That format was added to several other chart configs, and I'm assuming they should all be reverted, but I'm not entirely sure... the gridlines are the only thing that's obviously not working as expected.

Either way, that fix probably belongs in a separate PR, and maybe @antoinem can chime in since I'm not super familiar with `oklch`.



> Do the metrics graphs auto-refresh every minute?

Not as far as I can tell.

> Do you see any need to upgrade to Chart.js 4.x?

Other than taming the [libyear](https://libyear.com), I don't see any need.

PR #6620, opened 2025-02-22
Fixes & improvements to multi-day metrics

adamlogic:


mperham:
Can the internal bucket name be different from the axis tick label? Does Chart.js need an upgrade? Maybe a later version allows that? The internal name could be an ISO8601 timestamp which the Result classes could provide.

mperham:
We allow one deploy per minute today. With rounding, any deploys within that 10 minute bucket will rollup, is that right? I'm mostly ok with that, I don't think it'll matter much.

adamlogic:
> With rounding, any deploys within that 10 minute bucket will rollup, is that right? I'm mostly ok with that, I don't think it'll matter much.

Correct.

> The internal name could be an ISO8601 timestamp which the Result classes could provide.

Yeah, this would be ideal. As I mentioned in #6614, the challenge right now is that we're also providing the "labels" (you can think of them as the x-axis "ticks") for the chart, as well as the datapoints, and the labels of each must match. The chart itself has no concept of time—it's just key-value pairs matched up with the labels we provide.

I'll take a swing at refactoring the charts to have a time-scale x-axis. Ideally we wouldn't need to provide the labels, and we could provide the datapoints with a timestamp, telling the chart how we want those timestamps formatted in the x-axis and tooltips.

test/metrics_test.rb
```diff
@@ -130,13 +128,9 @@ def create_known_metrics(time = fixed_time)
 
       q = Sidekiq::Metrics::Query.new(now: fixed_time)
       result = q.top_jobs(class_filter: /some/i)
-      assert_equal fixed_time - 59 * 60, result.starts_at
+      assert_equal fixed_time - 60 * 60, result.starts_at
```
adamlogic, line 131: To be honest I'm not sure what caused this to change, but it's the difference of one minute, so... 🤷‍♂️

lib/sidekiq/web/application.rb
```diff
@@ -77,9 +77,11 @@ def initialize(inst)
       get "/metrics/:name" do
         @name = route_params(:name)
         @period = h(url_params("period") || "1h")
-        q = Sidekiq::Metrics::Query.new
-        @periods = METRICS_PERIODS
+        # Periods larger than 8 hours are not supported for histogram chart
+        @period = "8h" if @period.to_i > 8
+        @periods = METRICS_PERIODS.reject { |k, v| k.to_i > 8 }
```
adamlogic, line 82: Here's this constraint in action. Without it, you'd encounter an error when viewing 24h metrics and linking to the job metrics.


https://github.com/user-attachments/assets/c5b074b5-8ac4-49a5-812a-48893c98c44d

web/assets/javascripts/chartjs-adapter-date-fns.min.js
adamlogic, line 1: Adds 50kB. Seems worth it. (For reference, chart.min.js is 200kB)

web/assets/javascripts/metrics.js
```diff
@@ -206,7 +188,7 @@ class HistBubbleChart extends BaseChart {
     });
 
     // Chart.js will not calculate the bubble size. We have to do that.
-    const maxRadius = this.el.offsetWidth / this.options.labels.length;
+    const maxRadius = this.el.offsetWidth / 100;
```
adamlogic, line 191: This is an arbitrary number that results is reasonable looking bubbles. The previous implementation was equally arbitrary, but it relied on `labels`, which we're no longer passing.

adamlogic (commented):
I added some changes that update the metrics charts to be true time-series charts. The server passes down data with ISO8601 timestamps, along with start/end times, and the chart figures out how to format the ticks (labels). A side benefit is that the times are now localized.

CleanShot 2025-02-23 at 09 42 00@2x

CleanShot 2025-02-23 at 09 41 52@2x


CleanShot 2025-02-23 at 09 51 58@2x

CleanShot 2025-02-23 at 09 51 50@2x

Note that the dashboard charts are still functional, but they have NOT been refactored to be time-series charts like the metrics charts.

@mperham LMK what you think!

mperham:
I’ll review the code tomorrow but this looks great with your usual excellent communication. 😎

PR #6621, opened 2025-02-24
Add ability to specify current attributes to pass into the Sidekiq thread

pertsenga:


mperham:
This is a little too Rails-specific and a little too specialized to go into Sidekiq core. I'd prefer you publish this functionality as a 3rd party gem. You are welcome to send refactorings which make this easier for you.

pertsenga:
> This is a little too Rails-specific and a little too specialized to go into Sidekiq core.

I agree this but having the `CurrentAttribute` middleware in the first place is already rails-specific anyway, I just refined the existing feature.

Another problem with loading the whole `CurrentAttribute` singleton, aside from the problem I encountered, is that there might be sensitive attributes in it that should not be logged into the **Redis** store as job attributes. The current feature loads everything from the singleton into the job attributes

If the module placement is too conspicuous for core, I can move it maybe next to Its middleware so it's bundled close together or If you have any suggestions

mperham:
You could also use refinements within the Sidekiq module to patch the `Current#attributes` method to use persisted_attributes instead. I've never used refinements before but I think that could work.

pertsenga:
> In retrospect we should have a `sidekiq-rails` gem which houses tools and APIs like CurrentAttributes support. It's likely too late for Sidekiq 8.0 but this is a possible direction for Sidekiq 9.0. That project could directly depend on things like ActiveSupport, provide a better home for the ActiveJob layer, etc. Sidekiq core doesn't (can't!) use any Rails-isms like Concerns.
> 
> A workaround for now would be to create a client middleware which comes after CurrentAttributes in the chain; it can prune the persisted attributes before the job goes to Redis. You can configure that middleware like:
> 
> chain.use AttributePruner, only: [:key1, :key2]
> 
> which looks like:
> 
> ```ruby
> class AttributePruner
>   include Sidekiq::ClientMiddleware
>   def initialize(opts)
>     @only = Array(opts[:only]).map(&:to_s)
>   end
>   def call(klass, queue, job, pool)
>     ctx = job["cattr"].select { |k,v| @only.include?(k) }
>     job["cattr"] = ctx
>     yield
>   end
> end
> ```

Ah yeah, did the same thing for the workaround. Just attempted to get the ability from upstream.

Thanks for the great work with Sidekiq!

PR #6625, opened 2025-02-27
Update scrub method to properly handle password as a Proc

emmanuel-chime:
Fixes: https://github.com/sidekiq/sidekiq/issues/6624.
This change is needed to allow passing a proc in as password, which is the best way to do IAM auth with ElastiCache.

It would be great if this could be included as part of 7.3.x (preferred) or 7.4.x


mperham:
Ok, I think 7-x is in good shape. I've adjusted things to keep backwards compatibility. Like redis-client, Sidekiq will silently Procify any String passwords to prevent log disclosure.

emmanuel-chime:
Thanks @mperham, I think the `redis-client` floor should also be bumped once the new version of redis-client is released allowing `sentinel_password` to be provided as a Proc.

PR #6650, opened 2025-03-18
Remove null values from metrics charts to prevent chart shifting

kyubey1228:
### Issue Summary
Real-time Sidekiq metrics charts currently experience shifting issues over time due to null or empty data points being passed from the server-side to the frontend. This causes the starting point of line charts to shift leftward, negatively impacting readability and user experience.

**Update:** After receiving feedback, I've updated the approach to address this issue at the API level, not the view level. The metrics query API now explicitly returns `0` instead of `nil`.

### Proposed Changes
Updated the metrics query API (`lib/sidekiq/metrics/query.rb`) to replace `nil` values with explicit zeros (`0`) when no data is available.

### Implementation Details

```ruby
# lib/sidekiq/metrics/query.rb
redis_results.each do |(ms, p, f)|
  result.job_results[klass].add_metric "ms", time, ms.to_i
  result.job_results[klass].add_metric "p", time, p.to_i
  result.job_results[klass].add_metric "f", time, f.to_i
  # nil values are now explicitly converted to 0 by using `to_i`
end
```

### Expected Results
- Prevent chart shifting issues over time.
- Provide consistent and accurate real-time chart visualization.

### Testing
Manually verified on local environment that charts no longer shift upon prolonged monitoring.

### Additional Notes
This change addresses the root cause, ensuring API data integrity and simplifying front-end logic.

### DIFF
--- a/lib/sidekiq/metrics/query.rb
+++ b/lib/sidekiq/metrics/query.rb
@@ -101,9 +101,9 @@ module Sidekiq
         time = @time
         @pool.with do |conn|
           redis_results.each do |(ms, p, f)|
-            result.job_results[klass].add_metric "ms", time, ms.to_i if ms
-            result.job_results[klass].add_metric "p", time, p.to_i if p
-            result.job_results[klass].add_metric "f", time, f.to_i if f
+            result.job_results[klass].add_metric "ms", time, ms.to_i
+            result.job_results[klass].add_metric "p", time, p.to_i
+            result.job_results[klass].add_metric "f", time, f.to_i
             result.job_results[klass].add_hist time, Histogram.new(klass).fetch(conn, time).reverse if minutes
             time -= stride
           end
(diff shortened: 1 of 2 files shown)


mperham:
This is a data structure returned by the metrics query API. I'd think that nil values should be 0 instead. Should the query API be returning nil values?

kyubey1228:
> This is a data structure returned by the metrics query API. I'd think that nil values should be 0 instead. Should the query API be returning nil values?

Yes, that's a good point. Returning nil values from the API is probably not the best approach.
Instead, it would be better for the Metrics Query API itself to return zeros explicitly when no data is available.

I'll update the PR accordingly to handle this issue at the API level rather than the view level.
Thanks for pointing this out!

kyubey1228:
Thanks for the feedback! I agreed with your suggestion and changed the approach:

- The API now explicitly returns `0` instead of `nil`.
- Reverted previous changes on the view side.

This should resolve the root cause of the chart shifting issue.
Please review the updated changes when you have time.

kyubey1228:
I was unprepared. I will resume after resolving the test code.

PR #6674, opened 2025-04-11
Add the ability to set directory for profiles and fix profiler options

epoberezhny:
Fixes https://github.com/sidekiq/sidekiq/issues/6671. Also fixes passing a different mode to the Vernier profiler

### DIFF
--- a/lib/sidekiq/profiler.rb
+++ b/lib/sidekiq/profiler.rb
@@ -13,6 +13,7 @@ module Sidekiq
     include Sidekiq::Component
     def initialize(config)
       @config = config
+      @profiles_dir = ENV.fetch('PROFILES_DIR') { Dir.tmpdir }
     end
 
     def call(job, &block)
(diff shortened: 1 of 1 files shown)


lib/sidekiq/profiler.rb
```diff
@@ -13,6 +13,7 @@ class Profiler
     include Sidekiq::Component
     def initialize(config)
       @config = config
+      @profiles_dir = ENV.fetch('PROFILES_DIR') { Dir.tmpdir }
```
mperham, line 16: Everything else looks solid but I'm not sure about the ENV name. Generally I think it's a good idea to prefix the variable with the intended recipient, e.g. `VERNIER_OUTPUT_DIR`. That would also allow Vernier to adopt the variable internally too. See https://github.com/jhawthorn/vernier/issues/146

epoberezhny: No problem. Updated

PR #6680, opened 2025-04-22
[Web]: Allow loading of fonts from data URLs

lloydwatkin:
Update CSP (content security policy) so it allows fonts to load from `data:` style URLs. This was tested on sidekiq 8.0.2.

Before:

Screenshot 2025-04-22 at 16 54 02

After:
Screenshot 2025-04-22 at 16 54 31

### DIFF
--- a/lib/sidekiq/web/application.rb
+++ b/lib/sidekiq/web/application.rb
@@ -15,7 +15,7 @@ module Sidekiq
         "default-src 'self' https: http:",
         "child-src 'self'",
         "connect-src 'self' https: http: wss: ws:",
-        "font-src 'none'",
+        "font-src 'self' data:",
         "frame-src 'self'",
         "img-src 'self' https: http: data:",
         "manifest-src 'self'",


mperham:
Sidekiq doesn't use any custom fonts that I'm aware of. Is a browser plugin injecting them in the current page? Or a 3rd party plugin, i.e. the Cron tab?

lloydwatkin:
It could be `sidekiq-cron`. Is there a mechanism to override CSp headers in the code? If so I can make a contribution to that project instead.

mperham:
We specifically have this code to prevent vectors for content injection. That error is a GOOD thing, something is mucking with page content.

Can you show the page html which is loading the fonts? Maybe that'll give you a clue.

lloydwatkin:
I'm getting these errors from all sidekiq pages. The above example was running locally on the retry page. I'll take a look through our plugins tomorrow and report back

lloydwatkin:
Hmm, I'm not finding the call anywhere obvious. I'll close this until I've found the culprate. Thanks!

PR #6685, opened 2025-04-25
Don't swallow NoMethodErrors in jobs setting current_attributes

tomchipchase:
~Instead of waiting for an error and then filtering out any current attributes, only attempt to set the ones that currently exist.~

If we have successfully set the current attributes, we don't retry if we catch an error.

The `NoMethodError` that was being rescued from also applied to the job that was being wrapped. This caused errors to be silently hidden and the job retried without going into the retry queue, or alerting any bug tracking tools that have been set up.

This recently bit us in a worker calling a third party API because of a bug in our code after the response was received. A NoMethodError was raised, and we retried the API call without being alerted to the failure. The API responded differently the second time (because we'd already made that request), and the second time the code ran we hit the failure branch, which also didn't raise an error as our code assumed it was a legitimate failure.

Ofcourse, ideally our job would have been idempotent, so the retry wouldn't have mattered. But if we were writing perfect code we wouldn't have hit the NoMethodError in the first place.

I saw some discussion on the original PR for this where the rescue was preferred https://github.com/sidekiq/sidekiq/pull/6343 but NoMethodError seems too generic for a blanket rescue. This approach seems cleaner to me, but if the performance of the select is a concern we could try inspecting the error to ensure it was a result of CurrentAttributes instead?

### DIFF
--- a/lib/sidekiq/middleware/current_attributes.rb
+++ b/lib/sidekiq/middleware/current_attributes.rb
@@ -71,11 +71,14 @@ module Sidekiq
         retried = false
 
         begin
+          set_succeeded = false
           klass.set(attrs) do
+            set_succeeded = true
             wrap(klass_attrs, &block)
           end
-        rescue NoMethodError
-          raise if retried
+        rescue NoMethodError => e
+          # Don't retry if the no method error didn't come from current attributes
+          raise if retried || set_succeeded
 
           # It is possible that the `CurrentAttributes` definition
           # was changed before the job started processing.
--- a/test/current_attributes_test.rb
+++ b/test/current_attributes_test.rb
@@ -124,6 +124,23 @@ describe "Current attributes" do
     end
   end
 
+  it "doesn't swallow errors raised in the job" do
+    cm = Sidekiq::CurrentAttributes::Load.new({
+      "cattr" => "Myapp::Current"
+    })
+
+    job = {"cattr" => {"user_id" => 123}}
+    assert_raises do
+      first_time = true
+      cm.call(nil, job, nil) do
+        if first_time
+          first_time = false
+          raise nil.this_method_is_undefined
+        end
+      end
+    end
+  end
+
   private
 
   def with_context(strklass, attr, value)


mperham:
I don’t really like the overhead. Can we yield outside of the rescue?

mperham:
Can you benchmark and show that the overhead is minimal?

tomchipchase:
> Can we yield outside of the rescue?

I'm not sure this is possible, certainly not with the `set(...)` syntax, it requires the block is passed in, but there might be another way to acheive the same thing with a different method.

The other thing that _would_ work would be to do a check that `e.cause&.receiver.class < ActiveSupport::CurrentAttributes`, which would be the same as the current implementation with that as an extra guard clause. 

> Can you benchmark and show that the overhead is minimal?

Yea, I can give this a go.

tomchipchase:
I've tested with a range of current attributes (we have a total of 11 in our app across a few different classes). I was expecting it to be less significant with fewer, attributes, but that doesn't seem to be the case:

```
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "activesupport"
  gem "benchmark-ips"
end

require "active_support"
require "benchmark/ips"

[1, 3, 5, 8, 13, 21, 34, 55, 89].each { |total_attributes|
  current_attributes = Class.new(ActiveSupport::CurrentAttributes) do
    total_attributes.times {
      attribute :"foo#{it}"
    }
  end

  # Current attributes need a constant name assigned
  current_attributes_class = Object.const_set("Attrs#{total_attributes}", current_attributes)

  # Creates a hash of { "foo1" => "bar1", ... } for each attribute
  attributes = total_attributes.times.map { ["foo#{it}", "bar#{it}"] }.to_h

  Benchmark.ips do |bm|
    bm.report("#{total_attributes} no select") do
      current_attributes_class.set(attributes) do
        true
      end
    end

    bm.report("#{total_attributes} select") do
      filtered = attributes.select { |attr| current_attributes_class.respond_to?(attr) }
      current_attributes_class.set(filtered) do
        true
      end
    end
  end
}
```

```
         1 no select    62.339k i/100ms
            1 select    51.746k i/100ms
         1 no select    625.165k (± 2.3%) i/s    (1.60 μs/i) -      3.179M in   5.088135s
            1 select    523.683k (± 1.5%) i/s    (1.91 μs/i) -      2.639M in   5.040545s
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [x86_64-linux]
         3 no select    31.773k i/100ms
            3 select    26.440k i/100ms
         3 no select    315.194k (± 1.6%) i/s    (3.17 μs/i) -      1.589M in   5.041531s
            3 select    271.398k (± 1.2%) i/s    (3.68 μs/i) -      1.375M in   5.066662s
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [x86_64-linux]
         5 no select    21.079k i/100ms
            5 select    18.468k i/100ms
         5 no select    210.855k (± 1.3%) i/s    (4.74 μs/i) -      1.075M in   5.099248s
            5 select    179.437k (± 2.4%) i/s    (5.57 μs/i) -    904.932k in   5.046174s
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [x86_64-linux]
         8 no select    13.579k i/100ms
            8 select    11.937k i/100ms
         8 no select    136.178k (± 1.5%) i/s    (7.34 μs/i) -    692.529k in   5.086578s
            8 select    119.380k (± 1.3%) i/s    (8.38 μs/i) -    608.787k in   5.100414s
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [x86_64-linux]
        13 no select     8.096k i/100ms
           13 select     7.200k i/100ms
        13 no select     81.723k (± 1.6%) i/s   (12.24 μs/i) -    412.896k in   5.053606s
           13 select     71.629k (± 3.0%) i/s   (13.96 μs/i) -    360.000k in   5.030310s
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [x86_64-linux]
        21 no select     4.973k i/100ms
           21 select     4.317k i/100ms
        21 no select     49.596k (± 1.5%) i/s   (20.16 μs/i) -    248.650k in   5.014517s
           21 select     43.132k (± 2.5%) i/s   (23.18 μs/i) -    215.850k in   5.007465s
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [x86_64-linux]
        34 no select     3.239k i/100ms
           34 select     2.834k i/100ms
        34 no select     31.729k (± 1.3%) i/s   (31.52 μs/i) -    158.711k in   5.002867s
           34 select     27.683k (± 1.9%) i/s   (36.12 μs/i) -    138.866k in   5.018056s
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [x86_64-linux]
        55 no select     2.022k i/100ms
           55 select     1.774k i/100ms
        55 no select     20.212k (± 1.2%) i/s   (49.48 μs/i) -    101.100k in   5.002784s
           55 select     17.724k (± 1.6%) i/s   (56.42 μs/i) -     88.700k in   5.005857s
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [x86_64-linux]
        89 no select     1.255k i/100ms
           89 select     1.104k i/100ms
        89 no select     12.553k (± 0.9%) i/s   (79.67 μs/i) -     64.005k in   5.099372s
           89 select     11.097k (± 1.3%) i/s   (90.11 μs/i) -     56.304k in   5.074433s
```

I think we'd be fine with the extra overhead, but totally understand that might not be the case for every user of sidekiq. I'll update the PR to use the more restrictive rescue.

tomchipchase:
Checking the exception looks like it would be a case of having to make sure it still works with each ActiveSupport version. We're still on 7.1, and the error raised there was a NoMethodError that has a `cause` which is the `NoMethodError` on the CurrentAttributes. Here, in 8.0, the error is directly from the CurrentAttributes.

Instead I've gone with an additional boolean that just checks if we've successfully set the attributes, and doesn't retry if we have.

mperham:
Much happier with that solution, well done!

PR #6689, opened 2025-04-27
fix: reorder Sidekiq job and ActiveJob adapter requires

seuros:
In a typical Rails setup, Rails itself is loaded first, followed by Bundler.require, which then loads Sidekiq.

However, in various test suites, Sidekiq is manually required before loading Rails with `require_relative '../config/environment'`. This can lead to unexpected initialization issues.

This patch ensures that Sidekiq loads after Rails is fully initialized, maintaining the correct loading order across all environments.

_No production environments were harmed in the making of this change._


mperham:
How do we get rid of that Copilot bullshit?

mperham:
Oh, I see you asked for it. I don't know what value it provides you, it doesn't look useful to me.

seuros:
GitHub Copilot hallucinates a lot in big projects, but it's an excellent description writer. 

Sometimes I push code and notice it's describing more than what's actually in the PR. That's when I realize I've pushed too many files or didn't split my work enough (like when it mentions gem updates). If the PR is too big, Copilot just faint in a crash, and that's when I realize that if AI can't review it, humans won't either.

jdelStrother:
I still think this loads ActiveJob::Base too soon. eg in a fresh rails app, add a config/initializers/active_job.rb, containing

```ruby
Rails.application.config.active_job.logger = Logger.new('jobs.log')
```

Prior to sidekiq 8.0.3, this would result in ActiveJob::Base.logger writing to jobs.log.
As-of sidekiq 8.0.3 (or with this PR), it doesn't affect ActiveJob::Base.logger - it would use Rails' default logger instead.


I think it needs to any references to ActiveJob::Base need to happen within a `ActiveSupport.on_load(:active_job_base)` hook. (Which might be easier said than done)

PR #6694, opened 2025-05-01
Reduce minimum rack version.

benubois:
Hello,

Wanted to check in about the rack minimum version change. Feedbin uses unicorn and unicorn requires `rack < 3` so I've got a gem conflict there.

I tried this out locally and all tests passed the the web UI loaded fine, but I'm not sure if there were other reasons for the change. Happy to try to address them if so.

Thanks!

Screenshot 2025-05-01 at 5 06 38 PM

### DIFF
--- a/sidekiq.gemspec
+++ b/sidekiq.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency "redis-client", ">= 0.23.2"
   gem.add_dependency "connection_pool", ">= 2.5.0"
-  gem.add_dependency "rack", ">= 3.1.0"
+  gem.add_dependency "rack", ">= 2.0.0"
   gem.add_dependency "json", ">= 2.9.0"
   gem.add_dependency "logger", ">= 1.6.2"
 end


mperham:
If I allow a version, I have to support that version. I don't really want to support Rack 2.0 at this point. 2.0 came out in 2016 so that's 10 years of changes I'd need to understand. Even 2.2 is five years old.

Sidekiq 7.x supports Rack 2.2 but Sidekiq 8.0 removed all of the hacks and workarounds necessary to support 2.2 and 3 at the same time. I believe Rails 7.1+ also only supports Rack 3+?

The last release I see for Unicorn was in Dec 2021; is it safe to assume that Eric isn't supporting Unicorn anymore? I'd strongly consider migrating to a more modern app server. Pitchfork seems to be in good shape.

https://github.com/Shopify/pitchfork/blob/master/docs/MIGRATING_FROM_UNICORN.md
https://byroot.github.io/ruby/performance/2025/03/04/the-pitchfork-story.html

benubois:
Makes sense! 

There’s some more recent activity in the unicorn repo, but yeah it’s been a while since the last release. 

I like the look of pitchfork too, but there’s a bit of work attached to figure out what to do in the absence of hot reloading.

PR #6710, opened 2025-05-16
Don't load ActiveJob::Base during initialization

jdelStrother:
This is an extension of @seuros's PR #6689

Referencing ActiveJob::Base during the Rails initialization process can cause it to load too soon, which results in app-level config (eg `config.active_job.logger`) being ignored.  (These settings are only copied onto ActiveJob::Base once, the first time it loads at https://github.com/rails/rails/blob/fcf01b8820b6e4c3bd70ae3cc653229bbe5a1fba/activejob/lib/active_job/railtie.rb#L80-L82).
This broke in Sidekiq 8.0.3 due to https://github.com/sidekiq/sidekiq/pull/6669.

I've tried to demonstrate that in a new test case here - https://github.com/jdelStrother/sidekiq/blob/16c2f8589e3022fdc081ebd0196ad4a8b31e4774/test/rails_test.rb#L51-L59 - which passes on 8.0.2 and fails on 8.0.3. The test case itself is kind of hacky, open to better suggestions.

I think we can fix it by replacing references to `ActiveJob::Base` in active_job/sidekiq_adapter.rb with `ActiveSupport.on_load(:active_job) { ... }`, but would be good to get confirmation from @seuros that this still works ok with their ActiveJob-without-Rails use case.

### DIFF
--- a/lib/active_job/queue_adapters/sidekiq_adapter.rb
+++ b/lib/active_job/queue_adapters/sidekiq_adapter.rb
@@ -17,7 +17,7 @@ begin
     end
   end
 
-  unless ActiveJob::Base.respond_to?(:sidekiq_options)
+  ActiveSupport.on_load(:active_job) do
     # By including the Options module, we allow AJs to directly control sidekiq features
     # via the *sidekiq_options* class method and, for instance, not use AJ's retry system.
     # AJ retries don't show up in the Sidekiq UI Retries tab, don't save any error data, can't be
@@ -29,7 +29,7 @@ begin
     #     def perform
     #     end
     #   end
-    ActiveJob::Base.include Sidekiq::Job::Options
+    include Sidekiq::Job::Options unless respond_to?(:sidekiq_options)
   end
 
   # Patch the ActiveJob module
--- a/test/dummy/config/initializers/active_job.rb
+++ b/test/dummy/config/initializers/active_job.rb
@@ -1,0 +1,3 @@
+# frozen_string_literal: true
+
+Rails.application.config.active_job.logger = Logger.new(STDOUT)
--- a/lib/sidekiq/rails.rb
+++ b/lib/sidekiq/rails.rb
@@ -1,12 +1,11 @@
 # frozen_string_literal: true
 
-require "sidekiq/job"
-require_relative "../active_job/queue_adapters/sidekiq_adapter"
-
 module Sidekiq
   begin
     gem "railties", ">= 7.0"
     require "rails"
+    require "sidekiq/job"
+    require_relative "../active_job/queue_adapters/sidekiq_adapter"
 
     class Rails < ::Rails::Engine
       class Reloader
--- a/test/dummy/bin/rails
+++ b/test/dummy/bin/rails
@@ -1,0 +1,6 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+APP_PATH = File.expand_path("../config/application", __dir__)
+require_relative "../config/boot"
+require "rails/commands"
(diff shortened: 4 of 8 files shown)


lib/active_job/queue_adapters/sidekiq_adapter.rb
```diff
@@ -17,7 +17,7 @@ def perform(job_data)
     end
   end
 
-  unless ActiveJob::Base.respond_to?(:sidekiq_options)
+  ActiveSupport.on_load(:active_job) do
```
seuros, line 20: This will trigger only if railties exist. 

What if i ant a script with AJ and Sidekiq only ?

test/dummy/config/initializers/active_job.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+Rails.application.config.active_job.logger = Logger.new(STDOUT)
```
seuros, line 3: why ? this will default to Rails.logger.

lib/active_job/queue_adapters/sidekiq_adapter.rb
jdelStrother: I'm pretty sure this works without Railties - it's part of ActiveSupport:

```ruby
require "bundler/inline"
gemfile do
  gem "activejob"
end

require "active_job"

ActiveSupport.on_load :active_job do
  puts "In load hook!"
end

puts "Loading ActiveJob::Base..."
ActiveJob::Base
puts "Done!"
```

prints

```
Loading ActiveJob::Base...
In load hook!
Done!
```

test/dummy/config/initializers/active_job.rb
jdelStrother: I don't think that should matter - it's a demonstration of a config option on active_job being ignored if AJ::Base loads too soon.  Rails.application.config.active_job.retry_jitter is another one that's affected.
(Again, that particular option is a bit academic if you're using sidekiq, but my point is that Rails doesn't expect you to load the `Base` part of a framework until the very end of the initialization process)

mperham:
@seuros If you're ok with this, please merge and update Changes.md.

seuros:
@mperham , i will shortly.

jdelStrother:
Anything I can do to help get this merged?

jdelStrother:
Any chance of a new release with this? 🙏

mperham:
8.0.5 released

franzliedke:
I am getting this error now on boot with 8.0.5 and Rails 7.1:

```
NameError: uninitialized constant ActiveJob::QueueAdapters::AbstractAdapter (NameError)

      class SidekiqAdapter < AbstractAdapter
                             ^^^^^^^^^^^^^^^
Did you mean?  AbstractController
.../vendor/bundle/ruby/3.3.0/gems/sidekiq-8.0.5/lib/active_job/queue_adapters/sidekiq_adapter.rb:46:in `<module:QueueAdapters>'
.../vendor/bundle/ruby/3.3.0/gems/sidekiq-8.0.5/lib/active_job/queue_adapters/sidekiq_adapter.rb:37:in `<module:ActiveJob>'
.../vendor/bundle/ruby/3.3.0/gems/sidekiq-8.0.5/lib/active_job/queue_adapters/sidekiq_adapter.rb:36:in `<main>'
.../vendor/bundle/ruby/3.3.0/gems/sidekiq-8.0.5/lib/sidekiq/rails.rb:8:in `require_relative'
.../vendor/bundle/ruby/3.3.0/gems/sidekiq-8.0.5/lib/sidekiq/rails.rb:8:in `<module:Sidekiq>'
.../vendor/bundle/ruby/3.3.0/gems/sidekiq-8.0.5/lib/sidekiq/rails.rb:3:in `<main>'
```

jdelStrother:
It looks like AbstractAdapter was only introduced in Rails 7.2 - https://github.com/rails/rails/commit/e922c59207c406d51541622690737783e0ec338e
In which case, I think https://github.com/sidekiq/sidekiq/pull/6727 is the likely culprit

cc @djmb

PR #6712, opened 2025-05-19
ActiveJob Arguments via Serialization

ryansch:
This is still very WIP. I haven't included any of the code to make AJ args display nicely in the web interface.

I'm looking for general approach feedback.

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -137,6 +137,10 @@ module Sidekiq
     yield default_configuration unless server?
   end
 
+  def self.serialize_as(type)
+    default_configuration.serialize_as(type)
+  end
+
   # We are shutting down Sidekiq but what about threads that
   # are working on some long job?  This error is
   # raised in jobs that have not finished within the hard
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -137,10 +137,6 @@ module Sidekiq
     yield default_configuration unless server?
   end
 
-  def self.serialize_as(type)
-    default_configuration.serialize_as(type)
-  end
-
   # We are shutting down Sidekiq but what about threads that
   # are working on some long job?  This error is
   # raised in jobs that have not finished within the hard
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -410,7 +410,7 @@ module Sidekiq
 
     def display_args
       @display_args ||= begin
-        job_args = Sidekiq::Displayers.display_args(item)
+        job_args = Sidekiq::Displayers.display_args(self)
 
         if self["encrypt"]
           # no point in showing 150+ bytes of random garbage
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -409,7 +409,6 @@ module Sidekiq
     end
 
     def display_args
-      # Unwrap known wrappers so they show up in a human-friendly manner in the Web UI
       @display_args ||= begin
         job_args = Sidekiq::Displayers.display_args(item)
 
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -4,6 +4,7 @@ require "zlib"
 
 require "sidekiq"
 require "sidekiq/metrics/query"
+require "sidekiq/displayers"
 
 #
 # Sidekiq's Data API provides a Ruby object model on top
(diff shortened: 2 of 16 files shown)


ryansch:
@mperham Possible AJ args approach.

lib/sidekiq/api.rb
```diff
@@ -409,23 +410,15 @@ def display_class
...
-          job_args.drop(3)
-        elsif (self["wrapped"] || args[0]) == "ActionMailer::MailDeliveryJob"
-          # remove MailerClass, mailer_method and 'deliver_now'
-          job_args.drop(3).first.values_at("params", "args")
-        else
-          job_args
-        end
-      else
+      @display_args ||= begin
+        job_args = Sidekiq::Displayers.display_args(item)
```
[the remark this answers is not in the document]
ryansch: I'm also open to a better name here.

lib/sidekiq.rb
```diff
@@ -137,6 +137,10 @@ def self.configure_client
     yield default_configuration unless server?
   end
 
+  def self.serialize_as(type)
```
mperham, line 140: Not a fan of the global method. Putting it in S::Config should allow us to do:

```ruby
Sidekiq.configure_client do |cfg|
  cfg.serialize_as(...)
end
```

Better?

lib/sidekiq/displayers/active_job.rb
```diff
@@ -0,0 +1,77 @@
+module Sidekiq
+  module Displayers
+    module ActiveJob
+      ACTIVE_JOB_PREFIX = "_aj_"
+      GLOBALID_KEY = "_aj_globalid"
+      SYMBOL_KEYS_KEY = "_aj_symbol_keys"
+      RUBY2_KEYWORDS_KEY = "_aj_ruby2_keywords"
+
+      def self.valid_for?(job)
```
mperham, line 9: I'd like to see an effort to move away from global/class methods. Store an instance in S::Config and reference that instance rather than global methods and ifs everywhere.

lib/sidekiq/serializers.rb
```diff
@@ -0,0 +1,46 @@
...
+module Sidekiq
+  module Serializers
+    def self.serializer
+      serializers[Sidekiq.default_configuration[:serialize_as]] ||
+        raise(ArgumentError, "Invalid serializer: #{Sidekiq.default_configuration[:serialize_as]}")
+    end
+
+    def self.serializers
+      @serializers ||= {
+        basic: Sidekiq::Serializers::Basic
```
mperham, line 12: Why not use instances here, e.g. `S::Serializers::Basic.new`?

lib/sidekiq/serializers/active_job.rb
```diff
@@ -0,0 +1,26 @@
...
+      def self.deserialize(job)
+        job["args"] = ::ActiveJob::Arguments.deserialize(job["args"])
+        job.delete("_f")
+        job
+      end
+    end
+  end
+end
+
+Sidekiq::Serializers.register_serializer(:active_job, Sidekiq::Serializers::ActiveJob)
```
mperham, line 26: Any reason not to put this stuff directly in S::Config?

lib/sidekiq/displayers/active_job.rb
ryansch: @mperham Honestly, what's the difference? I can make it into a singleton but isn't it functionally identical?

lib/sidekiq/serializers.rb
ryansch: I don't currently have a need to store state on that object.

lib/sidekiq/serializers/active_job.rb
ryansch: I wanted to build an interface that other gems could extend.

We could certainly move it into the config object but I didn't see the need at this point.

lib/sidekiq.rb
ryansch: We will have to be careful to call `serialize_as` on both the client and server though. A retry might blow up otherwise.

lib/sidekiq/serializers/active_job.rb
ryansch: We have a pattern where the client and server have _different_ configurations. I'm not sure it makes sense for serializers/displayers to live in those areas.

lib/sidekiq/serializers.rb
ryansch: Put another way: this pattern gets interesting to me when you have something to pass to `new`

lib/sidekiq.rb
ryansch: I pulled it out.

lib/sidekiq/serializers.rb
ryansch: I suppose we could do something similar to the middleware chain where we allow the user to pass arguments to the serializers.

test/api_test.rb
```diff
@@ -304,7 +304,7 @@ class JobWithTags
 
           expected_args_by_version = {
             "5.x" => [1, 2, 3],
-            "6.x" => [{"user" => "gid://app/User/1"}, [1, 2, 3]],
+            "6.x" => [{user: "gid://app/User/1"}, [1, 2, 3]],
```
ryansch, line 307: @mperham Now that we support symbols, the API responses can be different. How do you want to handle this?

mperham:
Could you move this PR into a sidekiq/sidekiq branch so I can work on it too?

ryansch:
@mperham Absolutely!

ryansch:
@mperham I don't have the correct permissions to create a branch directly on sidekiq/sidekiq.

ryansch:
Closing in favor of https://github.com/sidekiq/sidekiq/pull/6715

PR #6713, opened 2025-05-21
fix: We should load rails if rails engine is defined for rspec

marcmillien:
We had an issue in our test suite where `RAILS_ENV=test` seems not to be set with `bundle exec rspec ...`.
`RAILS_ENV=test bundle exec rspec ...` fixed the problem, it should not be required

We tracked the issue to be from sidekiq 8.0.3.

It seems that getting back the following line solved the issue:
`require "sidekiq/rails" if defined?(::Rails::Engine)`

(instead of `require "sidekiq/rails"`)

https://github.com/sidekiq/sidekiq/compare/3765cfd67b0dc5d1b8d6d85739c6e2cdd0aa2f2a...205df02327a2402f4c844df07324d8ac3da80ffb 

Maybe related to https://github.com/sidekiq/sidekiq/issues/6683

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -146,4 +146,4 @@ module Sidekiq
   class Shutdown < Interrupt; end
 end
 
-require "sidekiq/rails"
+require "sidekiq/rails" if defined?(::Rails::Engine)


mperham:
Please try main. We've fixed some things since 8.0.3 was released.

benoittgt:
Hello 

I can confirm that the issue is still visible with last main commit https://github.com/sidekiq/sidekiq/commit/e5ba0ec77552d7d02152a38cb448608686fcb07a but disapear if I apply this patch. 

(I am working on the same project as Marc).

mscrivo:
FWIW, this also caused problems for our shoryuken workers. They all failed to start with:

> ERROR: Processor failed: uninitialized constant ActiveJob::QueueAdapters::ShoryukenAdapter

Our hunch is:
1. shoryuken loads, but doesn't require the active job adapter because activejob hasn't loaded.
2. sidekiq 8.0.3 loads and requies activejob for us. it assumes that if it's there, we want to use it.
3. Later when shoryuken computes the worker name, it sees activejob has loaded and tries to initialize the adapter. But the adapter hasn't been required because of step 1.

Some context: We have a monolith where we have the rails gem included, but not required, in our Gemfile, since not every project in our monolith uses rails. We only introduced rails to our codebase in the last year or so. We have a single image with various entrypoints, many of which don't load rails at all.

PR #6715, opened 2025-05-22
ActiveJob Arguments via Flavor

ryansch:
Continued from https://github.com/sidekiq/sidekiq/pull/6712.

Example config:
```ruby
  config.flavor_with(:active_job)
  # ...
end
  config.flavor_with(:active_job)
  # ...
end
```


lib/sidekiq/displayers.rb
```diff
@@ -0,0 +1,39 @@
...
+      displayer
+    end
+
+    def self.display_args(job_record)
+      displayer_for(type: :job, item: job_record).display_args(job_record)
+    end
+
+    def self.display_hash(hash)
+      displayer_for(type: :hash, item: hash).display_hash(hash)
+    end
```
ryansch, line 35: Should these just be methods required on a serializer?

test/helper.rb
```diff
@@ -3,8 +3,8 @@
 require "bundler/setup"
 Bundler.require(:default, :test)
 
-require "minitest/pride"
 require "maxitest/autorun"
```
ryansch, line 6: Moving `maxitest/autorun` above the other minitest related requires caused "line focusing" to start working again.

test/api_test.rb
```diff
@@ -304,7 +304,7 @@ class JobWithTags
 
           expected_args_by_version = {
             "5.x" => [1, 2, 3],
-            "6.x" => [{"user" => "gid://app/User/1"}, [1, 2, 3]],
+            "6.x" => [{user: "gid://app/User/1"}, [1, 2, 3]],
```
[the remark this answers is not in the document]
ryansch: This is only for display_args.

lib/sidekiq/displayers.rb
ryansch: The answer is yes.

lib/sidekiq/serializers.rb
```diff
@@ -0,0 +1,54 @@
+require "sidekiq/serializers/basic"
+
+module Sidekiq
+  module Serializers
```
ryansch, line 4: `Serializers` -> `Flavors`

ryansch: (I'm assuming we're going with the American spelling.)

lib/sidekiq/flavors/active_job.rb
```diff
@@ -0,0 +1,125 @@
...
+        when :job
+          job_record = item
+          job_record.klass == "ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper" || job_record.klass == "Sidekiq::ActiveJob::Wrapper" || job_record["_f"] == "aj"
+        when :hash
+          hash = item
+          hash["_f"] == "aj"
+        end
+      end
+
+      def self.display_args(job_record)
```
ryansch, line 59: @mperham How do we want to handle the case where a user has multiple ruby services sharing a sidekiq server? It's possible that only some of them are Rails-based and the server might not have this flavor loaded. It's a bit of an edge case, do we care?

mperham: I'm not sure that's a mode of operation we want to officially support. Just fall back to `to_s`?

mperham:
Sorry for the delay. This isn't dead, I'm just taking my time in deliberating the direction I want to go here.

I'm thinking about making the Rails support go all in: if you're using native jobs in a Rails app, you'd need to use the Active Job flavor but it would give you symbols + GlobalID, etc. This would be for Sidekiq 9.

ryansch:
For what it's worth, I'm building on top of this feature.

mperham:
I don't have an ETA for this. I haven't been able to get an implementation that I'm happy with and summer vacation time is about to ramp up so my productivity will plummet until August/Sept but few changes means it should be easy to keep your branches up-to-date.

ryansch:
@mperham No worries! It's working great for my purposes and we can port to your solution in the future.

ryansch:
@mperham We're running this in production now and it works great. Thoughts?

mperham:
I'm sorry Ryan, I never really got it into a form I was comfortable with. Maybe this is something I can revisit for 9.0.

ryansch:
@mperham We're still using this in production!

mperham:
Ryan, I'm still unconvinced this is a good direction for Sidekiq, that the benefits here outweigh the additional complexity. Sidekiq needs to maintain independence from Rails and Active Job for broad compatibility with the current and future Ruby ecosystem. Additionally flavors mean that persisted job payloads within Redis will have different formats, adding additional complexity for plugins which want to add functionality based on the payload.

I absolutely want to make Active Job usage as ergonomic as possible, but the cost here is too high.

Since I don't have any plans to merge this, I'm going to close it.

joevandyk:
@mperham was curious about the state of keyword arguments and followed the various PR/issues here.. I'm assuming that still no support for keyword arguments in jobs and nothing's actively planned for it?

mperham:
There's still no keyword argument support for native Sidekiq. You'd need to use Active Job. Things I think about:

1. Sidekiq's lingua franca is JSON and I view that as a strength: adopting conventions and standards are empowering and provide for interoperability with other ecosystems (i.e. other languages, Redis JSON support, etc).
2. Providing a user-selectable flavor adds significant complexity to anything touching job data, in return for better Ruby ergonomics with `perform_async`. Features which work with persistent job data would need to support different flavors.
3. Today, parsing a persisted job just requires `JSON.parse`, not a more complex deserialization scheme.
4. The simplicity benefits performance and scalability. Simple is fast.

I'm not against keyword argument or Symbol support but I haven't found a trade off I'm happy with so far.

ryansch:
@mperham Could we instead come up with injection points that plugins could use?

My team has been using this feature in production for a while and it's _amazing_ to be able to pass objects to sidekiq without having to worry about the serialization process. There's an entire class of code I wrote at my last job that isn't even needed with this approach.

I get wanting to keep sidekiq the way it is but I'd really love a way to extend it that isn't going to break.

In our case: Simple code is fast.

Edit: Using ActiveJob means giving up all of the things that Sidekiq is good at. See: Batches.

Edit 2: You pushed back on my original design looking for a way to add "flavour" to jobs instead of requiring it. Let your users use flavours instead of requiring the least common denominator.

ryansch:
I've been working around Sidekiq's lack of kwarg support for a number of years now. Maybe it's time to treat Sidekiq as a provider?

mperham:
Sometimes it takes a long time and/or several attempts to get something which works well. I know that's frustrating but I'm working on a new branch right now which shows promise. I'll push it real soon, would love feedback.

mperham:
#6966

ryansch:
@mperham Thank you so much! I was really concerned that I was going to be stuck with a patch set I had to maintain.

Issue #6717, opened 2025-05-24
Locale selection does not follow the Accept-Language header spec


pbstriker38:
Ruby version: 3.4.2
Rails version: 8.0.2
Sidekiq / Pro / Enterprise version(s): sidekiq (8.0.3) / sidekiq-pro (8.0.1)

The web UI is defaulting the first 5 character match of the `Accept-Language` header. This is not how the spec is defined. It should be matching in order of quality value.

For example, this value is returning `zh-TW` when it should be matching on `en` first.
```
en-US,en;q=0.9,es;q=0.8,zh-TW;q=0.7,zh;q=0.6
```

Other weirdness comes into play with `zh-XX` locales because there is no `zh` locale. So an accept header with a zh variant that is not `zh-CN` or `zh-TW` will match `zh-CN`. This is not a valid fallback language and would be better to just fallback to the default `en`.

The [user_prefered_languages](https://github.com/sidekiq/sidekiq/blob/v8.0.3/lib/sidekiq/web/helpers.rb#L169) method already sorts locales in the header based on their quality value, so the `locale` method just needs to respect it.

I can work on the fix if needed.

pbstriker38:
Are you in agreement that that language tags with subtags should only fallback to the primary language tag and not another subtag? So, `zh-HK` should not fallback to `zh-CN`.

PR #6721, opened 2025-05-27
Update Accept-Language header locale matching

pbstriker38:
- Matching is case-insensitive
- Exact matches are done in order of quality value first
- Fallback to base language is attempted next, "es-CO" => "es"
- If no matches, then fallback to "en"

Fixes: #6717

https://www.rfc-editor.org/rfc/rfc9110.html#section-12.5.4
https://www.rfc-editor.org/rfc/rfc4647.html#section-3.4


mperham:
Seems fine and the code looks great, thanks for the additional tests.

PR #6723, opened 2025-05-28
Web `/stats` endpoint does not return Valkey or Dragonfly Versions

maddiesch:
We noticed that after migrating to Valkey and Sidekiq 8, the initial dashboard page load would display the correct 8.0.1 version number for Valkey. But after the auto-refresh of data, that number would switch back to 7.2.4.

This change introduces a JavaScript helper function that returns the correct version number for the backing store. The order is the same as the behavior in `#store_version` in helpers.rb.

### DIFF
--- a/lib/sidekiq/web/application.rb
+++ b/lib/sidekiq/web/application.rb
@@ -9,7 +9,7 @@ module Sidekiq
       extend Router
       include Router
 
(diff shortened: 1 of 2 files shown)


mperham:
This isn't the right solution. "redis_version" should be "store_version", the key should not change based on Redis flavor. Make sense?

mperham:
I think that solves your issue and retains the (overly eager IMO) redis version update.

PR #6727, opened 2025-06-03
Make SidekiqAdapter subclass AbstractAdapter

djmb:
https://github.com/rails/rails/pull/55127 adds a new `stopping?` method to the `AbstractAdapter`.

`SidekiqAdapter` should implement that at some point, but for now by subclassing `AbstractAdapter` we ensure that it falls back to the default implementation.


mperham:
The same method can be found at `Sidekiq::Job#interrupted?`

mperham:
How does an Active Job access the adapter?

djmb:
Active Jobs have a [`queue_adapter` method](https://github.com/rails/rails/blob/c7c72ee6d39667014e7413c1430a0776b52801b2/activejob/lib/active_job/queue_adapter.rb#L27).

PR #6732, opened 2025-06-08
Modify SidekiqAdapter so it becomes the single source of truth for Sidekiq's specific logic.

ruyrocha:
Add missing `stopping?` method to `SidekiqAdapter`.

Update initializer with `@stopping` so it is initialized before Sidekiq hooks.

Alias properly `JobWrapper` to `Wrapper` since there are no customizations.

It's related with https://github.com/rails/rails/pull/55146 and https://github.com/rails/rails/pull/55146.


mperham:
I’d advise against callbacks in an initialize method. Makes for memory leaks.

ruyrocha:
Got it. Would it be OK to use `weakref` within the initializer to avoid memory leaks? cc @mperham

mperham:
I would use a single callback with a class variable. You’re tracking process shutdown so a global is reasonable but I’m open to other ideas.

mperham:
I’ll review and merge Monday. We have a few other Rails fixes which are pending for 805.

mperham:
Thanks!

ruyrocha:
Anytime! 

Quick question did you publish it? I'm receiving this:

```shell
 ~/D/R/fake-movie-detector-app (main) [7]> bundle install
Fetching gem metadata from https://rubygems.org/........
Resolving dependencies...
Your bundle is locked to sidekiq (8.0.5) from rubygems repository https://rubygems.org/ or installed locally, but that version can no longer be found in that source. That means the author of sidekiq (8.0.5) has removed it. You'll need to update
your bundle to a version other than sidekiq (8.0.5) that hasn't been removed in order to install.
```

mperham:
No, I generally let main sit for a day or two so people can report issues if they are running Sidekiq main with their CI.

ruyrocha:
Got it! Thank you so much, sir

PR #6737, opened 2025-06-09
Update selected text background color on dark mode.

### LINKED ISSUE #6734: [Sidekiq 8] Highlight Color Dark Mode
In Dark UI the highlight color when marking text is almost invisible. Please change so that it's clear what is being highlighted when copy / pasting job names or arguments from the Web UI

ruyrocha:
# main branch
Screenshot From 2025-06-09 17-23-12

# this branch
Screenshot From 2025-06-09 17-25-38

Closes #6734

### DIFF
--- a/web/assets/stylesheets/style.css
+++ b/web/assets/stylesheets/style.css
@@ -567,6 +567,7 @@ body > footer .nav {
     --color-border: oklch(25% 0.01 256);
     --color-input-border: oklch(31% 0.01 256);
     --color-selected: oklch(27% 0.01 256);
+    --color-selected-text: oklch(55% 0.11 45);
     --color-table-bg-alt: oklch(24% 0.01 256);
     --color-shadow: oklch(9% 0.01 256 / 10%);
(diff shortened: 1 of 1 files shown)


mperham:
Why do we need to set this at all? Shouldn’t the browser implement some logic to use a contrasting color automatically?

ruyrocha:
The color is set by the stylesheet file here https://github.com/sidekiq/sidekiq/blob/e255bf78b3b64cc5bbe16e153515a4bce4ff3891/web/assets/stylesheets/style.css#L569

image

And it looks like this within the UI
image

mperham:
Ah, then this is a dark mode bug in the styling and makes sense. If we remove line 569, does it look good in light and dark, I.e. is the browser default acceptable?

ruyrocha:
It looks like this when I comment out line 569
Screenshot From 2025-06-09 20-29-26

Other buttons look different as well
Screenshot From 2025-06-09 20-29-44

mperham:
Sounds reasonable, thanks!

PR #6750, opened 2025-07-02
Test against supported Rails versions

mlarraz:


mperham:
This takes CI from 12 runs to 48 runs. Is that too much? Where's the line in terms of excessive utilization for little benefit, ie diminishing returns? Would it be reasonable to test against latest minor version (8.0) and latest minor version from previous major (7.2)?

I get that that policy would have missed this regression but I'm trying to trade off between excessive CI runs on every single commit vs comprehensive testing.

mperham:
How about earliest supported minor (7.0) + whatever is current (i.e. unlocked, would be 8.0 today, could be 8.1 tomorrow)?

PR #6765, opened 2025-07-21
Introduce transactional client with AR tooling

rewritten:
At @amplemarket, we have several databases, and the existing AfterCommitEverywhere only supports transactions on one database at a time. Additionally, with the default configuration of the transactional client, only transactions in the default database are honored.

Since ActiveRecord 7.2, there is a hook that waits for all databases. Here is an alternative client that mimics exactly the existing one, but using ActiveRecord tooling instead of an external library.

### DIFF
--- a/lib/sidekiq/transaction_aware_client.rb
+++ b/lib/sidekiq/transaction_aware_client.rb
@@ -7,6 +7,12 @@ module Sidekiq
   class TransactionAwareClient
     def initialize(pool: nil, config: nil)
       @redis_client = Client.new(pool: pool, config: config)
+      @transaction_backend =
+        if ActiveRecord.version >= Gem::Version.new("7.2")
+          ActiveRecord.method(:after_all_transactions_commit)
+        else
+          AfterCommitEverywhere.method(:after_commit)
+        end
     end
 
     def batching?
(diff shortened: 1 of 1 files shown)


mperham:
Rather than introduce a separate class, what do you think about integrating this into the existing client? Pre 7.2 apps will use a_c_e and newer apps will use the built in API. I’d like to eventually migrate off a_c_e as that’s one less 3rd party dependency.

rewritten:
Much neater!

davidwessman:
Does this mean that we  should update the Wiki
https://github.com/sidekiq/sidekiq/wiki/Advanced-Options#transactional-push
to only require the gem if Rails 7.1 and below?

mperham:
Thanks for the reminder, updated.

colszowka:
Hi all,

since I ran into this I thought I'd leave a note about it here:

* After the upgrade to sidekiq 8.0.6 exactly one of our test cases started failing
* It's an integration spec that verifies some notifications got "sent", and it tests for 2 specific notifications being triggered by an API call
* We're using sidekiq testing fake adapter there and inspect the jobs queue

What we found was that this change actually changed the order the jobs ended up in on the sidekiq fake jobs list for inspection - it wasn't a big deal in our case, the order didn't matter and we just changed the expectation order and left it at that, and generally I guess in actual operations the order isn't in any way guaranteed anyway either and that's fine.

However, in case maybe someone stumbles across this, it seems that between after commit everywhere and this new logic, there is some subtle difference in the order of how the callbacks get triggered. 

Just wanted to leave this here in case someone runs across this too :)

davidwessman:
@colszowka we had the exact same. One test case failed after the upgrade.

Issue #6295, opened 2024-05-15
Update Sidekiq API with capsule data


stephenwiebe:
It appears that the Sidekiq data API doesn't expose any information about capsules. I think this would be an added attribute on the Process class? Basically I use a gem which uses the API in order to report process utilization (ie the busy/concurrency ratio), but I'm finding that I would like it to calculate this based on just the default capsule, not any other capsules.

mperham:
Sidekiq::Process should include the capsules, as they show up on the Busy tab.

stephenwiebe:
@mperham Can you confirm that Sidekiq::Process includes the capsules? Or can you let me know how to access the capsule data? I tried accessing it via `process['capsules']` and it doesn't seem to be valid. I also checked the code at https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/api.rb and didn't see anything on the Process class related to capsules. Nor does the documentation for the Process class at https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/api.rb#L1054 document any way to access capsule information.

mperham:
The only place that shows capsule setup is the Busy page. Take a look at busy.erb. I will also document.

stephenwiebe:
Ok, so in that case, it sounds like the request in this ticket still remains unimplemented? There's no way to use the Sidekiq API to access capsule information from Sidekiq::Process objects? Or am I misunderstanding?

mperham:
Yeah, diving in more, it looks like the API doesn't explicitly provide a capsule entity. Right now launcher.rb flattens the capsule data in order to maintain backwards compatibility so there is no way to see how capsules are configured in each process. We'd need to bring that data into Redis and the API.

Let's start with use cases first. Why do you need this info? What are you looking to do with it?

mperham:
Ah, you added your use case in the original comment.

mperham:
Comments welcome on that PR ^^^^^

PR #6775, opened 2025-08-04
Add and expose capsule data in Sidekiq::Process, see #6295

mperham:
Deprecate `weights` and `queues` as those elements are now lossy.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -5,6 +5,7 @@
 HEAD
 ----------
 
+- Updated `Sidekiq::Process` API to provide capsule data. The `queues` and `weights` data will be removed from Redis in Sidekiq 8.1, as this data can now be found in the `capsules` element. [#6295]
 - Job iteration now exposes `current_object` for easy access within the `around_iteration` callback [#6774]
 - Fix JS race condition which could skip confirmation dialogs when Live Polling [#6768]
 - Fix edge case which could lose CurrentAttributes [#6767]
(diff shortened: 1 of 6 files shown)


stephenwiebe:
Looks great to me!

web/views/busy.erb
```diff
@@ -69,11 +69,7 @@
             <% end %>
             <br>
             <b><%= "#{t('Queues')}: " %></b>
-            <% if process.weights %>
-              <%= busy_weights(process.weights) %>
-            <% else %>
-              <%= process.queues.sort.join(", ") %>
-            <% end %>
+            <%= process.queues.sort.join(", ") %>
```
fatkodima: @mperham Would you be open to extend this information with capsule names?

mperham: The whole UI layout needs to be rethought. It was designed well before Capsules existed. Want to create a mockup?

fatkodima: If you have ideas for a better UI, would appreciate it 🙏. I am not good at it.

mperham: I've thought about it. The conclusion I came to was that the information was simply not important enough to add more visual "noise" to the table. Without a solid usecase to understand how this info is useful on this page, it's just adding noise. We do note all of the queues that each process is handling.

jjb: (this PR removed the semicolons, in case you didn't realize - otherwise, I don't understand your comment 😅)

PR #6776, opened 2025-08-07
Update uk.yml to be more polite

loqimean:
🔄 Replaced harsh terminology: Killed → Stopped
This pull request updates the Ukrainian translations related to job termination in the UI. Specifically, we’ve replaced the word "вбиті" (translated from "Killed") with "зупинені" ("Stopped") to make the interface more user-friendly and emotionally neutral.

✅ What was changed:
"Dead" → "Зупинені"

"DeadJobs" → "Зупинені задачі"

"NoDeadJobsFound" → "Зупинених задач не знайдено"

"Kill" → "Зупинити"

"KillAll" → "Зупинити все"

🧠 Rationale:
The word "вбиті" ("Killed") sounds unnecessarily violent or aggressive in Ukrainian, which can be jarring to users — especially non-technical audiences.

"Зупинені" ("Stopped") conveys the same technical meaning (a job that was terminated and won’t be retried), but in a more neutral, professional, and user-friendly way.

This aligns with modern UX writing practices, which recommend avoiding emotionally charged or violent language in user interfaces wherever possible.

📌 Note:
If the context absolutely requires emphasis on forced termination (e.g., for internal tools or logs), the term "Примусово зупинені" ("Force-stopped") may be considered instead. However, for general UI use, "Зупинені" remains the preferred choice.

### DIFF
--- a/web/locales/uk.yml
+++ b/web/locales/uk.yml
@@ -14,8 +14,8 @@ uk:
   CreatedAt: Створено
   CurrentMessagesInQueue: Поточні задачі у черзі <span class='title'>%{queue}</span>
   Dashboard: Панель керування
-  Dead: Вбитих
-  DeadJobs: Вбиті задачі
+  Dead: Зупинених
+  DeadJobs: Зупинені задачі
   Delete: Видалити
   DeleteAll: Видалити усі
   Deploy: Деплой
@@ -33,8 +33,8 @@ uk:
   History: Історія
   Job: Задача
   Jobs: Задачі
-  Kill: Вбити
-  KillAll: Вбити все
+  Kill: Зупинити
+  KillAll: Зупинити все
   LastRetry: Остання спроба
   Latency: Затримка
   LivePoll: Постійне опитування
@@ -42,7 +42,7 @@ uk:
   Name: Назва
   Namespace: Простір імен
   NextRetry: Наступна спроба
-  NoDeadJobsFound: Вбитих задач не знайдено
+  NoDeadJobsFound: Зупинених задач не знайдено
   NoRetriesFound: Спроб не знайдено
   NoScheduledFound: Запланованих задач не знайдено
   NotYetEnqueued: Ще не в черзі


fatkodima:
>  Replaced harsh terminology: Killed → Stopped

There is a difference between "killed" and "stopped".

> The word "вбиті" ("Killed") sounds unnecessarily violent or aggressive in Ukrainian

I think "killed" is similarly aggressive in English too.

mperham:
In general I'm in favor of this style of change, making our terminology less violent and more accurate is worthwhile (c.f. Redis master/slave vs primary/replica).

Jobs aren't really alive or living but death and life are pretty core to the terminology we use and surround us in nature. "Dead" is the state in the job "life"cycle: created, enqueued, running, retrying, dead. I do wish I had used less violent terminology in English but these are the terms people have been using for years now when discussing their Sidekiq jobs and changing that requires the entire community to learn new terminology. I've done this before (c.f. Sidekiq::Worker -> Sidekiq::Job) but it's not easy and takes many years.

Since I don't speak Ukrainian, I will defer to @fatkodima and other Ukrainian users to decide what is appropriate in that language but thank you for opening the conversation.

loqimean:
> In general I'm in favor of this style of change, making our terminology less violent and more accurate is worthwhile (c.f. Redis master/slave vs primary/replica).
> 
> Jobs aren't really alive or living but death and life are pretty core to the terminology we use and surround us in nature. "Dead" is the state in the job "life"cycle: created, enqueued, running, retrying, dead. I do wish I had used less violent terminology in English but these are the terms people have been using for years now when discussing their Sidekiq jobs and changing that requires the entire community to learn new terminology. I've done this before (c.f. Sidekiq::Worker -> Sidekiq::Job) but it's not easy and takes many years.
> 
> Since I don't speak Ukrainian, I will defer to @fatkodima and other Ukrainian users to decide what is appropriate in that language but thank you for opening the conversation.



> Personally, I always try to use English interfaces wherever possible and recommend to other people too. Because in English the terminology is always the same between different programs. But each program decides to translate to other languages in different ways, so one word in English now have 3-5 alternatives in Ukrainian and you need to figure out what translator meant.
> 
> But these translations in this PR probably are the best possible for these words.

Hello, I appreciate your desire for improvement and a little revolution on the side of a less provocative world with comfort for everyone. Regarding English terminology, I agree with you and developers do almost always use English, but the web interface is also used by non-technical people, for whom it would be inconvenient for everyone to look at trigger words, especially considering the current situation in the country Ukraine. So, thank you for your trust and have a nice day, until new commits!

vmyts539:
> > Replaced harsh terminology: Killed → Stopped
> 
> There is a difference between "killed" and "stopped".
> 
> > The word "вбиті" ("Killed") sounds unnecessarily violent or aggressive in Ukrainian
> 
> I think "killed" is similarly aggressive in English too.

It's much more common term in English when it comes to this context, but not that much for Ukrainian. This change makes sense iyam

PR #6779, opened 2025-08-07
Fixes "Logs are colorized regardless of colorized_logging setting"

knightq:
Sidekiq::Logger: colorize logs only is `::Rails.application.config.colorize_logging`.


mperham:
1. Sidekiq must remain independent of Rails.
2. The value should be looked up and configured once when booting up, in sidekiq/rails.rb line ~50.

PR #6786, opened 2025-08-11
Remove custom CSS ::selection color

frederfred:
The selection color had too little contrast and overriding the browser default harms accessibility.

See https://developer.mozilla.org/en-US/docs/Web/CSS/::selection#accessibility:

> Don't override selected text styles for purely aesthetic reasons — users can customize them to suit their needs. For people experiencing cognitive concerns or who are less technologically literate, unexpected changes to selection styles may hurt their understanding of the functionality.

When selecting everything on the page (cmd+A), this is the result (last row in the table hovered):

### Before

selection_before_light
selection_before_dark

### After

selection_after_light
selection_after_dark

### DIFF
--- a/web/assets/stylesheets/style.css
+++ b/web/assets/stylesheets/style.css
@@ -29,8 +29,6 @@
 
 *, *::before, *::after { box-sizing: border-box; }
 
-::selection { background: var(--color-selected); }
-
 :focus-visible {
   outline: 1px solid oklch(from var(--color-primary) l c h / 50%);
 }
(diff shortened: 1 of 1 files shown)


mperham:
Should you remove any CSS related to “selected”? Like line 569.

frederfred:
Good question, and I should have mentioned in the PR why I kept it. `--color-selected` is used in a few other places (4 to be specific). Not great variable naming, and I could take on improving that and a few other CSS improvements that I see. But that should probably be its own PR.

PR #6787, opened 2025-08-12
show iterable state

joevandyk:
https://github.com/sidekiq/sidekiq/issues/6537

image

DEVELOPMENT  Sidekiq 2025-08-12 02-40-33

DEVELOPMENT  Sidekiq 2025-08-12 02-41-38

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -437,6 +437,30 @@ module Sidekiq
       self["jid"]
     end
 
+    # Returns persisted iteration state for an iterable job, if any.
+    # Reads Redis key "it-<jid>" set by Sidekiq::Job::Iterable and decodes cursor JSON.
+    # Includes executions (Integer), runtime (Float), cursor (Object), cancelled (Integer timestamp or nil).
+    def iterable_state
+      j = jid
+      return nil if j.nil? || j == ""
+
+      state = Sidekiq.redis { |c| c.hgetall("it-#{j}") }
+      return nil if state.nil? || state.empty?
+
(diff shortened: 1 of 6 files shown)


mperham:
I like parts of this but the fundamental issue is that this adds a significant amount of network chatter to the Busy and Queue pages. If we are looking at a page of queued jobs which shows 50 jobs, currently that's 1 call to Redis to call LRANGE in paginator.rb. With this change, we're going to make 51 round trips. I think it is worth refactoring to take advantage of pipelining.

Busy is even worse because large users might have hundreds of jobs running concurrently.

mperham (changes requested):
Needs performance tuning as noted.

fatkodima:
> but the fundamental issue is that this adds a significant amount of network chatter to the Busy and Queue pages

I wonder if it is really needed on Busy and Queue pages? Imo, having it on a details page would be enough.

mperham:
I think it would be fine to add to the job details page.

joshuaylee:
Hi, I picked this up since it's been sitting for a while, and I've been working with iterable jobs recently. Started a fresh branch off main since the original had diverged quite a bit.

Re: the suggestion to only show this on job detail pages — there's no detail page for running jobs on Busy, and iteration state is interesting to see for running jobs. So I kept the inline display in the table.

I'd be happy to open a PR if you're open to resuming this proposal. Here is the branch: https://github.com/joshuaylee/sidekiq/tree/jl/iterable-state-web

mperham:
Please do send a PR. I’m open to the feature.

PR #6792, opened 2025-08-14
Allow spreading batches of jobs over a specified interval

### LINKED ISSUE #6791: Option for easier spreading of bulk pushed jobs
We are currently seeing an issue - when many jobs are enqueued via push_bulk/perform_bulk, they start executing at the same time and can cause some db locking, because of accessing the same rows etc.

 The solution to this problem is to add some random delays to these jobs. The pattern is like:
```ruby
delay = 10.minutes
now = Time.current
timestamps = args.map { now.to_i + rand(delay) }
MyJob.perform_bulk(args, at: timestamps)   
```

Would you be open for an option to push_bulk/perform_bulk, that does basically the same. 
Something like:
```ruby
MyJob.perform_bulk(args, spread_over: 10.minutes)
```

The best option name is yet to be found.

fatkodima:
Closes #6791.

```ruby
MyJob.perform_bulk(args, spread_interval: 10.minutes)
```

I chose `spread_interval` name, because it reads better with simple numbers (`spread_interval: 10`), compared to `spread_over` (`spread_over: 10`).


mperham:
The `spread_interval` must be greater than or equal to Sidekiq::Config's `average_scheduled_poll_interval`, which defaults to 5 so I would make 5 the minimum value. For instance, spreading jobs over one second doesn't make any sense because Sidekiq's scheduled job polling doesn't check that frequently -- the jobs would effectively run simultaneously, which is what you are trying to avoid.

I would suggest instead of rand(spread_interval), that you use `rand * spread_interval` as that will get a smoother spread across the entire interval, otherwise you will get jobs clustered at the same moment of each second.

fatkodima:
Thanks for suggestions 👍 Updated.

lib/sidekiq/client.rb
```diff
@@ -138,6 +141,17 @@ def push_bulk(items)
       jid = items.delete("jid")
       raise ArgumentError, "Explicitly passing 'jid' when pushing more than one job is not supported" if jid && args.size > 1
 
+      spread_interval = items.delete(:spread_interval) || items.delete("spread_interval")
+      raise ArgumentError, "Jobs 'spread_interval' must be a positive Numeric" if spread_interval && (!spread_interval.is_a?(Numeric) || spread_interval <= 0)
+      raise ArgumentError, "Only one of 'at' or 'spread_interval' can be provided" if at && spread_interval
+
+      if !at && spread_interval
+        # Do not use spread interval smaller than pooling interval.
+        spread_interval = [spread_interval, 5].max
```
franzliedke, line 150: @fatkodima Should there be an explicit `to_i` or `to_f` here, to be consistent with the implementation of `perform_at`?

Right now, I don't dare to pass an `ActiveSupport::Duration`, even though it will probably work. 😉

PR #6794, opened 2025-08-21
use iteration_key method in Job::Iterable

joevandyk:
This pull request makes a small update to the job cancellation logic in the `Sidekiq::Job::Iterable` module. The change ensures that the cancellation status is checked using the correct iteration key, which improves accuracy when determining if a job has been cancelled.

* Updated the `is_cancelled?` method to use `iteration_key` instead of a hardcoded key format when checking for job cancellation status in Redis.


fatkodima:
Please do the same for `cancel!` method in that class (`key = iteration_key`).

PR #6795, opened 2025-08-21
add array_batches_enumerator

joevandyk:


fatkodima:
> My use case: in my iterable job, it's expensive/slow to compute the array contents on every iteration. Doing that on every 100 or 1000 saves substancial time.

Can you give more details about the use case?

joevandyk:
> > My use case: in my iterable job, it's expensive/slow to compute the array contents on every iteration. Doing that on every 100 or 1000 saves substancial time.
> 
> Can you give more details about the use case?

```
def build_enumerator(cursor:)
  # might take a couple seconds to run, or more
  tokens = SomeHugeTable.where(expensive_filters_and_joins: true).distinct.pluck(:canonical_token).sort
  array_enumerator(tokens, cursor: cursor)
end

def each_iteration(token)
  SomeHugeTable.where(canonical_token: token).each { ... }
end
```

would be nice to have the 'batches' thing that ActiveRecord and CSV supports.

fatkodima:
Had you seen this? https://github.com/sidekiq/sidekiq/blob/c8488d780352061f2f50877db74325e2ff97c1a3/lib/sidekiq/job/iterable/active_record_enumerator.rb#L22-L42

And the same for csv enumerator

fatkodima:
I misread the comment. I think you should use

```ruby
def build_enumerator(cursor:)
  # might take a couple seconds to run, or more
  relation = SomeHugeTable.where(expensive_filters_and_joins: true).distinct.select(:canonical_token)
  active_record_records_enumerator(relation, cursor:)
end

def each_iteration(record)
  ...
end
```

mperham:
I'm not clear on current status, please revise or close this.

PR #6802, opened 2025-09-02
Handle ActiveSupport::Logger.broadcast deprication

ojsdude:
As per https://apidock.com/rails/ActiveSupport/Logger/broadcast/class ActiveSupport::Logger.broadcast is no longer supported after rails v7.0.0

### DIFF
--- a/lib/sidekiq/rails.rb
+++ b/lib/sidekiq/rails.rb
@@ -48,8 +48,10 @@ module Sidekiq
           unless ::Rails.logger == config.logger || ::ActiveSupport::Logger.logger_outputs_to?(::Rails.logger, $stdout)
             if ::Rails.logger.respond_to?(:broadcast_to)
               ::Rails.logger.broadcast_to(config.logger)
-            else
+            elsif ::ActiveSupport::Logger.respond_to?(:broadcast)
               ::Rails.logger.extend(::ActiveSupport::Logger.broadcast(config.logger))
+            else
+              ::Rails.logger = ::ActiveSupport::BroadcastLogger.new(Rails.logger, config.logger)
             end
           end
         end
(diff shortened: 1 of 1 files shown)


mperham:
Any Sidekiq committers want to be in charge of reviewing Rails-related changes? I don't use Rails so I'd like another set of eyes to review. @seuros @fatkodima

ojsdude:
@mperham thank you for the quick reply.

Just for reference; I have tested this in production now @seuros and with the minor scope update it works perfectly!

mperham:
Since Sidekiq main only supports Rails 7+, is there any need for the `elsif`? That's only used for Rails 6 and below, right?

ojsdude:
> The last existing version (v7.0.0)

☝️ it looks like it's in 7.0.0 but happy to remove if you think it's the best solution.

ojsdude:
Alternatively is there a convention for TODO comments to remove the `elsif` when 7.0.0 is no longer supported? 🤔

seuros:
From the CI, we are not even testing 7.0.

The code could be simplified by removing the defensive coding.

ojsdude:
@mperham @seuros is there anything I need to do here then?

- Adding tests for this specific example?
- Removing the `elsif`

mperham:
That's what's stopping me from merging, yes. Does Rails 7.0 require the `elsif` or does it work without it? Remove it if possible.

ojsdude:
I can't get 7.0.0 running because of dependencies, but as far as I can see with version 7.0.8.7 `::ActiveSupport::BroadcastLogger.new` would not work here so I think we need to keep the `elsif` to support `7.0.*`

image

mperham:
Looks like BroadcastLogger was added in 7.1 so this PR looks correct.

ojsdude:
Thanks @mperham will this be in the next sidekiq release then? 👍

PR #6813, opened 2025-09-14
introduce quarantine queue

codepawpaw:
This pull request introduces a new Job Quarantine feature to Sidekiq that allows jobs to be automatically moved to a quarantine queue when they fail with specific exceptions for a configured number of times. This provides better error handling and prevents problematic jobs from continuing to retry indefinitely or immediately going to the dead queue.

## Benefits

1. **Better Error Handling**: Prevent problematic jobs from consuming retry resources
2. **Manual Intervention**: Allow developers to inspect and manually handle specific failures
3. **Reduced Noise**: Keep retry queue focused on transient failures
4. **Debugging Aid**: Quarantined jobs preserve full error context for investigation
5. **Operational Control**: Flexible criteria per job class for different failure scenarios

### Core Logic (`lib/sidekiq/job_retry.rb`)

The retry logic was enhanced with quarantine checking:

1. **Before Retry Exhaustion**: Check if job meets quarantine criteria
2. **Quarantine Check**: Match exception class and retry count against configured criteria
3. **Quarantine Action**: Move job to quarantine queue instead of continuing retries
4. **Fallback**: Jobs without quarantine criteria continue normal retry/dead flow

### Basic Usage
```ruby
class EmailJob
  include Sidekiq::Job
  quarantine_on Net::SMTPServerBusy, counts: 5
  
  def perform(email_data)
    # send email
  end
end
```

### Multiple Criteria
```ruby
class DataProcessingJob
  include Sidekiq::Job
  quarantine_on DataValidationError, counts: 1    # Immediate quarantine
  quarantine_on ExternalAPIError, counts: 10      # After many retries
  quarantine_on TimeoutError, counts: 3           # After few retries
  
  def perform(data)
    # process data
  end
end
```

### Core Files
- `lib/sidekiq/job_retry.rb` - Enhanced retry logic with quarantine support
- `lib/sidekiq/job.rb` - Added quarantine configuration methods
- `lib/sidekiq/api.rb` - Added QuarantineSet class and stats integration

### Web UI Files
- `lib/sidekiq/web/application.rb` - Added quarantine routes and handlers
- `web/views/quarantine.erb` - New quarantine job listing view
- `web/locales/en.yml` - Added quarantine-related translations

Screen Shot 2025-09-14 at 15 42 55

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -47,6 +47,10 @@ module Sidekiq
       stat :dead_size
     end
 
+    def quarantine_size
+      stat :quarantine_size
+    end
+
     def enqueued
       stat :enqueued
     end
@@ -88,12 +92,13 @@ module Sidekiq
           pipeline.zcard("schedule")
           pipeline.zcard("retry")
           pipeline.zcard("dead")
+          pipeline.zcard("quarantine")
           pipeline.scard("processes")
           pipeline.lindex("queue:default", -1)
         end
       }
 
-      default_queue_latency = if (entry = pipe1_res[6])
+      default_queue_latency = if (entry = pipe1_res[7])
         job = begin
           Sidekiq.load_json(entry)
         rescue
@@ -123,7 +128,8 @@ module Sidekiq
         scheduled_size: pipe1_res[2],
         retry_size: pipe1_res[3],
         dead_size: pipe1_res[4],
-        processes_size: pipe1_res[5],
+        quarantine_size: pipe1_res[5],
+        processes_size: pipe1_res[6],
 
         default_queue_latency: default_queue_latency
       }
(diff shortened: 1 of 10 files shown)


mperham:
The purpose of the dead set is to be a global quarantine queue, requiring no configuration. Why is the dead set not sufficient?

codepawpaw:
> The purpose of the dead set is to be a global quarantine queue, requiring no configuration. Why is the dead set not sufficient?

The dead set is a valuable global quarantine, but it relies on jobs reaching their retry limit. For jobs with a high CPU or memory footprint, this can lead to repeated, resource-heavy operations that strain the system.

The proposed quarantine mechanism addresses this by stopping the retry process early based on configurable criteria, such as a specific error type. This avoids repeated processing of a failing job and moves it to a paused state, where it can be inspected and dealt with later without consuming further resources

mperham:
Thanks. I appreciate all the work you've put into this PR but I do suggest opening an issue before spending so much time in the future. I don't think this feature has enough demand to justify, it seems really specialized. Most people will reduce the retries to something more like 8 or 10 and find that a sufficient solution.

Side note: you could implement this as server-side middleware and publish it as a 3rd party gem, `sidekiq-quarantine` or such. You'd need a bit of a hack to interact with the retry subsystem but I think it could be done.

PR #6819, opened 2025-09-24
Allow to configure max job runtime for iterable jobs

### LINKED ISSUE #6693: Document Iterable max_job_runtime (if supported?)
The https://github.com/fatkodima/sidekiq-iteration project that was the source for Sidekiq iteration, supports a `max_job_runtime` property, which interrupts an iterable job if it runs longer than a certain amount of time.This helps to prevent long-running iterable jobs from hogging all of the workers on the server.

I don't see this property documented on the official wiki: https://github.com/sidekiq/sidekiq/wiki/Iteration I also don't see the property used in the codebase.

However, I do see a default value of nil set in the default configuration: https://github.com/sidekiq/sidekiq/blob/5abdcbf18fc376632ee9b7d5822a8392051fedf4/lib/sidekiq/config.rb#L21

Is this property supported? If so, please add documentation to the wiki.

If it is not supported, I think it would be great to add - as one of my concerns using iterable (vs multiple jobs) is that if an iterable job takes an hour to run - it will hog the workers and not let other jobs interleave.

fatkodima:
Closes #6693.


mperham:
I'd rename the option to :max_iteration_runtime.

Can you add a code comment for that option which briefly describes what it will do? Specifically it's not clear what Sidekiq will do when the max runtime is triggered. Is it rescheduled? Become a retry? Re-enqueued at the back of the queue?

fatkodima:
Renamed the option name and added a brief description.

mperham:
Thanks. Here we re-enqueue:

https://github.com/sidekiq/sidekiq/blob/d6395641571eba33050d34526bf93bed92504d4d/lib/sidekiq/job/interrupt_handler.rb#L13

Would it be beneficial to store some state about iteration interrupts here? Like an `interrupt_count`?

fatkodima:
I don't know. Won't https://github.com/sidekiq/sidekiq/blob/d6395641571eba33050d34526bf93bed92504d4d/lib/sidekiq/job/iterable.rb#L29 be enough as an approximate number of interrupts?

`job-iteration` gem has a similar attribute https://github.com/Shopify/job-iteration/blob/13d981529f7b74dfaa929843e59d517c590a9384/lib/job-iteration/iteration.rb#L89

Can add, if you think it is worth it.

mperham:
Ah right, yeah I think that’s semantically identical or close enough.

lib/sidekiq/config.rb
```diff
@@ -17,10 +17,9 @@ class Config
       poll_interval_average: nil,
       average_scheduled_poll_interval: 5,
       on_complex_arguments: :raise,
-      iteration: {
-        max_job_runtime: nil,
-        retry_backoff: 0
```
mperham, line 22: Was this unused? Should we worry about backwards compatibility?

fatkodima: This was unused and you forgot to remove this part when removed this feature from the original PR, so just removing lgtm.

lib/sidekiq/job/iterable.rb
```diff
@@ -273,6 +273,11 @@ def build_enumerator(params, cursor:)
         end
       end
 
+      def should_interrupt?
```
franzliedke, line 276: Can this method be considered a public feature of iterable jobs now? I was previously overwriting `interrupted?` to add a custom per-job max runtime. This one makes it much easier. 👍🏼

lib/sidekiq/job/iterable.rb
```diff
@@ -308,6 +313,10 @@ def handle_completed(completed)
           raise "Unexpected thrown value: #{completed.inspect}"
         end
       end
+
+      def mono_now
+        ::Process.clock_gettime(::Process::CLOCK_MONOTONIC)
+      end
```
franzliedke, line 319: Nice. 😎 

Would you also accept a method like this?

```ruby
def runtime = mono_now - @_start_time
```

Very handy, e.g. in combination with my other suggestion.

franzliedke (commented):
Two questions for @mperham. Happy to send a PR if you like the ideas.

lib/sidekiq/job/iterable.rb
mperham: If it's a private method, it is not public.

lib/sidekiq/job/iterable.rb
mperham: Is that runtime or wall time? Iterable jobs can be interrupted and sit on the queue, not executing.

PR #6821, opened 2025-09-28
Update restricted uses section in COMM-LICENSE.txt

mensfeld:
This EULA has been poiting to incorrect location on the modifications usage. This fixes this pointing to modification rules section (2) instead of ownership (4)

Note: Not sure why it highlights in diff whole section because I changed only in `g)` the reference from 4 to 2 :shrug:


mperham:
Thank you!

Issue #6733, opened 2025-06-09
[Sidekiq 8] Retry All / Delete All


OskarEichler:
Can we please get back the "Retry All" button in the Retry queue.

And also the "Delete All" button in the Dead queue.

mperham:
We've discussed this previously. I'm open to it but there are serious UX issues (especially when mixed with the filter functionality).

OskarEichler:
When Filter is active or certain jobs are selected maybe it should say "Retry Selected" or "Delete Selected". If none are selected or filtered can say "Retry All" / "Delete All" - so it's clear.

tansengming:
thanks for looking into it @mperham !

Here's why it's needed in case you needed a reason. The Heroku API is down right now and I need to retry all 20,000 items on the retry queue. While I write this, I am taking a break from clicking on the select all checkbox, the "retry now" button for the last 30 mins. I realize this might be a one time issue, but dang I wish there was a retry all button this one time!

mperham:
It’s a one line API call. Make a rake task?

tansengming:
Sorry I didn't make it clear. I can't run rake right now because the Heroku API is down 😢 . All rakes, deploys, server operations are disabled. So all I have is the web interface.

To be clear this rarely happens on Heroku. Unfortunately it's happening right now.

mperham:
If Heroku is down I don't think you have any options.

The endpoint is still there but it requires a CSRF token which can't be determined. :-(

```
% curl -X POST http://localhost:3000/sidekiq/retries/all/retry
Forbidden
```

Here's a hack: add `$TESTING = true` in your initializer and restart Puma. Sidekiq will exclude the CSRF protection, allowing you to call that endpoint. I realize that's a bit convoluted but it's a workaround. Unfortunately I think you need the Heroku API to restart Puma.

mperham:
I’m going to look into removing CSRF for 8.1. SameSite cookies should provide sufficient protection these days (speak up if you know different, I’m not a security expert).

tansengming:
Thanks for the tips @mperham ! For folks who are following, Heroku is still broken. But we can now run the rails runner through Heroku which has made things a lot easier.

Also a hack to retry a lot of jobs is to filter the jobs with a very loose search term (e.g. "1"). This can return a surprisingly large number of jobs.

shleeable:
+1 I make regular use of the retry all buttons.....

XenorPLxx:
[...]

PR #6833, opened 2025-10-15
Add bulk job management actions back to Retries, Dead, and Scheduled tabs

deve1212:
image

### Summary
This PR adds three additional bulk action buttons — **Retry All**, **Delete All**, and **Kill All** — to the Retries page in the Sidekiq Web UI.  

These buttons provide a more efficient way to manage large retry queues without manually selecting all jobs or using the console.

### Details
- Added three buttons in the existing form alongside "Retry Now", "Delete", and "Kill".
- Each button uses `formaction` to target the existing endpoints:
  - `/retries/all/retry`
  - `/retries/all/delete`
  - `/retries/all/kill`
- Maintains existing CSRF protection and styling (`btn btn-primary` / `btn btn-danger`).
- No JavaScript or dynamic “Selected vs All” behavior added — this version focuses on restoring the previously requested bulk action buttons.

### Motivation
Restores functionality that multiple users have requested in issue discussions (e.g. [#6705](https://github.com/mperham/sidekiq/issues/6733)) and recent GitHub comments where the **Retry All** and **Delete All** buttons were found useful, especially during large retry queue operations or platform outages (like Heroku downtime).

### Testing
- Verified buttons correctly trigger existing bulk endpoints.
- Confirmed CSRF token and form submission are preserved.
- Checked visual layout — all six buttons display on one row with minimal CSS changes.

### Future improvements
- Could later add dynamic labels (“Retry Selected” vs “Retry All”) when filters or checkboxes are active.
- Potential refactor to unify form actions or enhance accessibility.

### DIFF
--- a/examples/webui-ext/lib/sidekiq-redis_info.rb
+++ b/examples/webui-ext/lib/sidekiq-redis_info.rb
@@ -3,6 +3,7 @@ require "sidekiq"
 module SidekiqExt
   class RedisInfo
     include Enumerable
+
     def initialize
       @info = Sidekiq.default_configuration.redis_info
     end
--- a/lib/sidekiq/fetch.rb
+++ b/lib/sidekiq/fetch.rb
@@ -7,6 +7,7 @@ require "sidekiq/capsule"
 module Sidekiq # :nodoc:
   class BasicFetch
     include Sidekiq::Component
+
     # We want the fetch operation to timeout every few seconds so the thread
     # can check if the process is shutting down.
     TIMEOUT = 2
--- a/lib/sidekiq/middleware/i18n.rb
+++ b/lib/sidekiq/middleware/i18n.rb
@@ -11,6 +11,7 @@ module Sidekiq::Middleware::I18n
   # to be sent to Sidekiq.
   class Client
     include Sidekiq::ClientMiddleware
+
     def call(_jobclass, job, _queue, _redis)
       job["locale"] ||= I18n.locale
       yield
@@ -20,6 +21,7 @@ module Sidekiq::Middleware::I18n
   # Pull the msg locale out and set the current thread to use it.
   class Server
     include Sidekiq::ServerMiddleware
+
     def call(_jobclass, job, _queue, &block)
       I18n.with_locale(job.fetch("locale", I18n.default_locale), &block)
     end
--- a/lib/sidekiq/profiler.rb
+++ b/lib/sidekiq/profiler.rb
@@ -11,6 +11,7 @@ module Sidekiq
     }
 
     include Sidekiq::Component
+
     def initialize(config)
       @config = config
       @vernier_output_dir = ENV.fetch("VERNIER_OUTPUT_DIR") { Dir.tmpdir }
--- a/lib/sidekiq/ring_buffer.rb
+++ b/lib/sidekiq/ring_buffer.rb
@@ -6,6 +6,7 @@ module Sidekiq
   class RingBuffer
     include Enumerable
     extend Forwardable
+
     def_delegators :@buf, :[], :each, :size
 
     def initialize(size, default = 0)
--- a/myapp/app/sidekiq/hard_job.rb
+++ b/myapp/app/sidekiq/hard_job.rb
@@ -1,5 +1,6 @@
 class HardJob
   include Sidekiq::Job
+
   sidekiq_options backtrace: 5
 
   def perform(name, count, salt)
--- a/myapp/config/initializers/sidekiq.rb
+++ b/myapp/config/initializers/sidekiq.rb
@@ -51,6 +51,7 @@ end
 
 class Singler
   include Sidekiq::ServerMiddleware
+
   def call(w, j, q)
     puts "single-threaded #{w.class.name}!"
   end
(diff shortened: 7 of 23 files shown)


mperham:
The buttons need to pull right so they are separated from the other three.

I believe the old functionality hid the buttons if a filter was active because the buttons do not respect any active filter. I think that should be restored too if you want the buttons back. I don’t want people to think they will only delete the jobs which match a certain filter.

deve1212:
Sure @mperham  , will do the same .

deve1212:
> The buttons need to pull right so they are separated from the other three.
> 
> I believe the old functionality hid the buttons if a filter was active because the buttons do not respect any active filter. I think that should be restored too if you want the buttons back. I don’t want people to think they will only delete the jobs which match a certain filter.

@mperham Thanks for the feedback! I've made the requested changes and pushed them to the PR:

Visual separation: Added .bulk-retry-buttons.retry-all-button css to retry all button to pull all three buttons to right

Conditional hiding: To hide the buttons i have used js , here is a summary of the changes.
Classes (used in selectors and HTML):

retry-item-checkbox (for individual checkboxes).
check-all-retries (for the "check all" checkbox).
bulk-retry-buttons (for the buttons that get hidden/shown).

Functions:

toggleBulkRetryButtons (toggles button visibility: display: none if any checkbox checked, else display: block).
initializeBulkRetryToggle (sets up event listeners on checkboxes to trigger the toggle).
How it works: When a checkbox changes, toggleBulkRetryButtons checks if any are selected. If yes, it hides the .bulk-retry-buttons (via style.display = 'none'). If none are selected, it shows them (style.display = 'block').

Added Screen shot for better reference
Screenshot 2025-10-19 at 8 11 09 PM
Screenshot 2025-10-19 at 8 10 59 PM

web/assets/stylesheets/style.css
```diff
@@ -454,6 +454,10 @@ canvas + .table_container {
   gap: 8px;
 }
 
+.bulk-retry-buttons.retry-all-button {
+  margin-left: auto;
```
mperham: "left" and "right" does not work since the Sidekiq UI supports BiDi. You need to use start and end.

https://www.tutorialpedia.org/blog/bidtext-css/#4-best-practices

mperham:
Do we want similar functionality on the Dead or Scheduled tabs?

deve1212:
> Do we want similar functionality on the Dead or Scheduled tabs?

@mperham Sure we need it for these tabs too, will update the pr with these changes

web/assets/stylesheets/style.css
deve1212: updated the css with margin-inline-start: auto;

deve1212:
> Do we want similar functionality on the Dead or Scheduled tabs?

@mperham thanks for the input have added the following changes for this , please review it once and let me know-
- Added functionality for “Retry All” and “Delete All” actions in the Dead Jobs tab. 
- Added functionality for “Add All to Queue” and “Delete All” actions in the Scheduled Jobs tab.
- Introduced new APIs for bulk operations in scheduled jobs:
  POST /scheduled/all/add_to_queue
  POST /scheduled/all/delete
Both the Scheduled Jobs and Dead Jobs updates use the same JS function to toggle button visibility (similar to the previously added bulk retry functionality)

attaching images for better reference----
Screenshot 2025-10-20 at 1 13 58 PM
Screenshot 2025-10-20 at 1 14 05 PM

Screenshot 2025-10-20 at 1 11 50 PM
Screenshot 2025-10-20 at 1 11 45 PM

mperham:
Please add tests and the CSS needs to be refactored to remove redundancy.

lib/sidekiq/web/application.rb
```diff
@@ -318,6 +318,16 @@ def initialize(inst)
         redirect_with_query("#{root_path}scheduled")
       end
 
+      post '/scheduled/all/delete' do
+        Sidekiq::ScheduledSet.new.clear
+        redirect "#{root_path}scheduled"
+      end
+
+      post '/scheduled/all/add_to_queue' do
+        Sidekiq::ScheduledSet.new.retry_all
```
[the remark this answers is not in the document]
deve1212: fixed , introduced new method add_all_to_queue for scheduled jobs

deve1212:
> Please add tests and the CSS needs to be refactored to remove redundancy.

@mperham As suggested, I’ve removed redundant CSS by introducing common classes for all tabs. Also added RSpec tests for the scheduled#add_all_to_queue and scheduled#delete_all APIs. The changes have been tested successfully, and the full test suite passes.

mperham:
Almost there, nice job. 👍🏻

krasnoukhov:
Would be cool to see a release including this!

mperham:
Does main work like you want?

krasnoukhov:
@mperham Just tested the buttons, seem to work the same as on 7.x which is good

chaadow:
@mperham @deve1212 Hi! thanks for restoring those. However is it possible to add a confirm dialog when pressing `delete all`

I've just, intentionnally, pushed the button and it deleted everything instantly. However people sometimes can make mistakes, and the action is quite dangerous

Thanks in advance

mperham:
Open a new issue please.

chaadow:
Done

PR #6843, opened 2025-10-29
Assure that Sidekiq.server? and other similar methods return true/false

loomchild:
It looks like few environment-checking methods like `Sidekiq.server?` return `nil` instead of `false`.

I have the following line in Rails config:
```
config.eager_load = Sidekiq.server?
```

and Rails complains:
```
config.eager_load is set to nil. Please update your config/environments/*.rb files accordingly:

  * development - set it to false
  * test - set it to false (unless you use a tool that preloads your test environment)
  * production - set it to true
```

So I'd like to suggest a small improvement so this and other similar methods always return either `true` or `false`.

### DIFF
--- a/lib/sidekiq.rb
+++ b/lib/sidekiq.rb
@@ -48,7 +48,7 @@ module Sidekiq
   end
 
   def self.server?
-    defined?(Sidekiq::CLI)
+    !!defined?(Sidekiq::CLI)
   end
 
   def self.load_json(string)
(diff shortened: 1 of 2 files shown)


mperham:
It is idiomatic for boolean methods in Ruby to return "truthy" values, like nil. `if Sidekiq.server?` works perfectly fine, `if Sidekiq.server? == true` is not idiomatic Ruby code. In this way, Sidekiq nudges you to use proper Ruby idioms.

loomchild:
OK, thanks for explanation. You are right, due to this error message I got confused that methods ending with `?` should always return `true` or `false`.

I think it's a bit of a strange situation, since Rails assumes that `nil` means undefined and warns about that instead of quietly accepting it as a falsy value. I will fix it in my application.

Thanks a lot for a rapid response, and sorry for the noise.

PR #6846, opened 2025-10-30
Support Custom Logged Attributes

bschrag620:
Provides support for a custom log preparer to allow the sidekiq context to be customized so that all sidekiq logs can have added context when needed. See #6845.

### DIFF
--- a/lib/sidekiq/config.rb
+++ b/lib/sidekiq/config.rb
@@ -292,10 +292,6 @@ module Sidekiq
       @logger = logger
     end
 
-    def logged_job_attributes
-      @options[:logged_job_attributes]
-    end
-
     private def parameter_size(handler)
       target = handler.is_a?(Proc) ? handler : handler.method(:call)
       target.parameters.size
--- a/lib/sidekiq/job_logger.rb
+++ b/lib/sidekiq/job_logger.rb
@@ -30,7 +30,7 @@ module Sidekiq
         class: job_hash["wrapped"] || job_hash["class"]
       }
 
-      @config.logged_job_attributes.each do |attr|
+      @config[:logged_job_attributes].each do |attr|
         h[attr.to_sym] = job_hash[attr] if job_hash.has_key?(attr)
       end
 
--- a/test/job_logger_test.rb
+++ b/test/job_logger_test.rb
@@ -96,7 +96,7 @@ describe "Job logger" do
   end
 
   it "tests custom log attributes" do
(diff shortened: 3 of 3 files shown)


bschrag620:
Documentation updates still need to be made but wanted to toss this out their as a POC.

mperham:
Why? What are you trying to achieve? You can add context at any time in a server-side middleware which will wrap any job execution and the current context is added to all log output:

```ruby
class TraceMiddleware
  include Sidekiq::ServerMiddleware

  def call(instance, hash, queue)
    t = hash["trace_id"]
    Sidekiq::Context.add("trace_id", t) if t
    yield
  end
end
```

Is that what you want?

bschrag620:
> Why? What are you trying to achieve?

I would like to see our internal `trace_id=` show up in the sidekiq start/done/fail messages. If I'm reading things correctly, the middleware chain isn't called until after `sidekiq_logger.call`, which writes the initial start message. I Honestly hadn't thought about  setting Sidekiq::Context manually like this in the middleware. This solution would get us the trace_id on all messages except for `start`.

```
2025-10-30T15:24:33.022Z pid=23049 tid=89d class=MyJob jid=204d8f8b04c69f923bcbddbf INFO: start
2025-10-30T15:24:33.509Z pid=23049 tid=89d class=MyJob jid=204d8f8b04c69f923bcbddbf trace_id=9f78911d-6933-4a42-ae1a-6fb66e01c378 elapsed=0.487 INFO: done
```

mperham:
I think the direction I would support is an array of keys to add to the context, to genericize these lines:

https://github.com/sidekiq/sidekiq/blob/67187515db817c7fa77b680d81c8cc48f0fb35ae/lib/sidekiq/job_logger.rb#L32-L33

```
:logged_job_attributes => %w(bid tags)

  cfg.logged_job_attributes << "trace_id"
end
```

Does that give you better direction for a PR?

bschrag620:
> I think the direction I would support is an array of keys to add to the context, to genericize these lines:
> 
> https://github.com/sidekiq/sidekiq/blob/67187515db817c7fa77b680d81c8cc48f0fb35ae/lib/sidekiq/job_logger.rb#L32-L33
> 
> ```
> :logged_job_attributes => %w(bid tags)
> 
> Sidekiq.configure_server do |cfg|
>   cfg.logged_job_attributes << "trace_id"
> end
> ```
> 
> Does that give you better direction for a PR?

Absolutely! Will make some adjustments.

PR #6865, opened 2025-11-19
Add support for configuring Web UI static paths

stanhu:
We want to serve the Sidekiq Web UI assets from a CDN or another container. To do that, we copy the assets to another directory and set the backend to use this path:

```ruby
Sidekiq::Web.assets_path = '/path/to/my/assets'
```

### DIFF
--- a/lib/sidekiq/scheduled.rb
+++ b/lib/sidekiq/scheduled.rb
@@ -72,6 +72,7 @@ module Sidekiq
       include Sidekiq::Component
 
       INITIAL_WAIT = 10
+      attr_accessor :rnd
 
       def initialize(config)
         @config = config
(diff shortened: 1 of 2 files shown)


mperham:
Can you better explain why you want to do this? Assets use a Cache-Control header so they can be publicly cached by a CDN for one day. Manually distributing the assets to another location just adds complexity and something that can break without much advantage.

mperham:
I'm sorry, they use the `private` cache directive, meaning they can only be cached by the browser. Should we change this to `public`, afaik assets should not be user-specific.

stanhu:
@mperham As an optimization, we compile our assets in a single Webpack directory and then copy the assets to different containers. We have a proxy that serves static assets, but the Sidekiq gem isn't installed on this container. 

We started copying the Sidekiq assets into this Webpack bundle, but the `X-Sendfile` header causes the assets to be loaded from the Sidekiq gem directory. Being able to point the actual file location makes it possible to load the Sidekiq assets from the Webpack directory.

PR #6874, opened 2025-11-26
Add strict request validation using Sec-Fetch-Site 

deve1212:
issue - https://github.com/sidekiq/sidekiq/issues/6840

This PR introduces stricter request-safety handling by validating incoming requests based on HTTP method and the Sec-Fetch-Site header.

Following changes are introduced ---

- Added safe_methods? to allow only GET, HEAD, OPTIONS, and TRACE as inherently safe.
- Added safe_request? to allow unsafe HTTP methods only when HTTP_SEC_FETCH_SITE == "same-origin".
- Implemented deny(env) to log the blocked request and return a 403 Forbidden.

In config.rb

- disabling legacy CSRF behavior in favor of the new strict-mode checks.


mperham:
It looks like you've not made the changes necessary for POST actions in the test suite to run without CSRF.

deve1212:
> It looks like you've not made the changes necessary for POST actions in the test suite to run without CSRF.

let me fix that

deve1212:
> It looks like you've not made the changes necessary for POST actions in the test suite to run without CSRF.

changes done to support post requests in test suite , ran test suite successfully

PR #6897, opened 2025-12-30
GitHub Actions Services

frederikspang:


mperham:
Is this done? It's marked as a Draft.

frederikspang:
No - I just need to add the credentials for docker pull, to avoid regression with ratelimits for Redis.

Didn't wanna add it for all matrix builds to avoid sending credentials to docker.dragonfly (which is not on docker hub).

frederikspang:
Well that didn't work. Maybe from lack of secrets in PR's. 

Will Pick this back up when I can!

mperham:
Would moving to `services` mean we don't pull from docker as much? Does github provide a shared image cache here? We may not need to login?

frederikspang:
> Would moving to `services` mean we don't pull from docker as much? Does github provide a shared image cache here? We may not need to login?

To my knowledge it still counts towards rate limit.

frederikspang:
I think limitations of PR's and GitHub YAML may be stopping us here. Closing unless/until anyone has a greater knowledge than me :)

mperham:
See 7ec41b0e

frederikspang:
> See 7ec41b0e

@mperham That works - Just without Docker login and increased rate limits, right?

PR #6899, opened 2026-01-03
Add tui with Home tab implementation

Saidbek:
Adds a TUI for Sidekiq featuring a Home tab that displays statistics and redis information. The interface automatically refreshes data every 5 seconds.

Screenshot 2026-01-04 at 11 18 14 PM

https://github.com/sidekiq/sidekiq/issues/6898

### DIFF
--- a/.github/contributing.md
+++ b/.github/contributing.md
@@ -72,25 +72,32 @@ redis-server
 cd myapp/
 ```
 
-#### 8. Run required migration in order to launch Rails app
+#### 8. Install necessary gems for Rails app
+
+```
+bundle install
+```
+
+#### 9. Run required migration in order to launch Rails app
 
 ```
 rake db:migrate
 ```
 
-#### 9. Launch Rails app
+#### 10. Launch Rails app
 
 ```
 rails s
 ```
 
-#### 10. Create feature branch and start contributing!
+#### 11. Create feature branch and start contributing!
 
 ```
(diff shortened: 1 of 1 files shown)


bin/tui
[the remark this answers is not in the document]
mperham: I'm not sold on the history graph but the real time graph is a must.

mperham:
Those key/values would take a lot less space if you spread them across two lines, like:


```
 Key1          Key2       ...
Value1        Value2
```

bin/tui
```diff
@@ -13,12 +13,21 @@ end
 require "ratatui_ruby"
 require "sidekiq/api"
 
+# Suppress Sidekiq logger output to prevent interference with TUI rendering
+require "logger"
+Sidekiq.default_configuration.logger = Logger.new(IO::NULL)
```
Saidbek: This is to prevent sidekiq log msgs from appearing on the TUI screen
Image

Saidbek:
> Those key/values would take a lot less space if you spread them across two lines, like:
> 
> ```
>  Key1          Key2       ...
> Value1        Value2
> ```

@mperham Updated the stats and redis info sections to a two-line layout: keys on the first line, values on the second.

Screenshot 2026-01-04 at 11 18 14 PM

bin/tui
mperham: Another idea: move the stdout logging to stderr?

PR #6914, opened 2026-01-16
Polling in other TUI tabs besides Home

fpsvogel:
## Summary

The Home tab is currently the only one that polls for data every 5 seconds; the other tabs fetch data on every render.

This PR:

- Adds the same polling behavior to the other tabs.
- Changes the behavior of the chart in Home, so that the chart data is reset when navigating to another tab. This matches the web UI behavior, and prevents confusing omissions of time from appearing in the chart when navigating back to the Home tab.

| Queues (now updates only every 5s) | Home (now resets after navigating away) |
|---|---|
| Screen Cast 2026-01-15 at 11 27 44 PM | Screen Cast 2026-01-16 at 12 22 53 AM |

### DIFF
--- a/bin/tui
+++ b/bin/tui
@@ -60,20 +60,8 @@ module Sidekiq
       @current_tab = "Home"
       @selected_row_index = 0
       @base_style = nil
-      @data = nil
+      @data = {}
       @last_refresh = Time.now
-
-      stats = Sidekiq::Stats.new
-      @realtime_chart = {
-        previous_stats: {
-          processed: stats.processed,
-          failed: stats.failed
-        },
-        deltas: {
-          processed: Array.new(50, 0),
-          failed: Array.new(50, 0)
-        }
-      }
     end
 
     def run
@@ -212,6 +200,7 @@ module Sidekiq
       index_change = direction == :right ? 1 : -1
       @current_tab = TABS[(TABS.index(@current_tab) + index_change) % TABS.size]
       @selected_row_index = 0
+      @data = {}
       refresh_data
     end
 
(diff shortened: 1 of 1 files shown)


bin/tui
```diff
@@ -59,22 +59,9 @@ module Sidekiq
...
-        previous_stats: {
-          processed: stats.processed,
-          failed: stats.failed
-        },
-        deltas: {
-          processed: Array.new(50, 0),
-          failed: Array.new(50, 0)
-        }
-      }
     end
```
fpsvogel: Along with making the chart reset on tab navigation, I moved the chart data into `@data` (formerly `@home_data`). See [below](https://github.com/sidekiq/sidekiq/pull/6914/changes#diff-a8d716d2654a8e4934644cb6e0b7ca695cbda8d7d909964556c5d907981a2e6dR262-R271).

bin/tui
```diff
@@ -214,30 +200,31 @@ module Sidekiq
       index_change = direction == :right ? 1 : -1
       @current_tab = TABS[(TABS.index(@current_tab) + index_change) % TABS.size]
       @selected_row_index = 0
-      refresh_data if @current_tab == "Home"
+      @data = {}
+      refresh_data
     end
```
fpsvogel: Now that `@data` includes chart data, it must persist across polls (calls to `#refresh_data`), since the chart data is cumulative over time. So `@data` is now reset only on tab navigation.

fpsvogel: I like the idea of each tab having a specific `Data` definition. I can take a look next week.

PR #6917, opened 2026-01-26
Create tui directory

fpsvogel:
## Summary

Moves the contents of `bin/tui` to `lib/sidekiq/tui.rb`, except for the line `Sidekiq::TUI.new.run`. (These changes are easiest to review by commit: https://github.com/sidekiq/sidekiq/pull/6917/changes/8cf60527769dedbea8962d473f17102a44ca7a42 and https://github.com/sidekiq/sidekiq/pull/6917/changes/06f714eccc4719f1fb2733dfc9272de8990130e7.)

This file placement mirrors that of `lib/sidekiq/web.rb`.

This PR also adds `tui.log` to gitignore.

## Motivation

I've been thinking about how the TUI code can be organized so as to avoid a single large file. This PR makes the smallest possible step in that direction, both to simplify merge conflict resolution and also for me to get feedback on next steps.

In my mind, the next steps are to create a class for each tab, within a `TUI::Tabs` module. These classes would contain the behavior of each tab, including:

- Rendering the tab.
- Refreshing the tab's data.
- Declaration of controls that are unique to a tab. (Controls that are common across multiple tabs can either stay in `Sidekiq::TUI` or be moved to a separate module.)

This refactoring is one way to avoid the large `case @current_tab` statements for rendering and refreshing data. Rather than `Sidekiq::TUI` knowing everything about all tabs, it would only know where to find the tab classes (within a module `Sidekiq::TUI::Tabs`), and each tab class would contain its own behavior.

Feedback on this plan would be appreciated!

### DIFF
--- a/bin/tui
+++ b/bin/tui
@@ -1,0 +1,5 @@
+#!/usr/bin/env ruby
+
+require_relative "../lib/sidekiq/tui"
+
+Sidekiq::TUI.new.run
--- a/lib/sidekiq/tui.rb
+++ b/lib/sidekiq/tui.rb
@@ -1,5 +1,3 @@
-#!/usr/bin/env ruby
-
 require 'bundler/inline'
 
 gemfile do
@@ -946,5 +944,3 @@ module Sidekiq
     end
   end
 end
-
-Sidekiq::TUI.new.run


mperham:
Yes, we need to be able to break up code based on feature (tabs). I also imagine we'll want to discuss separation of concerns, i.e. the model (refresh_data), the views (render_*) and controller (the handle_input callbacks).

I've never built a real interactive TUI before so I'm treating it like any prototype: write the first version and be prepared to refactor it heavily as we figure out improved patterns and conventions over time. The only way to get to "perfect" is iteration and refactoring so this small step is great.

fpsvogel:
@mperham Thanks, I'll continue with as small steps as possible, then.

FYI RatatuiRuby plans to have [an app framework](https://sr.ht/~kerrick/ratatui_ruby/#chapter-5-the-framework) following an MVC pattern, but it is probably months away judging by [the roadmap](https://sr.ht/~kerrick/ratatui_ruby/#chapter-7-roadmap).

So, we can work toward a bespoke MVC separation, and then later if it makes sense, switch to RatatuiRuby's built-in framework.

Kerrick:
I shipped the beta of [Rooibos](https://www.rooibos.run) yesterday. It's the Model-View-Update piece of the RatatuiRuby ecosystem. It's separate from the framework on the roadmap, which definitely won't ship in the coming weeks. The framework is blocked[^1] on [codename "Kit,"](https://sr.ht/~kerrick/ratatui_ruby/#chapter-3-the-object-path--kit) which is an Ember-style[^2] component system. 

You may recognize Rooibos by its former codename, ratatui_ruby-tea. Even though it's not _the_ omakase framework, it's _a_ framework. It's lighter-weight and won't come with things like a local database. What it does have is an architecture that helps with separation of concerns.

Its MVU architecture separates the Model, the View, and the Update (not entirely unlike a Controller). It helps decompose larger apps into "fragments," in a [fractal](https://github.com/p69/dartea?tab=readme-ov-file#scaling-and-composition) way. So each tab could be its own fragment, and the app can [route messages](https://www.rooibos.run/docs/v0.6/examples/app_fractal_dashboard/dashboard/update_router_rb.html) to the appropriate tab based on the current state of the UI--such as which tab is active.

It lacks guides and examples, but if you've ever used React+Redux or [Elm](https://guide.elm-lang.org/architecture/), you'd find it familiar. I'd be glad to attempt a Rooibos conversion on a branch and open a PR, if that's a direction you'd like to go.

[^1]: Kit is only in the design phase; no code has been written for it yet. But Rooibos _should_ be useful today!
[^2]: Kit's architectural roots trace through Ember.js, to SproutCore, to Cocoa & Objective C, and even to Smalltalk if you squint hard enough.

mperham:
@Kerrick We definitely need some guidance/direction for code maintainability. I've been coding in "giant bag of code" mode for this TUI since I consider it a prototype and I'm still learning but we do need to start thinking about the final form of the codebase.

I've been less concerned about design/architecture patterns than reusable components with necessary features: one major example being a paginated Table which can smartly scale with terminal size, select multiple rows, take action on those rows, display error/flash messages. Having higher-level components on top of RR would be another win.

PRs always welcome.

PR #6918, opened 2026-01-26
update generator to consider config for rails.jobs_directory

evaniainbrooks:
https://github.com/sidekiq/sidekiq/discussions/6916

### DIFF
--- a/lib/generators/sidekiq/job_generator.rb
+++ b/lib/generators/sidekiq/job_generator.rb
@@ -1,13 +1,13 @@
 # frozen_string_literal: true
 
-require 'rails/generators/named_base'
+require "rails/generators/named_base"
 
 module Sidekiq
   module Generators # :nodoc:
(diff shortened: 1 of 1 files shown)


mperham:
I'm bikeshedding the config option name. Would `sidekiq_jobs_directory` be more explicit and less likely to conflict with Active Job?

evaniainbrooks:
Indeed. If it is in the rails config namespace, probably also this would need to be changed: https://github.com/rails/rails/blob/09c1c9fd4d5d68ba796c1b4950c8bf6f56d9a5a0/activejob/lib/rails/generators/job/job_generator.rb#L21

I moved it to `sidekiq` namespace from `rails`

mperham:
Document it. Can you add a comment which shows people how to set that option? ENV? Initializer? Other?

mperham:
Please don't reformat files, e.g. double to single quotes. We have a standard code format which is applied every time you run the test suite ("bundle exec rake") so all you need to do is run the tests and commit the restyle.

PR #6920, opened 2026-01-27
Update Pro-8.0-Upgrade.md with sidekiq batches migration instructions

freemanoid:
Added migration instructions for making 7.x batches compatible with 8.x

I didn't want to run sidekiq-pro 7.3 for a month before upgrading to sidekiq-pro 8.0.

So I've used this script to make redis "schema" compatible with 8.0


mperham:
I appreciate the work but this is not something I want to support.

freemanoid:
> I appreciate the work but this is not something I want to support.

No need to support it. The idea is to provide an option to perform an immediate redis migration instead of waiting for X days. It's just a note in the docs.

PR #6922, opened 2026-01-28
Run Sidekiq::CurrentAttributes in TransactionAwareClient

bguban:
Solves https://github.com/sidekiq/sidekiq/issues/6921

Writes the CurrentAttributes at the moment of calling perform instead of enqueuing. 

Here is the setup. 

```
Sidekiq.transactional_push!
Sidekiq::CurrentAttributes.persist('Current')
```

I need to run the job as a System user, that's why I temporarily substitute it in the current object.
```
Current.set(actor: SystemUser.fetch) do
  OrderExpirationJob.perform_at(@order.expires_at, @order.id)
end
```
before the fix the Current.actor is the whatever was set in the controller. After the fix, it's the system user.

### DIFF
--- a/lib/sidekiq/transaction_aware_client.rb
+++ b/lib/sidekiq/transaction_aware_client.rb
@@ -26,6 +26,11 @@ module Sidekiq
       # pre-allocate the JID so we can return it immediately and
       # save it to the database as part of the transaction.
       item["jid"] ||= SecureRandom.hex(12)
+
+      current_middleware.each do |entry|
+        entry.call(nil, item, nil, nil) { }
+      end
+
       @transaction_backend.call { @redis_client.push(item) }
       item["jid"]
     end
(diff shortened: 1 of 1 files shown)


bguban:
@mperham could you take a look? I don't like coupling between the CurrentAttributes middleware and the TransactionAwareClient but I can't find a better solution. If it looks good I'll add tests

bguban:
just a small workaround for those who have this problem:
```
Sidekiq::CurrentAttributes.persist('Current')
Sidekiq.transactional_push!

# This module calls the CurrentAttributes middleware at the moment push. This ensures that the job will store
# the CurrentAttributes values for that very moment but not the moment when the transaction is committed.
module TransactionalCurrentAttributes
  def push(item)
    save_middleware.call(nil, item, nil, nil) { }
    super
  end

  def save_middleware
    @save_middleware ||= @redis_client.middleware.find { |entry| entry.klass == Sidekiq::CurrentAttributes::Save }.make_new
  end
end
Sidekiq::TransactionAwareClient.prepend(TransactionalCurrentAttributes)
```

PR #6930, opened 2026-02-06
Guard against nil capsules in Sidekiq::Process#queues

ArunkumarRagu:
In Sidekiq 8.x, Sidekiq::Process#queues assumes capsules is always a Hash.
In mixed-version or legacy worker scenarios, capsules may return nil even
when the Redis "capsules" key exists, causing NoMethodError.

This change defensively falls back to legacy queue data when capsules is not
a valid Hash and adds a regression test to prevent crashes.

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -1134,6 +1134,8 @@ module Sidekiq
       if !self["capsules"]
         self["queues"]
       else
+        return self["queues"] unless capsules.is_a?(Hash)
+
         capsules.values.flat_map { |x| x["weights"].keys }.uniq
       end
     end


mperham:
I’d like to understand why capsules could ever be nil. It should either be set or not exist. Can you reproduce the scenario where it is nil?

ArunkumarRagu:
> I’d like to understand why capsules could ever be nil. It should either be set or not exist. Can you reproduce the scenario where it is nil?

I encountered this in a production environment while loading the /busy page in the Sidekiq dashboard.

/busy iterates over all registered Sidekiq::Process entries and calls Process#queues for each one. During rolling upgrades or worker restarts, Redis can contain mixed or transitional process metadata. In this state, the raw Redis "capsules" key may be present while Process#capsules returns nil because the structure cannot be built.

This results in self["capsules"] being truthy but capsules being nil, causing capsules.values to raise and the /busy page to fail entirely.

The added guard ensures /busy remains safe by falling back to legacy queue data when capsules is not a valid Hash, which keeps the dashboard functional during normal operational transitions.

mperham:
> During rolling upgrades or worker restarts, Redis can contain mixed or transitional process metadata.

That's what I'm trying to understand. If we have malformed persisted process data, I want to understand how it got in that state. The "capsules" value should never be nil. The key should not exist if there is no value.

ArunkumarRagu:
> > During rolling upgrades or worker restarts, Redis can contain mixed or transitional process metadata.
> 
> That's what I'm trying to understand. If we have malformed persisted process data, I want to understand how it got in that state. The "capsules" value should never be nil. The key should not exist if there is no value.

The Redis data itself is not malformed and does not contain a nil value.

self["capsules"] reflects raw process metadata persisted in Redis, while Process#capsules is a derived structure that interprets that data and builds a Ruby hash with an expected shape. During rolling upgrades or worker restarts, Redis can legitimately contain process entries written by different Sidekiq versions. In these cases, the "capsules" key may be present, but its contents do not match what the current version expects, so Process#capsules returns nil to indicate it cannot build a valid structure.

This results in self["capsules"] being truthy while capsules is nil, even though the Redis data itself is not nil. The failure only surfaces in introspection paths like /busy, which iterate over all processes and call Process#queues.

The guard ensures Process#queues remains safe by falling back to legacy queue data when capsules cannot be interpreted, without changing behavior when valid capsule data is present.

mperham:
Sidekiq does not support multiple different versions running simultaneously and never has.

Issue #6925, opened 2026-02-02
Decouple `Sidekiq::Testing.fake!` from `require "sidekiq/testing"` to prevent accidental production outages


golive:
Hello Sidekiq team,

I would like to open a discussion regarding the current behavior of `Sidekiq::Testing`.

### Context
Currently, requiring `"sidekiq/testing"` automatically enables `Sidekiq::Testing.fake!` mode by default ([line 81](https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/testing.rb#L81)):

While there is [a warning](https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/testing.rb#L333) that prints to stderr if the environment is not testing, this behaviour recently caused a significant production incident for my company.

### The Problem
We deployed a new release that included a one-off maintenance script, which was in turn autoloaded in the application run. This script required `sidekiq/testing` to be able to use the `inline` feature. Because the require automatically triggers `Sidekiq::Testing.fake!`, our production application stopped enqueuing jobs immediately.

The jobs were being pushed to the in-memory array instead of Redis, and since it was production, they were essentially lost upon process restart/exit.

While the warning exists, it is very easy to miss in high-volume production logs or during a deployment rush. We believe that a library's import side-effects should not drastically alter the runtime behavior of the core application in such a dangerous way.

### Proposed Solution
I propose we deprecate the automatic enabling of fake mode when the file is required.

**Option A:** Breaking Change. Remove `Sidekiq::Testing.fake!` from `lib/sidekiq/testing.rb`. Users must explicitly call `Sidekiq::Testing.fake!` or `Sidekiq::Testing.inline!` in their test setup.

**Option B:** Non-Breaking Transition Introduce an environment variable or configuration flag that prevents the auto-execution, eventually making it the default.

```ruby
# Potential change in lib/sidekiq/testing.rb
unless ENV['SIDEKIQ_TESTING_DISABLE_AUTO_FAKE']
  warn "DEPRECATION WARNING: Requiring 'sidekiq/test' will no longer automatically enable Sidekiq::Testing.fake! in Sidekiq XX. Please add `Sidekiq::Testing.fake!` to your test suite setup manually."
  Sidekiq::Testing.fake!
end
````

### Why change this?
Safety: requiring should generally be side-effect free regarding global state, especially state that disables core functionality (persistence).

[...]

PR #6931, opened 2026-02-09
New Testing API

mperham:
Associated with #6925

We want to stop the user from requiring "sidekiq/testing" as a means to implicitly but globally change the mode of operation. Instead we'll move to a more explicit API:

## Previous API

```ruby
require "sidekiq/testing"

Sidekiq::Testing.fake!
Sidekiq::Testing.fake! do
  # run tests
end
```

## New APIs

```ruby
# no need to require anything
Sidekiq.testing!(:fake)
Sidekiq.testing!(:fake) do
  # run tests in fake mode, 
end

# only works after calling `Sidekiq.testing!`
Sidekiq::Testing.fake!
Sidekiq::Testing.fake! do
  # run tests
end
```

Any use of `testing!` in the production environment will raise an error.


golive (commented):
Thank you for the new testing API proposal, it looks good!

adamniedzielski:
I arrived here after I received `⛔️ require "sidekiq/testing" is deprecated and will be removed in Sidekiq 9.0. See https://sidekiq.org/wiki/Testing#new-api` during a test run.

Unfortunately https://sidekiq.org/wiki/Testing#new-api is a broken link - there's no section about the new API there. https://github.com/sidekiq/sidekiq/wiki/Testing already contains a section about the new API, so it's some synchronization issue. Still, worth fixing, otherwise people get confused.

mperham:
@adamniedzielski Yeah, it's manually sync'd right now. I'm preparing for the day we have to move off GitHub as the organization seems increasingly dysfunctional.

dentarg:
FYI if your test helper only contained `require "sidekiq/testing"` this became a breaking change:

```ruby
$ b e rake
/Users/dentarg/.arm64_rubies/3.4.8/lib/ruby/gems/3.4.0/gems/sidekiq-8.1.3/lib/sidekiq/testing.rb:1:in '<top (required)>': uninitialized constant Sidekiq (NameError)

Sidekiq.testing!(:fake)
^^^^^^^
	from /Users/dentarg/.arm64_rubies/3.4.8/lib/ruby/3.4.0/bundled_gems.rb:82:in 'Kernel.require'
	from /Users/dentarg/.arm64_rubies/3.4.8/lib/ruby/3.4.0/bundled_gems.rb:82:in 'block (2 levels) in Kernel#replace_require'
```

PR #6937, opened 2026-02-20
Support profiling ActiveJob jobs

odlp:
Hello 👋 ,

This PR proposes a few minor changes to facilitate profiling ActiveJob jobs. As the [wiki page on Profiling](https://github.com/sidekiq/sidekiq/wiki/Profiling#active-job) states this is currently quite awkward because: 

> Active Job does not support adding arbitrary key/values to the job payload via set, making it impossible to dynamically enable profiling in your Active Jobs, e.g. `SomeJob.set(profile: "sometoken").perform_later(...)` will not work because `ActiveJob::Core#set` will discard the profile element.

This PR proposes updating the AJ Sidekiq adapter to check if a job instance `responds_to?(:profile)`. If it does then the profile option is set on the base Sidekiq job hash.

To profile ActiveJobs the end-user would still need a minor change to their application, for example:

```ruby
# app/jobs/concerns/profileable_job.rb
module ProfileableJob
  attr_accessor :profile

  def set(options = {})
    self.profile = options[:profile] || options["profile"]
    super
  end
end

# app/jobs/application_job.rb
class ApplicationJob < ActiveJob::Base
  include ProfileableJob
end

# Usage:
MyActiveJob.set(profile: "oli").perform_later
```

Or they could specify a `#profile` method on individual jobs if preferred.

I hope this suggestion is simple enough it could be added to the wiki, and provide support for profiling ActiveJobs whilst avoiding the need for Sidekiq as a library to patch `ActiveJob::Core#set` to support it.

## Notes

I've initially left out support for the profile option in `ActiveJob::QueueAdapters::SidekiqAdapter#enqueue_all` because bulk enqueueing jobs to be profiled feels contrary to the wiki advice to profile sparingly:

> Be careful not to profile every job and quickly fill up your Redis! Profiling data expires after 24 hours.

## Preview

Testing this branch locally I can see the profile in Sidekiq Web as expected:

sidekiq-activejob-profile

### DIFF
--- a/lib/active_job/queue_adapters/sidekiq_adapter.rb
+++ b/lib/active_job/queue_adapters/sidekiq_adapter.rb
@@ -63,19 +63,19 @@ begin
         def enqueue(job)
           # NB: Active Job only serializes keys it recognizes. We
           # cannot set arbitrary key/values here.
-          wrapper = Sidekiq::ActiveJob::Wrapper.set(
-            wrapped: job.class,
-            queue: job.queue_name
-          )
+          options = {wrapped: job.class, queue: job.queue_name}
+          options[:profile] = job.profile if job.respond_to?(:profile) && !job.profile.nil?
+
+          wrapper = Sidekiq::ActiveJob::Wrapper.set(options)
           job.provider_job_id = wrapper.perform_async(job.serialize)
         end
 
         # @api private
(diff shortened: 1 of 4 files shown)


mperham:
Excellent work!

PR #6955, opened 2026-03-18
fix: guard against nil workset when page offset exceeds collection size

Saidbek:
When the requested page exceeds available workers, `@workset` was set to `nil`, crashing the busy page. Guard `page_items` result with `|| []`.


mperham:
Thanks for the report, quite an edge case!

PR #6960, opened 2026-03-20
Fix pagination active page visibility in dark mode

### LINKED ISSUE #6929: Pagination active page is hard to distinguish in Dark Mode on Sidekiq Dashboard
When viewing the Sidekiq Web UI in Dark Mode, the pagination component makes it difficult to identify the currently active page.
The visual difference between the active page and inactive pages is very subtle, which hurts usability.

### Dark Mode
Image

### Light Mode
Image

Lang is ja.

sondh5:
## Summary

Fixes #6929

The pagination component makes it impossible to distinguish the currently active page from other pages. All page numbers share the same styling, the current page uses `class="disabled"` which only reduces opacity, making it look like a disabled nav arrow rather than the active page indicator.

**Changes:**
- Change current page `<li>` from `class="disabled"` to `class="active"` so it's semantically correct
- Add `.pagination .active a` styling with `color: var(--color-primary)` and `font-weight: 700`, uses existing CSS variables so it works in both light and dark mode

### Before
before-dark-final

before-light-final

### After
after-dark-final

after-light-final

## Test plan

- [x] Verify active page is clearly visible in dark mode (bold + primary color)
- [x] Verify active page is clearly visible in light mode
- [x] Verify disabled arrows (`«`/`»`) still appear dimmed on first/last page
- [x] Verify page navigation still works correctly

### DIFF
--- a/web/assets/stylesheets/style.css
+++ b/web/assets/stylesheets/style.css
@@ -534,6 +534,8 @@ input[type="checkbox"] { accent-color: var(--color-primary); }
 
 .pagination a { text-decoration: none; }
 
+.pagination .active a { color: var(--color-primary); font-weight: 700; }
+
 .pagination .disabled { opacity: 0.3; }
 
 div:has(.pagination.pull-right) { margin-left: auto; }
--- a/web/views/_paging.html.erb
+++ b/web/views/_paging.html.erb
@@ -10,7 +10,7 @@
           <a href="<%= url %>?<%= qparams(page: @current_page - 1) %>"><%= @current_page - 1 %></a>
         </li>
       <% end %>
-      <li class="disabled">
+      <li class="active">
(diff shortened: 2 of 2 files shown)


mperham:
Why are you merging this into the 6.x branch?

sondh5:
@mperham Apologies. I targeted 6-x because I'm on v6.5.12 and hit the same issue as #6929 with the bootstrap css. Should I retarget to main, or would you consider patching 6-x as well?

afurm:
The active state style hardcodes `color: #fff` on top of `background-color: #b1003e`, but this is applied globally across all pages using this partial. Is the active page color intentionally independent of the surrounding theme, or should it respect a contextual text color variable for consistency with the rest of the UI?

mperham:
Please fix main. 6.x is out of support and will not get this fix.

mperham:
Thank you!

PR #6966, opened 2026-03-31
Messing with flavors, follow up to #6715

mperham:
WIP, more detail to follow

### DIFF
--- a/lib/sidekiq/flavor.rb
+++ b/lib/sidekiq/flavor.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+module Sidekiq
+  ##
+  # Flavor is the term for a custom job-argument
+  # serialization scheme. :default is Sidekiq's long-standing
+  # default argument serialization scheme which only supports pure JSON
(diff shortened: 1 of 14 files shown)


ryansch:
That branch name: ✨

lib/sidekiq/flavor.rb
```diff
@@ -0,0 +1,80 @@
...
+    # You can default a flavor in your job or a base class:
+    #
+    #   class MyJob
+    #     sidekiq_options flavor: :marshal
+    #
+    def add(flvr)
+      @known[flvr.name.to_s] = flvr
+    end
+
+    def rb_to_wire(job)
```
ryansch, line 41: I like the naming scheme here.

mperham:
It was a fun experiment but it feels like the outcome is the same: this whole abstraction adds a layer of complexity that 99% of Sidekiq users aren't asking for. @ryansch I know you really want this feature but allowing users to change payload serialization means potentially breaking features which need to work with a well-known format within Redis Lua scripts, for example.

ryansch:
@mperham I understand. I'll figure something out.

PR #6973, opened 2026-04-10
Update 8.0 Upgrade docs with error_handler breaking change

javierjulio:
When upgrading, this was a problematic one to figure out. We encountered it as we recently switched from Bugsnag to Datadog for error reporting but the latter is a downgrade when it comes to Sidekiq integration for errors so we had to add our own error handler. I figure adding one isn't too common.

Resources: 
- https://github.com/sidekiq/sidekiq/blob/v8.1.2/lib/sidekiq/config.rb#L311
- https://github.com/sidekiq/sidekiq/pull/6051


mperham:
Thanks, I see the issue. Sorry for the trouble!

javierjulio:
No worries at all! Normally we would catch this as we upgrade regularly and would check release notes where it was included. We just went off the upgrade docs solely. I'm happy to finally contribute. Thank you!

PR #6976, opened 2026-04-16
add: info tag when dead queue is at max

### LINKED ISSUE #6975: Information tag when Dead Queue is filled up
For new users and the dead queue gets filled up for some reason it would be nice to know as a user that the max dead queue is 9999 and subsequent dead requests get replaced and lost.

Open to suggestions but also understand if this isn't needed.

Yegeun:
Closes #6975

### DIFF
--- a/web/locales/en.yml
+++ b/web/locales/en.yml
@@ -17,6 +17,7 @@ en:
   Dashboard: Dashboard
   Dead: Dead
   DeadJobs: Dead Jobs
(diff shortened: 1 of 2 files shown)


mperham:
I'm going to roll this back. I think there are plenty of usecases where having a full morgue is fine and acceptable, for instance Mastodon admins constantly have a full dead queue due to 3rd party Mastodon servers going offline. Showing an error message all the time isn't useful.

PR #6978, opened 2026-04-21
Show iterable job state in web UI

joshuaylee:
Displays cursor, executions, runtime for Iterable jobs on the busy page, queue/retries list, and job detail pages (retry/scheduled/dead).

Iterates on #6787. Batches the calls to Redis.

Busy list:
Screenshot 2026-04-21 at 11 17 32 AM

Queue/Retries list:
Screenshot 2026-04-21 at 11 27 42 AM

Job Detail:
Screenshot 2026-04-21 at 11 36 26 AM

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -478,6 +478,12 @@ module Sidekiq
       self["jid"]
     end
 
+    def iterable_state
+      return @iterable_state if defined?(@iterable_state)
+
+      @iterable_state = Sidekiq::IterableJobState.fetch(jid)
+    end
+
     def bid
       self["bid"]
     end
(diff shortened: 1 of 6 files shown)


lib/sidekiq/api.rb
```diff
@@ -393,6 +393,42 @@ def as_json(options = nil)
   class JobRecord
     include ApiUtils
 
+    # Batch-load iteration state for multiple jobs in a single Redis pipeline.
+    # Assigns each job's iterable_state so subsequent calls avoid Redis.
+    def self.load_iterable_states(jobs)
```
mperham: I'm wondering why this class method needs to be in JobRecord at all. I think this would be better placed in `sidekiq/web/helpers.rb`. We can also add a method to S::JobRecord which will lazy load an individual job's state while the helper method optimizes the bulk load. The helper should take a set of JIDs and return a Hash<JID, State>.

joshuaylee: Good point about JobRecord. I looked at web/helpers.rb but it's all presentation/formatting — no Redis data fetching there, so it'd be the odd one out.

One more idea: what do you think of a small standalone class in api.rb (e.g. `Sidekiq::JobIterableState`) with .bulk_fetch(jids) for bulk and .for(jid) for single lookups. That keeps data fetching with the rest of the data API, and JobRecord stays untouched. Route handlers and templates would call it directly.

Does that direction work for you?

mperham: That sounds reasonable. I think Sidekiq::IterableJob already exists so I would use that as a basis for the name: `Sidekiq::IterableJob::State` or similar. I just don't like utility class methods so I usually think a little bit harder about a proper object oriented home for the method.

joshuaylee: Thanks for the direction. I went with `Sidekiq::IterableJobState` as a flat class in api rather than nesting under Sidekiq::IterableJob since other API classes (JobRecord, WorkSet, ProcessSet, etc.) are all flat under Sidekiq::

web/views/_job_info.html.erb
```diff
@@ -34,6 +34,14 @@
             <code><%= job.jid %></code>
           </td>
         </tr>
+        <% if (state = Sidekiq::IterableJobState.fetch(job.jid)) %>
```
mperham: This should use a method like `Sidekiq::JobRecord#iteration_state`. The job object is right at hand.

web/views/queue.html.erb
mperham, line 1: Queues represent jobs which are not running currently but will run shortly. I don't think it's necessary to display state on this page.

joshuaylee: I was thinking that when an iterable job is interrupted, it can be helpful to see how far it got. Are you thinking that's rare enough (and only available for a short time before job runs) that it's not worth showing?

mperham: It is reenqueued but queue draining is typically so fast that you won’t see it. I don’t see this as useful for most. On Busy and Retries, sure.

joshuaylee: Alright, removed from the Queue page.

mperham:
Well done, thanks for working with me to polish it!

PR #6986, opened 2026-05-06
Identify Sidekiq connections via CLIENT SETINFO

vchomakov:
## What

Bumps the `redis-client` floor to `>= 0.27.0` and uses its new `driver_info` config to identify Sidekiq on every Redis connection via `CLIENT SETINFO`.

## Why

Connections from Sidekiq, the host app's `redis` gem, Rails cache, etc. are currently indistinguishable in `CLIENT LIST` / `CLIENT INFO`.
`redis-client` 0.27.0 (released Mar 2026) added `driver_info:` to set `LIB-NAME` automatically during the connection prelude.

After this change, ops will see:

    lib-name=redis-client(sidekiq_v8.1.3) lib-ver=0.28.0

per Sidekiq connection, matching the `<driver>_v<version>` convention used accross the econsystem. 

Stacking is supported — users can override:

```ruby
config.redis = { driver_info: "sidekiq_v#{Sidekiq::VERSION};myapp_v1.0" }

### DIFF
--- a/lib/sidekiq/redis_client_adapter.rb
+++ b/lib/sidekiq/redis_client_adapter.rb
@@ -107,6 +107,10 @@ module Sidekiq
       # on by default.
       opts[:reconnect_attempts] ||= 1
 
+      # Identify ourselves to Redis via CLIENT SETINFO so connections
+      # are distinguishable in CLIENT LIST / CLIENT INFO output.
+      opts[:driver_info] ||= "sidekiq_v#{Sidekiq::VERSION}"
+
       opts
     end
   end
--- a/sidekiq.gemspec
+++ b/sidekiq.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |gem|
     "rubygems_mfa_required" => "true"
   }
 
-  gem.add_dependency "redis-client", ">= 0.26.0"
+  gem.add_dependency "redis-client", ">= 0.29.0"
   gem.add_dependency "connection_pool", ">= 3.0.0"
   gem.add_dependency "rack", ">= 3.2.0"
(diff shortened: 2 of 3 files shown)


mperham:
Would it be beneficial to include the value of `config[:tag]` if it is set? That's the `--tag` option which allows the user to identify the app and would be useful if multiple apps are using the same Redis (e.g. using different DBs).

mperham:
This requires Redis 7.2 and Sidekiq still supports Redis 7.0. It will need to be conditional or wait for Sidekiq 9.0.

vchomakov:
> Would it be beneficial to include the value of `config[:tag]` if it is set? That's the `--tag` option which allows the user to identify the app and would be useful if multiple apps are using the same Redis (e.g. using different DBs).

Hi @mperham and thanks for the review!

For consistency, we're strictly following the format mandated by the `CLIENT SETINFO` command here: https://redis.io/docs/latest/commands/client-setinfo/

> We recommend using the following format for the custom suffixes for third-party libraries (?<custom-name>[ -~]+)[ -~]v(?<custom-version>[\d\.]+) and use ; to delimit multiple suffixes.

vchomakov:
> This requires Redis 7.2 and Sidekiq still supports Redis 7.0. It will need to be conditional or wait for Sidekiq 9.0.

Good catch — that was an upstream bug in `redis-client` 0.27/0.28 (silent-rescue path closed the socket, breaking Redis < 7.2). Fixed in 0.29.0 ("Fix connecting to Redis 7.1 and older with driver_info: set").

I bumped the floor to >= 0.29.0 and verified end-to-end against Redis 7.0.15 and 7.4.

PR #6995, opened 2026-05-22
Fix flaky queue_summaries latency test

Saidbek:
The `queue_summaries` test asserts latency within a 10ms window, which is tight enough that slow CI runners regularly trip it (e.g. `Expected |0.05 - 0.076| (0.026) to be <= 0.01`). This switches the assertions to lower-bound `assert_operator` checks, matching the pattern used a few lines below in the `enqueued` test.

### DIFF
--- a/test/api_test.rb
+++ b/test/api_test.rb
@@ -157,12 +157,12 @@ describe "API" do
 
         assert_equal "bar", result[0].name
         assert_equal 3, result[0].size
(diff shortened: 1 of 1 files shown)


Saidbek:
This test has been [failing](https://github.com/sidekiq/sidekiq/actions/runs/26315144208/job/77472614098) intermittently on #6994

Issue #6950, opened 2026-03-16
Remove YAML


mperham:
YAML is a complex beast, and can be a source of security issues. Don't load it if we don't actually need it.

1. Parsing `sidekiq.yml` if that file is detected.
2. Translations in `web/locales`

Can we switch to JSON-formatted locale files?

PR #7001, opened 2026-05-23
Remove parse_yaml_old from web helpers

hammadxcm:
## Summary

Commit 25afb793 (#6950) replaced YAML-based locale parsing with a custom parser to keep YAML off the web request path, but left `parse_yaml_old` behind — marked `# TODO Remove` and used only as a test oracle. It still pulled `require "yaml"` + `YAML.safe_load_file` into `web/helpers.rb`.

- Remove `parse_yaml_old` (and its `# TODO Remove`); `parse_yaml_new` is already the sole production parser via `strings`.
- The "can parse locale files" test keeps its exact oracle cross-check by calling `YAML.safe_load_file(path)` directly (tests may use YAML — the security boundary is production web code). Also drops two dead debug comments.

`git grep parse_yaml_old` is now empty.

## Testing

### DIFF
--- a/lib/sidekiq/web/helpers.rb
+++ b/lib/sidekiq/web/helpers.rb
@@ -72,13 +72,13 @@ module Sidekiq
       # so extensions can be localized
       @@strings[lang] ||= config.locales.each_with_object({}) do |path, global|
         find_locale_files(lang).each do |file|
-          strs = parse_yaml_new(file)
+          strs = parse_yaml(file)
           global.merge!(strs[lang])
         end
       end
     end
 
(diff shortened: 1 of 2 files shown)


mperham:
The old and new methods were there so tests could verify they parsed identically. If you’re removing the old method, please rename the new method.

hammadxcm:
Good call — renamed `parse_yaml_new` to `parse_yaml` now that the old method is gone. The locale test keeps the identical cross-check by loading the YAML oracle directly with `YAML.safe_load_file`. Pushed.

Issue #4717, opened 2020-10-14
Include process RSS on Busy page


mperham:
Seeing the process RSS can be really helpful in determining at a glance if a Sidekiq process is "healthy" or not. If RSS is high, one click on the Stop button can fix it.

While nice, things like sparklines or historical data or graphs are not in scope. This issue does not aim to compete with NewRelic or DataDog.

mperham:
<img width="835" alt="Screen Shot 2020-10-14 at 2 52 52 PM" src="https://user-images.githubusercontent.com/2911/96049903-94575280-0e2d-11eb-9a0e-bb4da127d361.png">

TALlama:
We use a multi-tenant architecture where each tenant is run as its own Rails app, on shared app and db servers. The sidekiq workers follow that same pattern: each tenant gets its own worker on the shared box, and its own queues, but they all share a Redis db so that we can monitor them all from one place.

Today we got an alert that the worker box was low on memory, and we used the Sidekiq Web UI to quickly identify which tenant was having the problem, allowing us to easily find the offending job. 

(the underlying bug, for the curious: the bulk download created by the job in a question was doing a FileMagic scan to identify the file type of the ZIP file it had just created, which apparently involves FileMagic reading the entire 14GB file into memory. Needless to say this didn’t work out; we made that job skip the scan and assume that our zip creation code does indeed create zip files)

PR #7011, opened 2026-05-28
Use the requested pid in the Linux memory grabber

hammadxcm:
## Summary

The Linux branch of `MEMORY_GRABBER` (`lib/sidekiq/launcher.rb`) takes a `pid` argument but ignores it, reading the current process instead:

```ruby
when /linux/
  ->(pid) {
    IO.readlines("/proc/#{$$}/status").each do |line|   # reads $$ , not pid
      ...
```

The `darwin|bsd` branch correctly uses `pid` (`ps -o pid,rss -p #{pid}`). The bug has been latent since #4717 because the only caller is `memory_usage(::Process.pid)`, so `$$` and `pid` are always equal in practice — but the lambda is parameterized by `pid` and should honor it (and stay consistent across platforms).

Fix: read `/proc/#{pid}/status`.

## Testing

Adds a Linux-guarded regression test: calling the grabber with a non-existent pid must raise `Errno::ENOENT` (it now reads *that* pid's `/proc` entry). This fails against the old `$$` behavior (which silently returns the current process's RSS) and passes after the fix. It skips on non-Linux platforms where this lambda isn't used.

`bundle exec rake` is green (standard + lint:herb + full suite; the new test skips locally on macOS and runs on Linux CI).

> Note: I develop on macOS, so the Linux `/proc` path can't execute locally — this fix is by inspection/consistency with the `darwin|bsd` branch, and the regression test exercises it on Linux CI.

### DIFF
--- a/lib/sidekiq/launcher.rb
+++ b/lib/sidekiq/launcher.rb
@@ -235,7 +235,7 @@ module Sidekiq
     MEMORY_GRABBER = case RUBY_PLATFORM
     when /linux/
       ->(pid) {
-        IO.readlines("/proc/#{$$}/status").each do |line|
+        IO.readlines("/proc/#{pid}/status").each do |line|
           next unless line.start_with?("VmRSS:")
           break line.split[1].to_i
         end
--- a/test/launcher_test.rb
+++ b/test/launcher_test.rb
@@ -16,6 +16,15 @@ describe Sidekiq::Launcher do
       refute_nil kb
       assert kb > 0
     end
+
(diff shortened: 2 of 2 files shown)


hammadxcm:
Follow-up: opened #7012 to broaden test coverage for this code — it adds a cross-platform check that `MEMORY_GRABBER` reads a *different* process by pid (`Process.ppid`, exercising exactly what this fix enables, including on macOS where the prior `/proc` test skips) and a direct test for the `memory_usage` wrapper.

PR #7019, opened 2026-05-30
Make Active Job adapter compatible with next version of Rails

bdewater-thatch:
See https://github.com/rails/rails/pull/57472


mperham:
Looks like maybe you didn’t run the tests?

PR #7022, opened 2026-06-02
Honor the pool kwarg in Sidekiq::Stats::History

hammadxcm:
## Summary

`Sidekiq::Stats::History.new(days, start_date = nil, pool: nil)` has accepted a `pool:` keyword since #5487 (the Sidekiq::Capsule introduction in 2022), but the kwarg was never assigned or used — `date_stat_hash` always reads from `Sidekiq.redis` regardless. Callers who passed a custom pool got their argument silently dropped.

Assign `@pool` in `initialize` and route `date_stat_hash` through a small `with_redis` helper that uses the passed pool when present and falls back to `Sidekiq.redis` otherwise. This matches the pool-arg precedent already in `Sidekiq::Deploy` (which takes a positional pool and uses `@pool.with`).

Non-breaking: existing callers (no `pool:` argument) continue to use `Sidekiq.redis`.

## Testing

Wraps `@cfg.redis_pool` in a small tracking decorator, passes it as `pool:` to `Stats::History`, and asserts the decorator's `.with` is invoked and the right hash comes back. `bundle exec rake` is green (standard + lint:herb + full suite).

### DIFF
--- a/lib/sidekiq/api.rb
+++ b/lib/sidekiq/api.rb
@@ -239,6 +239,7 @@ module Sidekiq
         raise ArgumentError if days_previous < 1 || days_previous > (5 * 365)
         @days_previous = days_previous
         @start_date = start_date || Time.now.utc.to_date
+        @pool = pool
       end
 
       def processed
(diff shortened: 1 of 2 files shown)


mperham:
I believe I changed my mind here and never implemented pool: on purpose. The arg needs to be removed.

hammadxcm:
Sure.

PR #7025, opened 2026-06-02
Add tests for Sidekiq::Component utility helpers

hammadxcm:
## Summary

`Sidekiq::Component` is mixed into nearly every server-side class, but only `handle_exception` had direct coverage. Add `test/component_test.rb` (test-only, no `lib/` changes) covering the pure helpers:

- **`#real_ms` / `#mono_ms`** — return integer milliseconds (wall-clock and monotonic respectively).
- **`#tid`** — base-36 string, memoized per thread, independent across threads.
- **`#hostname`** — prefers `ENV["DYNO"]`, falls back to `Socket.gethostname` when unset (with proper save/restore).
- **`#default_tag`** — pins the four branches: returns the directory basename, except when it sits inside a Capistrano `releases` directory (then returns the app name above), and handles the numeric/non-numeric basename × matching/non-matching parent cases.

## Testing

### DIFF
--- a/test/component_test.rb
+++ b/test/component_test.rb
@@ -1,0 +1,84 @@
+# frozen_string_literal: true
+
+require_relative "helper"
+require "sidekiq/component"
+require "socket"
+
+class ComponentHarness
+  include Sidekiq::Component
+end
+
+describe Sidekiq::Component do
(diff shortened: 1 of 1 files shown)


mperham:
Too specific, unnecessary.

PR #7029, opened 2026-06-04
Add tests for JobRecord error_backtrace decompression and error?

hammadxcm:
## Summary

Four cohesive tests around the error-information API on `JobRecord` / `SortedEntry` — the contract the Web UI uses to render the retry / dead job error pages. Test-only, no `lib/` changes.

- **`error_backtrace` round-trip decompression** — using the real `JobRetry#compress_backtrace` to produce the stored payload, `JobRecord#error_backtrace` decompresses it back into the original array of backtrace lines.
- **`error_backtrace` nil memoization** — when no backtrace was stored, the result is `nil`, and the cached `@error_backtrace` is respected (so it's not re-decoded on every call).
- **`SortedEntry#error?`** — reflects the presence of `error_class` in the payload (true when set, false when absent).
- **`SortedEntry` inheritance** — confirms the `JobRecord#error_backtrace` decompression behavior is inherited through to `SortedEntry`.

## Testing

### DIFF
--- a/test/api_test.rb
+++ b/test/api_test.rb
@@ -2,6 +2,7 @@
 
 require_relative "helper"
 require "sidekiq/api"
+require "sidekiq/job_retry"
 require "active_job"
 require "action_mailer"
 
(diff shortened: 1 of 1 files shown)


test/api_test.rb
```diff
@@ -2,6 +2,8 @@
 
 require_relative "helper"
 require "sidekiq/api"
+require "sidekiq/capsule"
```
mperham: I've adjusted the JobRetry object to only require a `config` parameter so you don't need to import and create a Capsule for testing JobRetry. You can just pass in the standard `@config = reset!` object to keep things simple.

mperham:
Please update tests based on my recent API refactor.

hammadxcm:
Rebased onto main and updated to use the new `JobRetry.new(@cfg)` signature — no longer importing or constructing a Capsule. Thanks!

PR #7039, opened 2026-06-13
Add instrumentation handlers for operational warnings (#6461)

### LINKED ISSUE #6461: Notifications and Events
Several people have asked for better/deeper integration with instrumentation like ActiveSupport's notifications. Things which are log warnings today could be emitted as notifications, which are more flexible in where they go and how they can be monitored. For instance, the slow RTT warning or slow job iteration  (i.e. an iteration which takes longer than Sidekiq's timeout setting).

Provide an adapter layer for AS::Notifications. Are there any other notification APIs of any significance in Ruby?

https://guides.rubyonrails.org/active_support_instrumentation.html#creating-custom-events

Edilbek:
# Add operational instrumentation handlers (#6461)

Fixes #6461

Operational events that were previously **log-only** can now be subscribed to via `instrumentation_handlers`. Existing log output is unchanged — instrumentation is additive.

Sidekiq already has hooks for errors (`error_handlers`, `death_handlers`). Operational signals (slow Redis RTT, slow iterable iterations, forced shutdown, Redis recovery) were only written to the log, so monitoring gems often patched internals. This PR gives them a supported hook, similar to `error_handlers`.

Core API is **standalone** (no ActiveSupport dependency). On Rails, events are also forwarded to `ActiveSupport::Notifications` via `Sidekiq::Rails::Instrumentation::ActiveSupportBridge`.

## Architecture

When something operational happens (slow RTT, hard shutdown, etc.), Sidekiq calls `instrument(name, payload)`. That walks every registered handler and invokes it.

**Every app (including Rails)** can register handlers directly:

```ruby
  config.instrumentation_handlers << MyHandler.new  # or a lambda
end
```

**Rails only — extra path:** on server boot Sidekiq also registers `ActiveSupportBridge`, which re-publishes the same event into `ActiveSupport::Notifications`. So Rails apps can use either (or both):

| Path | How you consume it | Who gets it |
|------|--------------------|-------------|
| `instrumentation_handlers` | Your `#call(name, payload, config)` | Any Ruby app |
| `ActiveSupport::Notifications` | `subscribe("slow_rtt.sidekiq")` | Rails only (bridge auto-registered) |

Flow:

```
Sidekiq detects condition
        │
        ▼
  instrument(name, payload)
        │
        ├── handler 1  (e.g. OpenObserve / Datadog shipper)
        ├── handler 2  (e.g. your lambda)
        └── ActiveSupportBridge   ← Rails only
                 │
                 ▼
           ActiveSupport::Notifications
           (subscribers in the Rails app)
```

Plain Ruby apps never load the bridge — they only use `instrumentation_handlers`.

| File | Role |
|------|------|
| `lib/sidekiq/config.rb` | `instrumentation_handlers`, `instrument` (`# :nodoc:`) |
| `lib/sidekiq/component.rb` | `instrument` delegate used by launcher/processor/etc. |
| `lib/sidekiq/rails/instrumentation.rb` | AS bridge |
| `lib/sidekiq/rails.rb` | registers bridge on server boot |

`instrumentation_handlers` is broader than “warnings”: v1 OSS mostly publishes operational warnings, but the same hook can carry other signals later (including Pro/Ent).

## API

Handlers register on **server** config (events fire from worker processes):

```ruby
  config.instrumentation_handlers << ->(name, payload, config) do
    StatsD.increment(name)
    # or ship to OpenObserve / Datadog / etc.
  end

end
```

| Argument | Type | Description |
|----------|------|-------------|
| `name` | String | e.g. `"slow_rtt.sidekiq"` |
| `payload` | Hash | JSON-friendly metadata |
| `config` | Config | `Sidekiq::Config` |

Handlers run synchronously (same model as `error_handlers`). If one raises, Sidekiq logs it and continues with the rest.

Internal publish (components / Pro/Ent):

```ruby
instrument("slow_rtt.sidekiq", { readings: [...], threshold: 50_000, identity: "..." })
```

### Rails

Bridge registers automatically. Subscribe as usual:

```ruby
ActiveSupport::Notifications.subscribe("slow_rtt.sidekiq") do |*args|
  event = ActiveSupport::Notifications::Event.new(*args)
  Rails.logger.warn("RTT degraded: #{event.payload[:readings].inspect}")
end
```

Non-Rails apps use `instrumentation_handlers` only.

## OSS v1 events

| Event | When | Payload |
|-------|------|---------|
| `slow_rtt.sidekiq` | Last 5 Redis RTT samples all &gt; 50,000µs | `readings`, `threshold`, `identity` |
| `slow_iteration.sidekiq` | Iterable iteration exceeds `config[:timeout]` | `class`, `jid`, `duration`, `timeout`, `cursor` |
| `hard_shutdown.sidekiq` | Busy threads force-killed after shutdown timeout | `thread_count`, `job_count` |
| `redis_recovered.sidekiq` | Processor fetch recovers after Redis errors | `downtime` (seconds) |

Event names and payload keys are intended as a **stable contract** once merged. New keys may be added; existing keys should stay.

Boot-time config checks (e.g. Redis `maxmemory-policy`) stay log-only for v1.

**Out of scope:** job perform timing, job errors, lifecycle `on` / `fire_event`, client enqueue events.

## Operational validation

I ran all four events through a **live** Sidekiq process into OpenObserve and asked: *from this payload alone, would an on-call engineer know what happened and what to do?*

- Harness (local only, **not** in this PR): https://gist.github.com/Edilbek/8b2be7c18049566f81e6fa945c9107e5  
- Captured samples: `results.json` in that gist  

Existing Sidekiq WARN/INFO logs still fired in parallel.

### How to reproduce

| Event | Steps |
|-------|--------|
| `slow_iteration` | Set timeout to 5s; enqueue iterable job that sleeps 6s per iteration |
| `hard_shutdown` | Enqueue a long sleep job; Ctrl-C Sidekiq; wait for quiet period |
| `slow_rtt` | Need 5 consecutive over-threshold heartbeat samples. Local Redis is too fast for the real 50ms gate — I used a local-only threshold override (50µs). Prod default remains 50_000µs |
| `redis_recovered` | Stop Redis briefly while Sidekiq is polling; start Redis again |
| Broken handler | Register a raising handler after the shipper; trigger any event |

### What I saw

**`slow_iteration`** — one clean event with `class`, `jid`, `duration≈6`, `timeout=5`, `cursor`. Actionable: which job, why it fired.

**`hard_shutdown`** — `thread_count: 1`, `job_count: 1`; Sidekiq requeued the job. Actionable for deploy/timeout tuning.

**`slow_rtt`** — fired with readings like `[306, 189, 584, 202, 189]`, threshold, and process `identity`. Publish path + shape validated; I did **not** Toxiproxy a true &gt;50ms path.

**`redis_recovered`** — works, but **noisy**. One outage produced ~6 events (≈ concurrency): each processor thread recovers independently with a slightly different `downtime`. Same `pid`. Useful as a breadcrumb; raw count is a bad pager signal.

**Broken handler** — shipper still posted to OpenObserve; Sidekiq logged `!!! INSTRUMENTATION HANDLER THREW AN ERROR !!!`; process kept running.

### Summary

| Check | Fired | Actionable? | Notes |
|-------|-------|-------------|--------|
| `slow_iteration` | yes | yes | Best example of the contract |
| `hard_shutdown` | yes | yes | Counts clear; job requeued |
| `slow_rtt` | yes | yes | Local threshold override; prod 50ms untested |
| `redis_recovered` | yes | weak | ~1×concurrency fan-out |
| Raising handler | yes | n/a | Isolated; Sidekiq survives |

Open question from ops: keep `redis_recovered` as-is (document dedupe), dedupe in core, or drop from v1?

## Tests

- [x] `test/instrumentation_test.rb`
- [x] `test/rails_instrumentation_test.rb`
- [x] Non-Rails smoke (`instrument` via `ruby -Ilib`)
- [x] Rails: handlers + `ActiveSupport::Notifications`
- [x] Live Sidekiq → OpenObserve for all four events + handler isolation
- [x] `bundle exec rake test` (dedicated Redis; no competing Sidekiq)


seuros (changes requested):
The instrumentation must not couple sidekiq with activesupport.

It should delegate to it when rails exist. 

But the feature need to be isolated and standalone.

Edilbek:
Thanks @seuros  — agreed, and I've updated the PR accordingly.

### Core is standalone
- `instrumentation_handlers` + `instrument()` live in core Sidekiq
- No ActiveSupport dependency
- No `activesupport` in the gemspec
- Plain Ruby apps use handlers directly

### Rails delegates optionally
- Moved the AS adapter out of `sidekiq/instrumentation/` into the Rails layer
- New class: `Sidekiq::Rails::Instrumentation::ActiveSupportBridge`
- New file: `lib/sidekiq/rails/instrumentation.rb`
- Registered from `sidekiq/rails.rb` only when `ActiveSupport::Notifications` is defined

### Docs
- Removed ActiveSupport references from core `instrumentation.rb`
- Rails integration is documented at the Rails layer only

### Verified locally
- [x] IRB smoke test with `cfg.instrument(...)` works without Rails/AS
- [x] Rails app still receives events via `instrumentation_handlers`
- [x] Rails app still receives events via `ActiveSupport::Notifications`

Let me know if you'd prefer a different namespace or registration point for the bridge.

seuros:
A few concerns before we can consider merging this:

The introduction of instrumentation_handlers creates a new public API surface. Once merged, event names and payload structures effectively become contracts that we need to support long-term. I don't see enough consideration given to API stability and versioning here.

Handler execution is synchronous. That means a slow or faulty instrumentation handler can directly impact Sidekiq internals. 

You have Sidekiq::Rails::Instrumentation::ActiveSupportBridge  but the Instrumentation module , could be the delagation class itself.

I see a lot of AI-style defensive coding in this PR. For example, the ActiveSupport checks added around the Rails integration dont appear to be based on actual runtime requirements. In a Rails environment, ActiveSupport is already a dependency. 

Before adding defensive branches like this, please validate the assumptions in a real application.
I recommend creating a small repository containing both:

- a Rails application running Sidekiq
- a non-Rails Sidekiq application.

Use those applications to validate what instrumentation is actually needed in each environment and how users are expected to consume it. The repository would also be a useful reference implementation for others adopting this feature.

The event naming deserves another look. Rails instrumentation uses suffix namespaces (process_action.action_controller, sql.active_record), so slow_rtt.sidekiq is not inconsistent with existing conventions.

However, from a discoverability perspective, users looking for Sidekiq instrumentation may expect a sidekiq.* namespace. Mike can advice if he prefers suffix or prefix.

Why did you instrument REDIS_EVICTION_POLICY ? That part of the configuration.. and fire once.

Each instrumentation you add should understood. 

Also you have funny comments like '# INTERNAL USE ONLY', in ruby we have private, protected and #nodoc#. 

---
Btw: When you push to your branch, we get notifications , no need to comment/tag in the issue.

If you need time to prepare all that, convert the issue into draft and then turn to ready when you want a review.

Edilbek:
Thanks for the detailed review

### Testing (Rails + non-Rails)
I did validate both paths locally (rails app: sidekiq/myapp) before the refactor, but I should have documented this more clearly in the PR:
**Non-Rails** — from the repo root, no Rails loaded:

```bash
bundle exec irb -Ilib -r sidekiq
```

```
cfg = Sidekiq.default_configuration
cfg.instrumentation_handlers.clear
cfg.instrumentation_handlers << ->(event, payload, _cfg) { puts "#{event} => #{payload.inspect}" }
cfg.instrument(Sidekiq::Instrumentation::SLOW_RTT, readings: [60_000], threshold: 50_000)
```

Edilbek:
Could you re-review this PR? I did some clean-up work and removed unnecessary moments from the code and documentation.

lib/sidekiq/instrumentation.rb
```diff
@@ -0,0 +1,20 @@
...
+  # Sidekiq publishes operational events for subscribers to monitor.
+  # Sidekiq Pro and Enterprise publish additional events using the same API.
+  #
+  #   Sidekiq.configure_server do |config|
+  #     config.instrumentation_handlers << ->(event, payload, _cfg) do
+  #       puts "#{event} #{payload.inspect}"
+  #     end
+  #   end
+  #
+  module Instrumentation
```
mperham: This module is unnecessary. There's no need to extract these constants afaict.

lib/sidekiq/component.rb
```diff
@@ -79,6 +79,10 @@ def handle_exception(ex, ctx = {})
       config.handle_exception(ex, ctx)
     end
 
+    def instrument(event, payload = {}) # :nodoc:
```
mperham: I think this is a poor method name. What does `instrument` do? `fire_event` is the closest in semantic. Maybe `fire_warning` or `trigger_warning`.

lib/sidekiq/config.rb
```diff
@@ -263,6 +264,30 @@ def error_handlers
       @options[:error_handlers]
     end
 
+    # Register a proc to receive Sidekiq instrumentation events.
```
mperham: I don't know what an "instrumentation event" is. If these are all operational warnings, shouldn't they be called operational warnings or events? Runtime event?

mperham:
I think we first need to settle on terminology. Words and naming matter so we need to be very clear on naming.

lib/sidekiq/component.rb
Edilbek: Agreed — `instrument` was too vague (and collides mentally with AS/APM usage). I'll rename to `fire_warning` to parallel `fire_event` while keeping lifecycle and operational warnings separate. Open to `trigger_warning` if you prefer that verb.

Edilbek:
- Renamed to warning_handlers / fire_warning; dropped Sidekiq::Instrumentation
- AS bridge moved to Sidekiq::Rails::Warnings::ActiveSupportBridge (core stays AS-free)
- Removed defined? guard; fire_warning is # :nodoc:
- Dropped redis_eviction_policy from v1

Tests pass (warnings_test, rails_warnings_test, full suite). Manual smoke on non-Rails + Rails (myapp). PR description updated. Still drafting until we settle naming — let me know if anything else should change.

seuros:
@Edilbek, I will be direct with you.

What you are contributing here is not just code. It is a public contract. Once Mike ships it, people will build monitoring, alerting, dashboards, and integrations on top of it. If the contract is incomplete or the events are not useful in real operations, Sidekiq ends up carrying that burden for years.

One thing I want to point out: I actually think instrumentation_handlers was a better name than warning_handlers.

Instrumentation is a broad concept. Warnings are only one category. Tomorrow you might want timings, counters, queue latency, throughput metrics, state transitions, or other operational signals. Those are instrumentation too, but they are not warnings.

The funny part is that if you ask Claude 'should we rename instrumentation_handlers to warning_handlers ?' and then ask again 'should we rename it back ?', it will happily agree both times. If Mike had suggested naming the method 🔥 or 🚨 (that valid ruby code :D), the AI would probably have replied with 'Good catch!' and generated the refactor.

That is why I keep pushing on ownership rather than implementation details.

Right now I still dont see evidence that this has been tested in a real operational environment. The test suite proves the callbacks fire. That is good. It does not prove the API is useful.

Set up Grafana, SigNoz, OpenObserve, Honeycomb, Datadog, whatever you like. Generate slow RTTs. Kill workers during deploys. Force Redis outages. Create stuck jobs. Break handlers. Then look at the data as if you were the operator receiving the alert at 4am.

The important question is: 'Can an operator understand what happened and what action to take from this event?'

I know this is your first contribution here, so I will share something I learned the hard way: maintainers are not looking for code anymore. They are looking for ownership.

You need to own every line you submit. If Mike pings you six months from now asking why a warning exists, why a payload contains a field, or why a name was chosen, the answer should be 'because I tested it and here is what I learned' rather than 'because the Claude suggested it but now i use DeepSeek, so no idea'

PS: I already have an implementation of this locally. It was not AI-generated, and it already targets Sidekiq. When I saw your proposal, that is why I offered my review time. 

**For the record, this entire comment is written by me, not an AI.**

Also, your last comment claims the PR still in draft state. It is not. The PR had already been marked ready for review 2 days ago. That is a good example of why AI-assisted development still requires ownership: the model lost track of the current state and replied with outdated information.

Edilbek:
@seuros That is very good feedback. Thanks for your time! I will be more careful and check everything that I push. I will spend more time on naming (sure, I used AI for naming too, but I also checked that in Sidekiq, it already has fire_method... and decided to follow it) and will test it by following your suggestion. 

Thank you so much :)

seuros:
@Edilbek any update ?

Edilbek:
> @Edilbek any update ?

Hi @seuros. I was busy last month. Now I have time to finalize it. Yesterday I walked through commits, reviews again, and I think you're right with the namings. I used the previous one, "instrumentation_handlers". I started testing it in a real operational environment. I'm going to put some files to show as examples. Hope I'll be back soon :)

Edilbek:
@seuros Hi, 

Thanks for the push to validate this as a real ops contract.
I ran the four OSS v1 events through live Sidekiq into OpenObserve:

I added files and steps to test it with OpenObserve in this gist. I think it would be helpful. (Only local changes, didn't commit it).
- Harness: https://gist.github.com/Edilbek/8b2be7c18049566f81e6fa945c9107e5  
- Results JSON: in that gist  
- Write-up: Part 2 of the PR description

**Findings**

- `slow_iteration` / `hard_shutdown` / `slow_rtt` — payloads look actionable  
- `redis_recovered` — works, but fires ~once per processor thread (noisy)  
- Raising handler — isolated; Sidekiq kept running  

Open question: keep `redis_recovered` as-is (document dedupe), dedupe in core, or drop from v1?

Happy to adjust based on your preference.
Screenshot 2026-07-18 at 18 47 06
Screenshot 2026-07-18 at 19 57 53

Edilbek:
@mperham @seuros Hey guys. As we discussed, I did some real operational tests using OpenObserve. All cases worked fine. This PR needs your review again :). Happy to hear your thoughts.

mperham:
I'm on vacation. I can't review until August, thanks for your patience.

mperham:
I'm looking at this now. I have a few quibbles but I'll work those out myself rather than bother you with implementing them.

Edilbek:
@mperham Hey, I'm there. If you need some help, let me know!

Edilbek:
Hi @mperham, any updates?)

mperham:
Thanks for the kick in the butt to give it another round of polish. APIs always need 3-4 rounds of prototyping and polish before they are ready for general usage and long-term support.

Take a look at the latest and feel free to leave feedback.

Edilbek:
@mperham Looks good. Did you test it with OpenObserve or similar ones?

Edilbek:
@mperham Looks good. Did you test it with OpenObserve or similar tools?

mperham:
Nope, I exercise APIs via test cases. That generally gives me a sense as to whether it is useful and pragmatic or not.

I realized we need to think about debouncing the notifications. If you have 10 threads and Redis goes down, you'll get 10 notifications, one per thread. I thought maybe we could rate limit the notifications to one per minute per name. Any other ideas?

Edilbek:
@mperham 
Agreed — I hit the same thing with Redis recovery: N threads -> N notifications for one outage.

Once per minute per name (per process) seems pragmatic for redis.up / redis.down. Alternative: coalesce only for Redis up/down (thread-local → process-level), and leave rarer events like hard_shutdown / slow_iteration undebounced.

Happy to implement whichever you prefer in this PR.

mperham:
Sure, I'm always happy to have contributors, that's why I named my business contributed systems!

I hope the spec is clear and focused enough. I think your description is correct, some events are naturally singleton (hard_shutdown) or naturally multiple (slow_iteration, per-JID) and won't need debouncing.

Edilbek:
Got it, by the end of the week, I'll tag you with new improvements

Edilbek:
@mperham I debounced `sidekiq.redis.*` to once per minute per name — PTAL

PR #7042, opened 2026-06-17
Add tests for BaseTab#chart_y_axis and #striped_rows

hammadxcm:
## Summary

Follow-up coverage for the `BaseTab` rendering helpers consolidated in #7038. That dedupe pulled `chart_y_axis` and `striped_rows` up from the individual tab classes into `BaseTab`, but neither helper had a direct test of its pure-data output.

Adds a small `CapturingTUI` stand-in (matching the `FakeTUIParent` pattern already in `test/tui_test.rb`) to inspect the structure passed to `axis()` / `table_row()`:

- **`chart_y_axis`** emits evenly-spaced integer labels from `0` to `y_max` with the requested `num_labels` (covered with 5-label and 3-label cases).
- **`chart_y_axis` with `y_max == 0`** emits all-zero labels (`["0", "0", "0", "0"]`) rather than blowing up — the empty-chart guard.
- **`striped_rows`** alternates row style: `nil` for even indices, `{bg: :dark_gray}` for odd, preserving the original cells in each row.

Test-only, no `lib/` changes.

## Testing

### DIFF
--- a/test/tui_test.rb
+++ b/test/tui_test.rb
@@ -25,6 +25,30 @@ class ConcreteTab < Sidekiq::TUI::BaseTab
   end
 end
 
+# Minimal stand-in that captures the args passed to the underlying TUI calls
+# so we can assert the structure of what BaseTab renders.
+class CapturingTUI
+  attr_reader :axis_args, :table_rows
+
+  def initialize
+    @table_rows = []
+  end
+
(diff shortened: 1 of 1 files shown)


mperham:
I don't think we need unit tests at this low level of granularity.

Issue #4870, opened 2021-04-09
FloatDomainError when opening /busy


EdCordata:
Ruby version: 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-linux]
Rails version: Rails 6.0.3.3
Sidekiq / Pro / Enterprise version(s): regular sidekiq version

```ruby
# config/initializers/custom/sidekiq.rb
require 'sidekiq/web'

Sidekiq.configure_server do |config|
  config.redis = 'redis://localhost:6379/1'
end

Sidekiq.configure_client do |config|
  config.redis = 'redis://localhost:6379/1'
end
```
```ruby
# Gemfile
gem 'sidekiq', github: 'mperham/sidekiq'
```
```ruby
# routes
Rails.application.routes.draw do
  mount Sidekiq::Web, at: '/jobs', constraints: SidekiqConstraint.new #  constraint always returns true
end
```
```ruby
# app/workers/hard_worker.rb
class HardWorker
  include Sidekiq::Worker

  def perform
    sleep 5
  end

end
```

Error screenshot:
![FloatDomainError at _busy - Google Chrome 2021-04-09 11 35 41](https://user-images.githubusercontent.com/71574639/114153454-2d928d80-98d4-11eb-994c-f76108232bd9.png)

This happens only in `/busy` path, others work fine:
![1](https://user-images.githubusercontent.com/71574639/114153799-91b55180-98d4-11eb-9e88-7fbeecf4e164.png)

This error happens when `sidekiq` is not running, only rails server.

mperham:
I need the full backtrace.

PR #7047, opened 2026-07-02
Fix crash on ?count=0 in Web UI pagination

chatman-media:
_paging.html.erb divides `@total_size` by `@count` to build the "jump to last page" link, and `@count` comes straight from the `count` query param (`url_params("count").to_i`). Hitting `/retries?count=0`, `/scheduled?count=0`, `/morgue?count=0`, or `/queues/:name?count=0` on any non-empty set raises `FloatDomainError: Infinity` instead of rendering the page.

It's basically the same class of bug as #7014 (fixed for `page_items`/the Busy page), just reachable through the other four paginated views since they don't share that guard — `_paging.html.erb` itself has no protection against `@count` being zero or negative. There's actually an older closed issue for this too (#4870), but it never got a fix.

Falls back to page 1 for the "last page" link when count isn't positive, and added a regression test for each of the four affected views (queue, retries, scheduled, morgue).

### DIFF
--- a/test/web_test.rb
+++ b/test/web_test.rb
@@ -236,6 +236,13 @@ describe Sidekiq::Web do
     refute_match(/\b1003\b/, last_response.body)
   end
 
+  it "does not crash on a zero count parameter" do
+    @config.redis { |conn| conn.lpush("queue:default", Sidekiq.dump_json(args: [1])) }
+
+    get "/queues/default?count=0"
+    assert_equal 200, last_response.status
+  end
+
   it "can delete a queue" do
     @config.redis do |conn|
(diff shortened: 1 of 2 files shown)


mperham:
There's no reason for the user to do this and it doesn't affect performance so I don't feel any urgency to fix this.

Your proposed fix is not the correct change IMO: you should add logic to ensure @count has a valid value in the Ruby code before it even gets to the ERB, e.g. `web/application.rb` line 295:

```diff
- @count = (url_params("count") || 25).to_i
+ @count = [url_params("count").to_i, 25].max
```

PR #7048, opened 2026-07-08
Use smaller batch size when using `spread_interval` with `perform_bulk`

fatkodima:
Follow up to #6953.

We use a smaller `batch_size` when enqueuing scheduled jobs via `perform_bulk` to avoid redis slow log entries. 
But using a `spread_interval` also makes them scheduled, so we need to use a smaller `batch_size` too.


sobrinho:
@fatkodima what about adding specs? Is it possible? :)

fatkodima:
If you have ideas how to implement them without ugly and brittle mocks, I will add them.

agrieser-healthie (approved):

seuros (approved):

PR #7052, opened 2026-07-16
TUI: Render controls dynamically based on terminal width

13-Sonal:
## Summary

This PR replaces the fixed split of controls after 8 items with dynamic wrapping based on the available terminal width.

This addresses the existing `# TODO Dynamically split based on term width?` in `lib/sidekiq/tui.rb`.

## Changes

- Remove the hardcoded split after 8 controls.
- Dynamically wrap controls based on the available terminal width.
- Support multiple lines of controls when needed.
- Preserve the existing controls and functionality.

## Before

The controls were always split after the first 8 items, regardless of the terminal width.

## After

The controls automatically wrap based on the terminal width, making better use of available space on wider terminals while remaining readable on narrower ones.

## Testing

- Started the TUI using `bundle exec ruby bin/kiq`.
- Verified the controls wrap correctly on different terminal widths.
- Verified all controls are displayed and continue to function as expected.

# Test results

### Before
image

### After
image

### DIFF
--- a/lib/sidekiq/tui.rb
+++ b/lib/sidekiq/tui.rb
@@ -154,17 +154,36 @@ module Sidekiq
     def render_controls(frame, area)
       active_keys = current_tab.controls.filter { |hash| hash[:description] }
 
-      # Split controls into two lines, 8 is arbitrary
-      # TODO Dynamically split based on term width?
-      first = active_keys[...8]
+      # Dynamically split controls based on terminal width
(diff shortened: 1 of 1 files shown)


mperham:
I don't see the point of the test code. This is something you test once manually. Could you remove the test code?

13-Sonal:
> I don't see the point of the test code. This is something you test once manually. Could you remove the test code?

Hi @mperham Got it, thanks
I added the tests to validate the wrapping logic, but I understand this is better verified manually. I've removed them and push an update.

PR #7055, opened 2026-07-19
Add specs for default batch size in push_bulk

sobrinho:
Cover the behavior introduced in #6953 (smaller default batch size of 100 for scheduled jobs) and #7048 (apply the same default when jobs are scheduled via spread_interval).

Since the batch size is only observable in how push_bulk slices its Redis round-trips, these specs attach a redis-client middleware to a dedicated connection pool to record pipelined command batches and assert how many jobs each ZADD/LPUSH carried, without mocking any Sidekiq internals.

### DIFF
--- a/test/client_test.rb
+++ b/test/client_test.rb
@@ -77,6 +77,28 @@ class ChildHash < Hash
   end
 end
 
+# Records the batches of commands sent to Redis via pipelining so tests can
+# observe how many round-trips an operation performed. Not registered globally:
+# attach it to a dedicated pool via RedisClient's `middlewares:` config option.
+module PipelinedCommandRecorder
+  class << self
+    attr_accessor :pipelines
+
+    def record
+      self.pipelines = []
+      yield
+      pipelines
+    ensure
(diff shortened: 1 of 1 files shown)


sobrinho:
cc @mperham @fatkodima

fatkodima:
I don't like tests that test implementation internals, but this is up to Mike.

mperham:
I'm on vacation until August. I don't see this as critical to review before then.

sobrinho:
@mperham have you had time to review this?

mperham:
Sorry, I lost this so thanks for the reminder.

I don’t particularly care for the test implementation either, is there a way to verify by mocking Sidekiq internals, instead of Redis-client internals? That will make the tests more robust.

sobrinho:
Well, we are not mocking Redis internals here, the middleware thing is a public documented API to intercept the commands to Redis.

I can try to mock but would rather prefer to keep as is right now since it uses a public API instead of actually mocking.

PR #7056, opened 2026-07-22
Metrics: draw deploy marks from every day the queried window spans

### LINKED ISSUE #7053: Metrics: deploy marks from previous days are never drawn (fetch_marks only reads today's key)
`Sidekiq::Deploy#mark!` stores marks under a **per-day** key (`YYYYMMDD-marks`, 90-day TTL), but `Metrics::Query#fetch_marks` only reads **today's** key and then filters by the window. On any period longer than the current UTC day — including the Metrics page's own `?period=72h` — deploys from previous days silently disappear from the graph, while their data is still sitting in Redis.

Nothing raises, so there is no backtrace: the marks are simply dropped from the result.

### Versions

Ruby 3.4.10, Rails 8.1.3, Sidekiq 8.1.6 (also present on `main` today), redis-client 0.30.0, Redis 7, Debian (official `ruby` image).

No metrics-related configuration: stock `Sidekiq::Web` mounted, marks written from a deploy hook via `Sidekiq::Deploy.mark!`.

### Reproduction

Stock Sidekiq, empty Redis db, no app code:

```ruby
require "sidekiq"
require "sidekiq/deploy"
require "sidekiq/metrics/query"

Sidekiq::Deploy.new.mark!(at: Time.now.utc - (26 * 3600), label: "yesterday-deploy")
Sidekiq::Deploy.new.mark!(label: "today-deploy")

q = Sidekiq::Metrics::Query.new
q.send(:fetch_marks, (Time.now.utc - (72 * 3600))..Time.now.utc, :hourly).map(&:label)
```

Actual:

```
fetch_marks over 72h => ["today-deploy"]
stored yesterday     => ["yesterday-deploy"]   # Sidekiq::Deploy.new.fetch(Date.today - 1)
stored today         => ["today-deploy"]
```

Expected: both labels — both timestamps fall inside the 72h range, and both are still in Redis.

Through the UI: mark a deploy, let the UTC day roll over, mark another, then open `/sidekiq/metrics?period=72h`. Only the newest is drawn. On a live install here, `Sidekiq::Deploy.new.fetch(Date.today - 1)` returns 3 marks that the 72h view never draws.

### Cause

The write side is per-day and long-lived — [lib/sidekiq/deploy.rb](https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/deploy.rb):

```ruby
datecode = floor.strftime("%Y%m%d")
key = "#{datecode}-marks"
pipe.hsetnx(key, stamp, label)
pipe.expire(key, MARK_TTL)   # 90 days
```

The read side reads exactly one day — [lib/sidekiq/metrics/query.rb#L170-L181](https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/metrics/query.rb#L170-L181) on current `main`:

```ruby
def fetch_marks(time_range, granularity)
  [].tap do |result|
    marks = @pool.with { |c| c.hgetall("#{@time.strftime("%Y%m%d")}-marks") }   # <- only @time's date

    marks.each do |timestamp, label|
      time = Time.parse(timestamp)
      if time_range.cover? time
        result << MarkResult.new(time, label, Query.bkt_time_s(time, granularity))
      end
    end
  end
end
```

`time_range` spans up to 72h (3–4 UTC days), so the `time_range.cover?` filter can only narrow a set that never contained more than today in the first place.

### Suggested fix

Read every day the window actually spans:

```ruby
def fetch_marks(time_range, granularity)
  days = (time_range.begin.utc.to_date..time_range.end.utc.to_date)
  pairs = @pool.with do |conn|
    days.flat_map { |day| conn.hgetall("#{day.strftime("%Y%m%d")}-marks").to_a }
  end

  pairs.filter_map do |timestamp, label|
    time = Time.parse(timestamp)
    MarkResult.new(time, label, Query.bkt_time_s(time, granularity)) if time_range.cover?(time)
  end
end
```

That is at most 4 `HGETALL`s for the largest window, each against a small hash.

Happy to send a PR with this plus a test if it's the direction you'd want.

mrlewap:
Fixes #7053.

`Sidekiq::Deploy#mark!` stores marks under per-day keys (`YYYYMMDD-marks`, 90-day TTL), but `Metrics::Query#fetch_marks` read only today's key before applying the `time_range` filter — so on any window longer than the current UTC day (including the Metrics page's own 72h period) deploys from previous days silently vanished from the graphs while still present in Redis.

This reads every date key the queried window spans (at most 4 `HGETALL`s for the 72h view, each against a small hash) and keeps the existing range filter. Windows within a single UTC day behave exactly as before (one key, as today).

Test included: red on current `main` (`Expected ["today deploy"] to include "yesterday deploy"`), green with the fix. Full `bundle exec rake` (standard + lint:herb + test, 749 runs) passes.

### DIFF
--- a/lib/sidekiq/metrics/query.rb
+++ b/lib/sidekiq/metrics/query.rb
@@ -171,8 +171,10 @@ module Sidekiq
         [].tap do |result|
           days = (time_range.begin.utc.to_date..time_range.end.utc.to_date)
           marks = @pool.with { |c|
-            days.flat_map { |day| c.hgetall("#{day.strftime("%Y%m%d")}-marks").to_a }
-          }
+            c.pipelined { |pipeline|
+              days.each { |day| pipeline.hgetall("#{day.strftime("%Y%m%d")}-marks") }
+            }
+          }.flat_map(&:to_a)
 
           marks.each do |timestamp, label|
             time = Time.parse(timestamp)
--- a/lib/sidekiq/metrics/query.rb
+++ b/lib/sidekiq/metrics/query.rb
@@ -169,7 +169,10 @@ module Sidekiq
 
       def fetch_marks(time_range, granularity)
         [].tap do |result|
-          marks = @pool.with { |c| c.hgetall("#{@time.strftime("%Y%m%d")}-marks") }
+          days = (time_range.begin.utc.to_date..time_range.end.utc.to_date)
+          marks = @pool.with { |c|
+            days.flat_map { |day| c.hgetall("#{day.strftime("%Y%m%d")}-marks").to_a }
+          }
 
           marks.each do |timestamp, label|
             time = Time.parse(timestamp)
--- a/test/metrics_test.rb
+++ b/test/metrics_test.rb
@@ -104,6 +104,18 @@ describe Sidekiq::Metrics do
       assert_equal 1, rs.size
       assert_equal "cafed00d - some git summary line", rs[floor]
     end
+
+    it "returns marks from every day the queried window spans" do
+      whence = Time.utc(2022, 7, 22, 22, 3, 0)
+      d = Sidekiq::Deploy.new
+      d.mark!(at: whence - 26 * 60 * 60, label: "yesterday deploy")
+      d.mark!(at: whence, label: "today deploy")
+
+      q = Sidekiq::Metrics::Query.new(now: whence)
+      labels = q.top_jobs(hours: 72).marks.map(&:label)
+      assert_includes labels, "today deploy"
+      assert_includes labels, "yesterday deploy"
+    end
   end
 
   describe "histograms" do


lib/sidekiq/metrics/query.rb
```diff
@@ -169,7 +169,10 @@ def self.bkt_time_s(time, granularity)
 
       def fetch_marks(time_range, granularity)
         [].tap do |result|
-          marks = @pool.with { |c| c.hgetall("#{@time.strftime("%Y%m%d")}-marks") }
+          days = (time_range.begin.utc.to_date..time_range.end.utc.to_date)
+          marks = @pool.with { |c|
```
mperham, line 173: You should use a pipelined block here to reduce network round trips to Redis.

mrlewap:
Done — the per-day HGETALLs now go through one `pipelined` block (same idiom as api.rb). Full `bundle exec rake` green.

mperham:
It looks like you didn't run the tests locally, standardrb will reformat your code when you run rake.

mrlewap:
I did run it locally - bundle exec rake (standard + lint:herb + test) is green on this branch, and standardrb lib/sidekiq/metrics/query.rb test/metrics_test.rb (standard 1.56.0) reports no offenses; --fix changes nothing in either file. The block formatting mirrors the existing pipelined blocks in api.rb. If CI flags anything once the workflows are approved, happy to fix.

mperham:
Strange, the brace block should be converted to do..end.

PR #7057, opened 2026-07-26
Update Tamil locale to match latest en.yml

ferdinandrosario:
This PR:
- Adds Tamil translations for the missing keys
- Removes the keys that aren't in `en.yml` anymore

### DIFF
--- a/web/locales/ta.yml
+++ b/web/locales/ta.yml
@@ -5,73 +5,107 @@ ta:
   AddToQueue: வரிசையில் சேர்
   AddAllToQueue: அனைத்தையும் வரிசையில் சேர்
   AreYouSure: நீங்கள் உறுதியாக இருக்கிறீர்களா?
-  AreYouSureDeleteJob: நீ இந்த வேலையை நீக்க வேண்டும் என்று உறுதியாக இருக்கிறீர்களா?
-  AreYouSureDeleteQueue: நீங்கள் %{queue} வரிசையில் நீக்க வேண்டும் என்பதில் உறுதியாக இருக்கிறீர்களா?
-  Arguments: வாதங்கள்,
-  Busy: பணிமிகுதி
+  AreYouSureDeleteJob: இந்த பணியை நீக்க விரும்புகிறீர்களா?
+  AreYouSureDeleteQueue: "%{queue} வரிசையை நீக்க விரும்புகிறீர்களா? இது வரிசையிலுள்ள அனைத்து பணிகளையும் நீக்கிவிடும், எதிர்காலத்தில் மேலும் பணிகளை அனுப்பினால் இது மீண்டும் தோன்றும்."
+  Arguments: அளவுருக்கள்
+  BackToApp: பயன்பாட்டிற்குத் திரும்பு
+  Busy: செயலில்
   Class: வகுப்பு
   Connections: இணைப்புகள்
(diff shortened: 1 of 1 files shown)


seuros:
Please attach screenshoots of the different pages and prompts. 

I see some string getting longer

ferdinandrosario:
@seuros, the Tamil translation of English will be different because of its grammar, so we cannot expect the same number of words.

mperham:
The main issue is AreYouSureDeleteQueue, which is significantly longer and may not fit on some screens. Can you shorten it?

seuros:
We need to see how it looks visually. 
if you tested it and use it , grab screenshot and drag and drop as comment.  

We better have broken grammar such as 'wanna delete this ?' than `Do you really want to ..`  while the rest of the sentence is overflowing the div.

PS: Arabic is more verbose than English, if we write the long form, we have to touch the css. Tamil might have same problem.

Btw thank for the contribution.

ferdinandrosario:
> The main issue is AreYouSureDeleteQueue, which is significantly longer and may not fit on some screens. Can you shorten it?
sure @mperham

ferdinandrosario:
> We need to see how it looks visually. if you tested it and use it , grab screenshot and drag and drop as comment.
> 
> We better have broken grammar such as 'wanna delete this ?' than `Do you really want to ..` while the rest of the sentence is overflowing the div.
> 
> PS: Arabic is more verbose than English, if we write the long form, we have to touch the css. Tamil might have same problem.
> 
> Btw thank for the contribution.

I will try to take some screenshots in the demo app @seuros. Thank you for the feedback.

ferdinandrosario:
Both `AreYouSureDeleteQueue` entries in `en.yml` and `ta.yml` have the same character length, so I don't anticipate any issues with UI CSS breaking.

Screenshot 2026-07-27 at 11:02:40 PM

The placeholder `%{queue}` appears at the beginning of `ta.yml` and in the middle of `en.yml`. If this seems unusual, don't worry; it's actually similar to English, as Tamil grammar allows for that structure. Both versions are identical in meaning.

Screenshot 2026-07-27 at 11:09:39 PM

ferdinandrosario:
@seuros will attempt to include the screenshot with the message to check for any CSS issues.

seuros:
> Both AreYouSureDeleteQueue entries in en.yml and ta.yml have the same character length, so I don't anticipate any issues with UI CSS breaking.

Unless i verify visually, i don't trust count.

iiiiiiiiii
1111111111
wwwwwwwwww
MMMMMMMMMM

Look above.. Same number of chars (10 chars), different size.

But if i change the font/formatting
```
iiiiiiiiii
1111111111
wwwwwwwwww
MMMMMMMMMM
```

suddenly they are similar length, still not pixel perfect. 4 lines, 4 difference visually.

ferdinandrosario:
@mperham @seuros, please see the attached screenshot.

Screenshot 2026-07-28 at 5 10 08 PM

Screenshot 2026-07-28 at 5 13 06 PM

@seuros, if you would like to see any other page, please let me know. I have a demo app and can provide the screenshot quickly. 😄

seuros (approved):

seuros:
No it fine. 

Just assert that smaller screens doesn't overflow.

ferdinandrosario:
Thank you, @mperham and @seuros, for the review! 🫂

PR #7062, opened 2026-08-24
Use real thread-locals for capsule and redis-pool routing

### LINKED ISSUE #6370: Not thread local
https://github.com/sidekiq/sidekiq/blob/0226e311d02de8ea77563c17c96a3d6527d0ec7f/lib/sidekiq/processor.rb#L74

This variable is not thread local.

I suggest you use a thread attribute, e.g. see <https://github.com/rspec/rspec-support/pull/610> for what I suggest as the recommended approach: `Thread.attr_accessor :sidekiq_capsule`.

kkaznor1:
Fixes #6370.

`Thread.current[:key] = value` sets a fiber-local, not a thread-local — a `Fiber` created on a thread doesn't inherit values its parent set this way. Sidekiq uses this pattern for capsule/redis-pool routing (`Sidekiq::Client.via`, `Processor`'s capsule tracking, job/sharding's pool override), so a job that spawns a Fiber internally (e.g. a Fiber-based async HTTP client) can't see the enclosing capsule or pool from inside it, and silently falls back to the default one.

This switches `:sidekiq_capsule`, `:sidekiq_redis_pool`, and `:sidekiq_client_class` to `Thread#thread_variable_get/set`, real thread-locals visible from any Fiber on that thread.

Added a test to `sharding_test.rb`: pushing a job from inside a `Fiber` during `Sidekiq::Client.via` now routes to the right shard. Red on `main`, green with the fix. Full `bundle exec rake` (750 runs) passes.

I acknowledge the Sidekiq contribution policy: by submitting this PR, I assign copyright of these changes to Contributed Systems LLC.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mperham:
Stepping back a bit, Is this a semantic we want to change at this point? Is there any backwards compatibility issue to be concerned about here?

mperham:
For sure I would need to coordinate new OSS, Pro and Ent releases. And I think this should likely go in a minor bump, not a patch release if we accept it.

kkaznor1:
Thanks for your reply.
Fair concern, and I honestly can't answer your question.
I don't have visibility into Pro/Ent's source. I can't tell if they read Thread.current[:sidekiq_capsule] or :sidekiq_redis_pool directly.

kkaznor1:
One option : write to both the old fiber-local key and the new thread-local. Reads would prefer the thread-local, falling back to the old key. That keeps today's behavior unchanged for anything still reading the old way. Only Fiber-spawning job code would newly benefit.

Downside: it doubles every read/write site. It's a compatibility shim, not a clean fix.

Happy to add this option above if useful, or to defer to your call on scope and versioning here.

mperham:
Are you using this functionality? Is anyone actually trying to use Fibers within jobs? There are cases where Fibers are useful but I discourage their use in application code as the Fiber execution model is more complex than Threads, most Rubyists do not have the mental model to use them.

If you want parallel IO, it's better to use a client library which implements that parallelism internally for you, Typhoeus being the oldest example of a gem I can think of:

https://github.com/typhoeus/typhoeus#example

kkaznor1:
That makes sense,  I don't have a real-world case where I hit this myself; I found it doing a code read of the thread-local patterns. Anyway thanks for taking time to think it through, appreciate this discussion.

PR #7073, opened 2026-09-02
Add Pro 8.1.6 changes

cbartlett:
I noticed 8.1.6 was released and I always check this file for the changes but they were not listed so I am adding them here.


mperham:
Bizarre, I wrote a changelog entry (didn't push it) that was worded identically to your PR. Thanks for the nudge to finish it.

cbartlett:
@mperham I got that wording from the changelog in the pro gem itself. I just prefer the public web version here since it's easier than spelunking through the code. Thanks for adding it!

PR #7076, opened 2026-09-07
Add partitioned queue processing middleware

paymaan73:
Jobs sharing a partition key (an argument, chosen via sidekiq_options partition:) run one at a time in enqueue order, while different keys run concurrently. A job whose partition is busy is rescheduled a few seconds later rather than blocking a worker thread.


mperham:
Sidekiq does not provide features to work around lack of concurrency. See (3): https://sidekiq.org/wiki/Best-Practices

This feature is already available using the Concurrent rate limiter in Sidekiq Enterprise. https://sidekiq.org/wiki/Ent-Rate-Limiting

